PR #7, opened 2017-09-22
A Flask application skeleton with Travis, Codecov & Coveralls integration. Fixes #5.

### LINKED ISSUE #5: Integration with test coverage services
We should set up integration with either Coveralls or Code Climate (or both) to monitor test coverage.

tvirolai:
Although there seems to be a some problems with the coverage reports. This seems to be at part due to the fact that the integrations are found only on this branch and not in the master.


osma:
Would it be better to place the code that's currently in `annif.py` in a directory/module `annif`? I think it makes sense to try to keep the root of the project as small as possible.

Other than that, looks very good!

osma:
OK to merge if you like...

PR #25, opened 2017-10-09
Issue3

tvirolai:
Basic implementation for create-project and drop-project.

### DIFF
--- a/annif/annif.py
+++ b/annif/annif.py
@@ -32,6 +32,15 @@ projectIndexConf = {
         }
 
 
+# projectType = {
+#         'mappings': {
+#             'project': {
+# 
+#                 }
+#             }
+#         }
+
+
 @annif.cli.command('init')
 def init():
     """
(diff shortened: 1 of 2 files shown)


annif/annif.py
```diff
@@ -54,9 +55,24 @@ def listprojects():
...
+                }
+            }
+    results = es.search(index=annif.config['INDEX_NAME'], doc_type='project',
+            body=doc)
+    projects = [x['_source'] for x in results['hits']['hits']]
+    for p in projects:
+        print(p)
+
+    print(projects)
+    report = cat.indices()  # This queries different indices and returns
```
osma: Is this really necessary? The result seems to be just thrown away.

annif/annif.py
```diff
@@ -54,9 +55,24 @@ def listprojects():
...
+                'match_all': {}
+                }
+            }
+    results = es.search(index=annif.config['INDEX_NAME'], doc_type='project',
+            body=doc)
+    projects = [x['_source'] for x in results['hits']['hits']]
+    for p in projects:
+        print(p)
+
+    print(projects)
```
osma: This is redundant after the previous for loop

tvirolai: Oops, true. I tried different ways to present the results when configuring tests. This I'll remove this.

annif/annif.py
tvirolai: Not really, same thing applies here. I'll do another commit to remove these before merging.

PR #29, opened 2017-11-03
Issue10

tvirolai:
This PR fixes issue 10.

It contains the following changes:

- Flask has been upgraded to Connexion
- The API specification is mapped to the code and the Swagger UI works (although not all of the specified functionality is implemented yet).
- The directory structure has been changed a little bit. The only way I could get the imports working everywhere was to move the stuff from the directory `./annif` to the project root.
- The annif.py file has also been refactored a bit. After struggling a little bit with how to call the functions both via the API and the command line, I separated these tasks to different functions. For example, the function `show_project` now returns JSON when called via the API. When called from the command line, the click command 'show-project' calls another a function that calls 'show_project' and parses the response to a human-readable string. This works and I think this separation of concerns makes testing easier in the long run. Dunno how elegant a solution this is.


osma:
Looks good! It's definitely a big step forward so I think we can merge this.
I hope we can find a way to restore the directory structure so that we don't need to have annif.py and other Python scripts in the root directory, but it's OK for now.

PR #159, opened 2018-09-04
Return hits from backends and projects as AnalysisHits instead of plain list

### LINKED ISSUE #158: Refactor the way analysis results are returned
Related to #141 

The way results are returned from backends should be re-architected. Currently the results are returned as sequences of AnalysisHit objects, but this makes it impossible to use e.g. sklearn metrics or any other vector based processing.

Instead hits should be returned as objects that implement an ABC, which could be called AnalysisHits. The ABC should have methods (or properties) that give the results either as a sequence of AnalysisHit objects (as currently used) or as vectors of size N, where N is the number of subjects in the vocabulary.

osma:
Fixes #158

### DIFF
--- a/annif/backend/dummy.py
+++ b/annif/backend/dummy.py
@@ -1,7 +1,7 @@
 """Dummy backend for testing basic interaction of projects and backends"""
 
 
-from annif.hit import AnalysisHit
+from annif.hit import AnalysisHit, AnalysisHits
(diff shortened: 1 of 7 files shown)


osma:
Could also add a static from_vector method that handles conversion from vector to list. Then the tfidf backend could use that instead of converting by itself

PR #219, opened 2018-11-28
First implementation of LSI backend, with tests. Fixes #201

### LINKED ISSUE #201: LSI backend
We are currently using Gensim only for the basic TF-IDF backend. It should be almost trivial to create an LSI backend, it's just one extra LsiModel layer and a single parameter (number of dimensions).

LDA would be possible too, but I'll leave that for another issue.

osma:
Implementation of #201


osma:
Closing this PR for now, see #201 for evaluation results.

Issue #24, opened 2017-10-05
Administration UI


osma:
We could have a UI for administering most/all aspects of an Annif installation. It would most likely be a single-page dynamic web app that uses the REST API to perform admin functions.

superciccio:
Hi, 
I came here from a thread on reddit a guy told me about the administration UI.
I'm still fairly new to python I only have written simple scripts ( connect to firebase download images and then compress them ) for personal use.

I also have a background as Java Web developer, but I want to make a shift to Python so that is that reason.
I've worked mostly with Angular but study by my self also Vue.

My question is, do you need help on this task ? Is something that I can pick up ?

Let me know, I would be happy to help

osma:
Thanks for your kind offer @superciccio !

The current UI is very basic. It was recently reworked a bit by @kinow (the guy from reddit I guess!), changing from jQuery to Vue.JS (see #245). That was merged, but there are still parts of jQuery being used. It would be great to be able to get rid of the rest of jQuery too. I think that's mainly used for performing a couple of different REST API calls. See [this comment](https://github.com/NatLibFi/Annif/pull/245#issuecomment-458504479) for some ideas about replacements. So I'd be happy to receive a PR which removes the jQuery dependency and switches over the functionality that VueJS cannot perform to some more lightweight alternative.

Perhaps more related to this issue at hand, eventually I think we want to extend the UI into more administrative functions. Right now it's just a simple form for testing Annif on different documents. However, many UI admin functionalities will require also the REST API to be extended, as right now it's pretty much a read-only API - the real administration happens by editing `projects.cfg` and executing CLI commands. 

An easy starting point would be to add UI functionality for looking at project configuration, similar to the `list-projects` and `show-project` CLI commands. This should already be possible with the current REST API, because you can use GET on a project to ask for information about it.

superciccio:
Hi @osma 

I have some question

Can I update Vue to latest version 2.6.10. Also I see that eveything is inside home.html, can I also do some refactor and export it in a new file ?

> Perhaps more related to this issue at hand, eventually I think we want to extend the UI into more administrative functions
 Do you have something in mind? Like having pages behind authentication or define user/role in config file ?

> It would be great to be able to get rid of the rest of jQuery too
I'll look into, no problem.

> An easy starting point would be to add UI functionality for looking at project configuration, similar to the list-projects and show-project CLI commands. 
Great !

osma:
> Can I update Vue to latest version 2.6.10

[...]

PR #245, opened 2019-01-26
Replace JQuery+HTML by Vue.JS

kinow:
This removes:

- `annif.js`
- `base.html` template

Then adds a `<script>` for the Vue.JS CDN (simplest form to get started with Vue.JS). And finally replaces the old logic from the JQuery APP by a VueJS with three components. The components are not really necessary.

We could have the same with a simple application, but it would be a bit harder to see what was doing what. Also, this serves as base for further enhancements.

Here's what it looks like running.

success

If the VueJS developer bar is installed in Firefox, it's possible to see the state of the application.

success-vuebar

Also added some validation. I followed the installation instructions up to the point of having TF-IDF English. Analyzing text with other projects resulted in a server error not displayed to the user. So added a simple validation to display an error.

ajax-server-error

Likewise, in case no results are returned after the analysis, an alert is displayed.

no-results

And in case you do not select a project, or do not enter the text, another alert is displayed.

validation

It would be much simpler to implement this with Vuex, but that's overcomplicating for now I believe. But in case using VueJS sounds like a good idea, before extending its use over the admin pages (#24 ) it would be a good idea to update this app, to use vue-cli, vuex, etc (which takes much longer, but well worth).

Cheers
Bruno


osma:
Great, thanks a lot! Indeed Vue is probably a better basis for further development of the UI. Will do more thorough review next week!

osma:
Found the time to look at this in more detail.

Observation: this isn't (currently) really replacing jQuery with VueJS, since the requests to the REST API are still done via jQuery as well as some of the form manipulation. I'm not very familiar with what Vue can and cannot do (never really used it myself), but I think it would be desirable to try to build on just one JS framework instead of having a mixture of two, WDYT? Vue itself apparently cannot perform HTTP requests, but there's [vue-resource](https://github.com/pagekit/vue-resource/) for that.

I also wonder whether [bootstrap-vue](https://bootstrap-vue.js.org/) fits into this somehow or whether it's better to keep Bootstrap and VueJS as separate dependencies?

Right now Bootstrap and jQuery are bundled with the Annif codebase, but you chose to load VueJS from a CDN. I have no strong opinion on which way is better but I think this should be done consistently.

Personally I don't have any big plans for the UI - so far it has been enough to have a form for testing where it's easy to paste text from elsewhere. But moving into the direction of an admin API (#24) - perhaps similar to the Fuseki2 web UI - would be nice and this seems like a sensible first step in that direction.

All in all I'm positive to this and if we can agree on the direction (e.g. the thoughts/questions above) and you feel that the code is ready for merging, I'm happy to pull it in.

kinow:
>I also wonder whether bootstrap-vue fits into this somehow or whether it's better to keep Bootstrap and VueJS as separate dependencies?
>
>Right now Bootstrap and jQuery are bundled with the Annif codebase, but you chose to load VueJS from a CDN. I have no strong opinion on which way is better but I think this should be done consistently.

Right. For this case, we probably need to go with `vue-cli` or `webpack`, to have the `.vue` components, stylesheet preprocessor, transpilers, etc.

kinow:
>Personally I don't have any big plans for the UI - so far it has been enough to have a form for testing where it's easy to paste text from elsewhere. But moving into the direction of an admin API (#24) - perhaps similar to the Fuseki2 web UI - would be nice and this seems like a sensible first step in that direction.
>
>All in all I'm positive to this and if we can agree on the direction (e.g. the thoughts/questions above) and you feel that the code is ready for merging, I'm happy to pull it in.

This pull request only provides a Vue.js app, where Vue is taking care of reactivity, components... we could replace the JQuery `$("...")` calls to create elements and locate elements by templates and vue/DOM manipulation.

But to completely ditch JQuery and use something like Axios or vue-resource, and to have bootstrap + vue (which I'm using in another project), or even something like `vuetify` (using for yet another project), we would have to either close this one and start a larger pull request (which will take a bit longer to implement), or merge this one and work on the other pull request, re-using parts of what was done here.

All in all, I think it really depends on whether you prefer to have a Vue app, or if JQuery+HTML would be enough for the admin interface #24 too... it could be an overkill to use Vue if that interface will have just one page perhaps.

osma:
Thanks for your thoughts @kinow! I think it makes sense to do small incremental changes, rather than try to overhaul everything at once. So moving from jQuery + Bootstrap to VueJS + jQuery + Bootstrap is a good first step. Later we may want to replace jQuery with Axios or even just the native [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) as was suggested by @natlibfi-arlehiko in another discussion.

Before merging this PR I would simply like to bundle the VueJS dependency with the Annif codebase instead of relying on a CDN. The simplest way I can think of is to store [what the CDN returns](https://cdn.jsdelivr.net/npm/vue@2.5.22/dist/vue.js) into a file, just like we do for jQuery and Bootstrap. Does that make sense?

CDNs have their benefits for public-facing web sites, but I think that in most cases this UI will be used from localhost and thus loading the dependencies (JS and CSS) from localhost too makes sense. It also makes it possible to use the Annif web UI without an internet connection. I've done Annif development on a plane occasionally...

kinow:
Hi @osma,

Sure, grabbing the CDN version, or this link for the production version: https://vuejs.org/js/vue.min.js (minified).

Will update the pull request tomorrow or Friday (few meetings tomorrow and Thu). Or feel free to update this PR if you prefer to merge it sooner.

osma:
Great @kinow ! I'm probably working on other stuff (mainly Vowpal Wabbit related) for version 0.39. I'll let you update the PR, then it can go in too.

kinow:
Great! And first time I ever heard about Vowpal Wabbit! Got to read about it now. Thanks @osma !

kinow:
Rebased branch, and also added `vue.min.js` in the static resources within Annif.

image

PR #258, opened 2019-02-13
Show detailed error messages from REST API in UI

osma:
This PR adds detailed error messages which are shown in the web UI in case the analyze request to the REST API fails.

annif-ui-error

@kinow what do you think, I did this using jQuery but should this be Vue-ized somehow?

### DIFF
--- a/annif/templates/home.html
+++ b/annif/templates/home.html
@@ -172,10 +172,19 @@ var app = new Vue({
             this_.results.push(result);
           }
         },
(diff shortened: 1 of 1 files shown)


kinow:
I think we can add a bit more of Vue and replace more of jQuery. Let me give it a try.

kinow:
See #260

osma:
#260 implemented this much better, so closing this one.

PR #260, opened 2019-02-14
Vue changes to display complete server error, disable and show spinner when submitting, and use less jQuery

kinow:
A few commits in this pull request, sorry. Started with a small change to display the problem (now a problem here = error or warning), but ended up needing to make a few more adjustments in order to properly use Vue.js and almost 0 jQuery.

First change replaces `v-on:` by the shorthand `@` as this one doesn't trigger an error on WebStorm/PhpStorm/etc. It is [the same thing](https://vuejs.org/v2/guide/syntax.html#Shorthands), as well as `v-bind:abc=` and `:abc=`.

Then created a `Problem` class in JS, to store the data returned from the server side. This class is now used to populate an array in the `App` Vue instance (the main component in Annif app).

There was data in three components, and it would go out of control later. So moved all data to the `App` instance, passing to components when necessary via [component props](https://vuejs.org/v2/guide/components.html#Passing-Data-to-Child-Components-with-Props).

In one case, it was not possible to use props, as we needed to send data from the component back to the parent (in the projects drop down, when one project is selected, you want to tell the `App` instance which one it was). Before it was done via the DOM, with jQuery. But to use Vue.js reactivity, needed to [emit an event](https://vuejs.org/v2/guide/components-custom-events.html) with the project selected.

Then added the method in the `App` instance to listen to the event. Did the same for the text area. Now when the user types anything, it updates the value in the `App` instance.

Finally, there is a button in the `App` instance, that triggers the function to analyze the text. It is easier now, as the `App` contains everything necessary for the request, without having to share global variables, or search around the DOM structure. Also, it was easier now to add a `loading` attribute, that is used to disable the button, and show a spinner.

The spinner is available in bootstrap 4.3, but not 4.1 (which Annif was using), so I took the freedom of upgrading it in this pull request too.

Oh, one last change was to add `venv` to `.gitignore`. Not sure if there is anywhere that says that this is the default or de-facto name of the virtualenv created, but I've always used it? I think the [GitHub Python default gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore) also includes `venv`, so it should be OK?

Let me know if you prefer that I drop some changes, or move to separate pull requests.

Cheers
Bruno

ps: validation could be much better, but I prefer to wait for the complete app and then use vuelidate or vee-validate

ps2: adding i18n is easy-peasy-japanese with [this vue libraryhttps://github.com/kazupon/vue-i18n), if it's OK to have i18n in the browser/client-side. Both these ideas for future I believe...


annif/templates/home.html
```diff
@@ -29,72 +29,91 @@ <h1 class="py-3">Annif</h1>
...
+  hasTitle() {
+    return this.title.trim() !== '';
+  }
+
+  hasDetail() {
+    return this.detail.trim() !== '';
+  }
+}
+
+Vue.config.devtools = true;
```
kinow: This is so that the browser extension can be used to inspect the app. Probably best to remove it when deploying to production. Let me know if you prefer that I remove it, or comment.

annif/templates/home.html
```diff
@@ -122,60 +137,96 @@ <h2 class="mt-4">Results</h2>\
...
   el: '#app',
   data: {
-    results: results_,
-    errors: []
+    text: '',
+    project: '',
+    projects: [],
+    results: [],
+    problems: [],
+    loading: false
```
kinow: Now all the data of the app is centralized here in the `App`. I think it is easier to maintain it now, but would be even easier with Vuex later on.

annif/templates/home.html
```diff
@@ -122,60 +137,96 @@ <h2 class="mt-4">Results</h2>\
...
     },
+    onProjectSelected(project) {
+      this.project = project;
+    },
+    onTextChanged(text) {
+      this.text = text;
+    },
+    loadProjects: function() {
+      var this_ = this;
+      $.ajax({
```
kinow: This is the only use of jQuery now. Can be easily replaced now by Vue Resource, Axios, etc.

kinow:
Data being loaded in the main `App` component.

screenshot_2019-02-14_15-07-40

screenshot_2019-02-14_15-08-09

Data after results are returned from server.

screenshot_2019-02-14_15-14-10

And the Loading button with the spinner.

screenshot_2019-02-14_15-53-10

Oh, and the error returned from server now should contain more information.

screenshot_2019-02-14_14-07-16

:tada:

osma:
Thanks @kinow , this is awesome! Will take it for a spin. All the changes you describe sound reasonable to me.

annif/templates/home.html
osma: I've always thought of the Annif UI as a way for the administrator to test projects/models, perhaps also do tasks like setting up projects and training them, but not something you want to expose to the world. So I think enabling the devtools shouldn't do any harm.

osma:
Seems to work well, so I will merge this. Thanks again!

Issue #262, opened 2019-03-06
Character encoding (Windows)


Veldhoen:
Python uses default system character encoding in calls to open(..). Under my Windows installation, this is cp1252, causing 'annif loadvoc' to fail with encoding errors. 
I created a pull request where I explicitly added the encoding (utf-8) in every call to open(..)

osma:
Thanks for the issue report and the PR! Will look into this. I've only used Annif on Linux (using a UTF-8 locale) so haven't encountered this problem so far...

PR #263, opened 2019-03-06
Explicit character encoding for all calls to open()

Veldhoen:
Issue #262 - Character encoding (Windows)


osma:
Had to fix the indent to make the PEP8 check pass.
Merged, thanks a lot!

PR #271, opened 2019-04-16
Remove unused imports and declare dependencies used

kinow:
Today I worked on a different project and decided to use `click` for the first time from scratch. Knowing Annif used it, I came right here to look for the `setup.py` to see what was there.

Only found `click-log` (interesting!), but not `click`. Then decided to look what other transitive dependencies were being used in the project.

I think the second commit covers all the dependencies used in code and tests. Added them to the respective groups of dependencies.

Cheers
Bruno

### DIFF
--- a/annif/backend/__init__.py
+++ b/annif/backend/__init__.py
@@ -1,7 +1,5 @@
 """Registry of backend types for Annif"""
 
-import configparser
-from flask import current_app
 from . import dummy
 from . import ensemble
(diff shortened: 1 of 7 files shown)


kinow:
Oh, and to avoid the unused imports issue in the `__init__.py`, I exported the imported code via `__all__`.

osma:
Thanks @kinow, looks good!

PR #275, opened 2019-05-12
Fix warning for joblib in scikit

kinow:
Following the documentation and noticed the following warning:

```
/home/kinow/Development/python/workspace/Annif/venv/lib/python3.7/site-packages/sklearn/externals/joblib/__init__.py:15: DeprecationWarning: sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.
  warnings.warn(msg, category=DeprecationWarning)
```

Not seen when running `pytest`.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -18,6 +18,7 @@ connexion = {extras = ["swagger-ui"]}
 swagger-ui-bundle = "*"
 flask-cors = "*"
 click-log = "*"
+joblib = "*"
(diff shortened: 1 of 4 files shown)


osma:
Thanks @kinow!

PR #276, opened 2019-05-12
Avoid calling update with empty dicts

kinow:
Was passing some functions with a debugger reading the code and noticed that here when the dict was not `None`, it could still be empty. And the `update` method was being called unnecessarily.

Won't cause any speed up, but still worth changing it I think?


osma:
Okay, why not. Thanks!

Issue #278, opened 2019-05-16
Dockerize Annif


juhoinkinen:
Make docker image containing Annif, upload to e.g. Dockerhub or Portainer (private repository).

osma:
I think this is done now and the issue can be closed? @juhoinkinen ?
Do we still need to add documentation somewhere, or improve the way images are built and/or distributed?

juhoinkinen:
I added the [Usage with Docker](https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker) page to the TOC of Wiki. 

Also added a section on creating&using the data image: https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker#using-annif-with-gunicorn-nginx-and-maui-backend

Closing this now.

PR #279, opened 2019-05-21
Issue278 dockerize annif

juhoinkinen:
Different Dockerfiles for creating images.
Based on python:3.6-slim (138 MB):
* Annif with all backends (Dockerfile, 1.19GB)
* Annif with only TFIDF backend (Dockerfile-slim, 653MB)
* optional dependencies/backends for Annif (Dockerfile-optional-dependencies, 1.19GB)

Based on python:3.6-alpine (79.1MB):
* Annif with only TFIDF backend (Dockerfile-alpine, already 1.24GB, and building takes ~15 min)

Alpine approach seems quite tricky regarding all backends and size increase (and the alpine base image is only 60 MB smaller than python-slim). Maybe best is to proceed on python:3.6-slim based images.

Tests pass in all containers. 

There was a problem using Vowpal Wabbit (related to libboost-python-dev, which has dependency on Python2 and also wrongly linked to it). Using VW version 8.4 solves this.

Few TODOs in Dockerfiles for shrinking image size.

Usage of containers is e.g. like this ~(or, actually, maybe the container doesn't need to be running all the time...)~:

1. Start bash in a container running in background (bind-mounting vocabylary and training data):
docker run -v /home/local/jmminkin/data:/data -itd --name annif_container annif bash

2. Execute commands in the already running container:
docker exec annif_container annif
docker exec annif_container annif loadvoc tfidf-en /data/vocab/yso-en.tsv
docker exec annif_container annif train tfidf-en /data/training/yso-finna-en.tsv.gz
echo This is the sample text to run suggest test on. | docker exec -i annif_container annif suggest tfidf-en


osma:
Excellent!

In the end we probably want a single Docker image which contains all optional dependencies. The TFIDF-only variant is only useful as an intermediate step, it's unlikely to be useful on its own.

Basing on python3.6-slim seems like the best choice. With Alpine it's going to be difficult to install all the dependencies, and the image size is already huge without them for some reason.

I think we also want to include a WSGI server in the image, for example uwsgi or gunicorn, so that we can run a production web server. Flask includes a simple HTTP server (`annif run` starts it) but it's intended for development only. Here's a [tutorial](https://medium.com/@smirnov.am/running-flask-in-production-with-docker-1932c88f14d0) for uwsgi. We probably need nginx too though I'm not 100% sure since I've only used Annif via WSGI with Apache mod_wsgi.

We also need documentation in the wiki for how to run Annif via Docker, something similar to [this page](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker) on the Skosmos wiki.

@kinow any comments?

(closes #278)

kinow:
@osma I saw this commit this morning and was really happy as this would make it easier for some devs - myself included - to run and contribute to Annif.

I noticed the several images, which is not a problem, if the intention is to allow users to choose the flavour, and use the image not only for development, but also in production - the Skosmos one was created with development in mind, but with some polishing should be able to run on Amazon, Docker Swarm, etc.

Only comment I have is that it would be good if the documentation included how to map the volumes needed for Annif. It would be really great if I could have all the corpora and configuration files needed in some volume, map to one or multiple containers, and be able to switch branches while developing, re-using the Docker files without destroying the data. :tada:

osma:
Thanks for chiming in @kinow!

I think we want to end up with one or maybe two Dockerfiles, not more, but it's good to test different variations first.

I agree that at least for development, the configuration and data files should live outside the Docker image, or at least that should be the default option. For some deployment scenarios it might actually be desirable to include the config and data (pretrained models) within the container, but I suppose we could support both.

juhoinkinen:
After some problems with setting file permissions, the Dockerization of Annif as a command-line tool seems finished (we tested this with @monalehtinen). 

The Wikipage (not yet in table-of-contents) describes use with bind-mounting a directory for storing data and config file: https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker

Should the image be uploaded to Docker Hub (as stated in the Wiki now, needs to create account for that; could be an organization account for NatLib), or should the image be accessed only by building it by youself?

I'll look at the WSGI server & nginx needs next (I'm not too familiar with them at present).

osma:
> Should the image be uploaded to Docker Hub (as stated in the Wiki now, needs to create account for that; could be an organization account for NatLib), or should the image be accessed only by building it by youself?

The goal is to publish the image on Docker Hub - perhaps by the time of the 0.41 release. For now it's okay that you have to build it yourself.

osma:
The current Dockerfile does everything in a single image - installs all the packages required for building e.g. fastText and Vowpal Wabbit, then removes as much as possible. I think a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) could simplify this. 

We could first build a "builder" image, use that to build and install the native code extensions, then start over with a clean image (no need to clean up afterwards) and copy over the built Python packages (from /usr/local/lib/python3.x/) to the new image without having to incllude the dependencies which are only necessary for compiling.

osma:
I tested this and successfully built and executed the image. The size is now 730MB. I suspect it might be reduced a bit using the multistage build approach I suggested above, but not more than by maybe 100MB or so.

I think we could rely on the standard nginx image for the proxying functionality. It is optional in any case - only necessary when you want to expose Annif to the Internet, but not necessary for local development and testing.

What we could do is to provide a `nginx.conf` file with the necessary proxy rules and a `docker-compose.yml` specifying that the standard nginx image is used, but overriding the `nginx.conf` file. [This tutorial](https://dev.to/domysee/setting-up-a-reverse-proxy-with-nginx-and-docker-compose-29jg) seems like a good starting point. There should be no need for the current `annif/nginx/Dockerfile` in my understanding.

juhoinkinen:
> I think we could rely on the standard nginx image for the proxying functionality. It is optional in any case - only necessary when you want to expose Annif to the Internet, but not necessary for local development and testing.
> 
> What we could do is to provide a `nginx.conf` file with the necessary proxy rules and a `docker-compose.yml` specifying that the standard nginx image is used, but overriding the `nginx.conf` file. [This tutorial](https://dev.to/domysee/setting-up-a-reverse-proxy-with-nginx-and-docker-compose-29jg) seems like a good starting point. There should be no need for the current `annif/nginx/Dockerfile` in my understanding.

Yes, the `annif/nginx/Dockerfile` is unnecessary, actually I had already removed it and overriden the default `nginx.conf` as in tutorial (only pushed this now).

juhoinkinen:
> The current Dockerfile does everything in a single image - installs all the packages required for building e.g. fastText and Vowpal Wabbit, then removes as much as possible. I think a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) could simplify this.
> 
> We could first build a "builder" image, use that to build and install the native code extensions, then start over with a clean image (no need to clean up afterwards) and copy over the built Python packages (from /usr/local/lib/python3.x/) to the new image without having to incllude the dependencies which are only necessary for compiling.

Used this approach, reduced image size 730MB -> 652MB

natlibfi-arlehiko:
Regarding the image size, here's a few recommendations:
- Use only one `RUN` instruction in Dockerfiles to prevent creation of multiple layers. Or better yet, use multi-stage builds and copy the relevant files from earlier build stages
- Run Swagger UI in separate container
- Do the actual training in separately and make the actual Annif container use the resulting data

- Some of the Python dependencies (Which I image are required for the training seem to be pretty large:
```sh
$ docker run -it --rm -u root annif sh -c 'du --si -s /usr/local/lib/python3.6/site-packages/*|sort -h|tail'
8.2M    /usr/local/lib/python3.6/site-packages/pip
11M     /usr/local/lib/python3.6/site-packages/Cython
12M     /usr/local/lib/python3.6/site-packages/boto
12M     /usr/local/lib/python3.6/site-packages/nltk
28M     /usr/local/lib/python3.6/site-packages/sklearn
39M     /usr/local/lib/python3.6/site-packages/swagger_ui_bundle
43M     /usr/local/lib/python3.6/site-packages/botocore
46M     /usr/local/lib/python3.6/site-packages/gensim
73M     /usr/local/lib/python3.6/site-packages/numpy
92M     /usr/local/lib/python3.6/site-packages/scipy
```

juhoinkinen:
>     * Some of the Python dependencies (Which I image are required for the training) seem to be pretty large:
> 
> ```shell
> $ docker run -it --rm -u root annif sh -c 'du --si -s /usr/local/lib/python3.6/site-packages/*|sort -h|tail'
> 8.2M    /usr/local/lib/python3.6/site-packages/pip
> 11M     /usr/local/lib/python3.6/site-packages/Cython
> 12M     /usr/local/lib/python3.6/site-packages/boto
> 12M     /usr/local/lib/python3.6/site-packages/nltk
> 28M     /usr/local/lib/python3.6/site-packages/sklearn
> 39M     /usr/local/lib/python3.6/site-packages/swagger_ui_bundle
> 43M     /usr/local/lib/python3.6/site-packages/botocore
> 46M     /usr/local/lib/python3.6/site-packages/gensim
> 73M     /usr/local/lib/python3.6/site-packages/numpy
> 92M     /usr/local/lib/python3.6/site-packages/scipy
> ```

Of these `boto` and `botocore` are for accessing Amazon Web Services, and at least they have no role in Annif, but are subdependencies of `gensim` via `smart-open`:

```
gensim==3.7.3
    ...
    smart-open==1.8.4
      boto==2.49.0
      boto3==1.9.161
        botocore==1.12.161
          docutils==0.14
          jmespath==0.9.4
          python-dateutil==2.8.0
            six==1.12.0
          urllib3==1.25.3
        jmespath==0.9.4
        s3transfer==0.2.1
          botocore==1.12.161
            docutils==0.14
            jmespath==0.9.4
            python-dateutil==2.8.0
              six==1.12.0
            urllib3==1.25.3
```
Unfortunately, for some reason even `import gensim.similarities` tries eventually to import `boto`, so it doesn't seem too straightforward to get rid of even `boto`s.

osma:
Thank you @natlibfi-arlehiko for your comments. Some responses:

> Use only one RUN instruction in Dockerfiles to prevent creation of multiple layers. Or better yet, use multi-stage builds and copy the relevant files from earlier build stages

The current Dockerfile uses a multi-stage build already.

> Run Swagger UI in a separate container

Swagger UI is built in to Connexion, which is a core component of Annif. I don't see how to easily move it to a separate container.

> Do the actual training separately and make the actual Annif container use the resulting data

This is the current plan. The main container only includes the software itself, not the data which is mounted separately.

> Some of the Python dependencies (Which I image are required for the training) seem to be pretty large:

Unfortunately it seems difficult to get rid of any of these, @juhoinkinen already tried excising the most useless ones, without success.

@juhoinkinen Do you think this is starting to be ready for merging? The PR is still marked as draft but I don't see any major actionable TODOs in the recent discussion.

natlibfi-arlehiko:
> Swagger UI is built in to Connexion, which is a core component of Annif. I don't see how to easily move it to a separate container.

The way we do it in [Melinda's record import API](https://github.com/NatLibFi/melinda-record-import-api):
- The REST API serves OpenAPI JSON with `GET /`
- Run a separate Swagger UI container that fetches the OpenAPI spec from the previously mentioned REST API

Might be [possible with connexion too](https://connexion.readthedocs.io/en/latest/routing.html#swagger-json).

osma:
@natlibfi-arlehiko I see. That's good from the perspective of separating concerns between containers, but I don't think it will reduce the container size, since Connexion will include Swagger UI regardless of whether we enable it or not. Did I understand correctly?

natlibfi-arlehiko:
> but I don't think it will reduce the container size, since Connexion will include Swagger UI regardless of whether we enable it or not. Did I understand correctly?

Yeah. Good point :D

juhoinkinen:
> @juhoinkinen Do you think this is starting to be ready for merging? The PR is still marked as draft but I don't see any major actionable TODOs in the recent discussion.

Yes, I just updated the wiki https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker about the docker-compose to combine Annif+NGINX+Mauiservice (before there were two docker-compose files, one for Annif+NGINX and other for Annif+Mauiservice). The current `docker-compose.yml` is now like a template, from where unnecessary services can be removed if wanted. 

There is also `docker-compose-portainer.yml`, but that is used to deploy the services in our own platform/instance, and there is still work to be done, but that maybe is not part of this pull request.

osma:
Looks good now. My only concern is that the Drone build is failing according to the above status display. I can see there is this error at the end of the Drone `docker` log:

```
unauthorized: access to the requested resource is not authorized
time="2019-06-20T13:43:08Z" level=fatal msg="exit status 1" 
```

Do you need to add some kind of credentials somewhere?

natlibfi-arlehiko:
> Looks good now. My only concern is that the Drone build is failing according to the above status display. I can see there is this error at the end of the Drone `docker` log:
> 
> ```
> unauthorized: access to the requested resource is not authorized
> time="2019-06-20T13:43:08Z" level=fatal msg="exit status 1" 
> ```
> 
> Do you need to add some kind of credentials somewhere?

Fixed.

PR #285, opened 2019-06-25
Drop requirements files, attempt to run "setup.py install" in rtd

juhoinkinen:
Remove unnecessary ´requirements` files and a package for their creation.


osma:
Is this working properly? Can it be merged now?

juhoinkinen:
Yes, now this is working and mergable. There were some randomish problems about builds failing, but adding the `.readthedocs.yml` seems to solve those (using the same settings via RTD webpage worked from time to time). 

For future reference: [Wiping build environment in RTD](https://docs.readthedocs.io/en/stable/guides/wipe-environment.html#wiping-a-build-environment) can help to solve some problems

(Also, the `.rst` files can be removed from version control as they can automatically generated: #287.)

PR #287, opened 2019-06-28
Auto generate documentation include sphinx apidoc run

juhoinkinen:
This allows running `sphinx-apidoc` automatically when `sphinx-build` runs in RTD, thus there is no need to the `.rst` files `docs/source/`.


osma:
#285 was just merged. Should this be adjusted accordingly? #285 dropped the `requirements.txt` file, but this PR seems to be changing it...

juhoinkinen:
`sphinxcontrib-apidoc` needed to be installed, but adding it to `setup.py` seemed unnecessary(?). It's now in specific file in `docs/requirements.txt`. 

Not optimal, but hopefully `Pipfile` support is coming soon in RTD, so the whole build dependency handling becomes more simple via it.

osma:
Much better now! I don't see a problem having a single-line `docs/requirements.txt` file specifically for RTD, since it doesn't duplicate dependency information in `setup.py` and `Pipfile`.

Issue #250, opened 2019-01-30
Support --backend-param option in train command


osma:
The `--backend-param` option can be used in analyze and eval commands but is not currently supported for the train command. It would be useful to be able to override e.g. learn rate for fastText and vw during train time.

juhoinkinen:
Closed in #289.

Issue #273, opened 2019-04-25
Default values for configuration settings


osma:
Currently some configuration settings, especially `limit`, must be set, otherwise Annif will raise an exception:

```
  File "/home/ozone/git/Annif/annif/cli.py", line 165, in run_suggest
    hits = hit_filter(project.suggest(text, backend_params))
  File "/home/ozone/git/Annif/annif/project.py", line 175, in suggest
    hits = self._suggest_with_backend(text, backend_params)
  File "/home/ozone/git/Annif/annif/project.py", line 115, in _suggest_with_backend
    hits = self.backend.suggest(text, project=self, params=beparams)
  File "/home/ozone/git/Annif/annif/backend/backend.py", line 44, in suggest
    return self._suggest(text, project, params=beparams)
  File "/home/ozone/git/Annif/annif/backend/tfidf.py", line 55, in _suggest
    return fullresult.filter(limit=int(self.params['limit']))
  File "/usr/lib/python3.6/configparser.py", line 1233, in __getitem__
    raise KeyError(key)
KeyError: 'limit'
```

This is unfriendly to the user. There should be sane defaults for all settings. `limit` should default to 100, project name should default to its id, chunksize should default to 0 (no chunking) etc.

juhoinkinen:
This was covered in #324 and made to release 0.43.

Issue #240, opened 2019-01-21
Hyperparameter optimization with hyperopt


osma:
Some backends require preselected (hyper)parameters and these must be carefully chosen to get the best results. Often the only way is trial and error. Especially fastText is notorious for requiring a large number of parameters and finding the magic combination can be a challenge.

The [hyperopt](https://github.com/hyperopt/hyperopt) library implements an algorithm for efficiently searching through the hyperparameter space in order to find a good combination. We could implement a command like `annif hyperopt <project> <corpus>` that looks for good hyperparameters.

osma:
Here is a snippet I've used previously to find suitable fastText hyperparameters. It requires training and validation files in the fastText format.

```python
from hyperopt import fmin, tpe, hp, Trials
import sys
import fastText

trainfile = sys.argv[1]
validfile = sys.argv[2]
evals = int(sys.argv[3])

trials = Trials()

# precision@K value
K_VALUE = 5

def eval(space):
    print("Trial {} of {}".format(len(trials), evals))
    print("Testing values:", space)
    model = fastText.train_supervised(
        input=trainfile,
        dim=int(space['dim']),
        lr=space['lr'],
        epoch=int(space['epoch']),
        minCount=int(space['minCount']),
        wordNgrams=int(space['ngrams']),
        minn=int(space['minn']),
        maxn=int(space['maxn']),
        loss='hs'
    )
    n, p, r = model.test(validfile, k=K_VALUE)
    print("Precision:", p)
    print()
    return 1-p

space = {
    'dim': hp.quniform('dim', 100, 500, 10),
    'lr': hp.loguniform('lr', -0.4, -0.1),
    'epoch': hp.quniform('epoch', 20, 120, 5),
    'minCount': hp.quniform('minCount', 1, 5, 1),
    'ngrams': hp.choice('ngrams', [1, 2]),
    'minn': hp.quniform('minn', 3, 5, 1),
    'maxn': hp.quniform('maxn', 5, 7, 1)
}

best = fmin(
    fn=eval,
    space=space,
    algo=tpe.suggest,
    max_evals=evals,
    trials=trials
)

print("Best parameters:", best)
print("Best precision:", 1 - trials.best_trial['result']['loss'])
```

mo-fu:
Two other alternative libraries for hyper parameter optimization:
* https://github.com/dragonfly/dragonfly
*  https://github.com/automl/SMAC3

osma:
[...]

PR #289, opened 2019-07-02
Issue250 support backend param option in train and learn commands

juhoinkinen:
This closes issue #250.


juhoinkinen:
A few question on the way backend parameters are handled.

Here https://github.com/NatLibFi/Annif/blob/master/annif/backend/backend.py#L42 the `beparams` are  first taken from the object's attribute, then updated if there are `params` coming from CLI option, and finally passed in `_suggest` call. However,  for `limit` at least TFIDF and fastText backends use the original `params` attribute, not the passed `params` variable: 
https://github.com/NatLibFi/Annif/blob/master/annif/backend/tfidf.py#L54
https://github.com/NatLibFi/Annif/blob/master/annif/backend/fasttext.py#L116

However, the `chunksize` is taken from passed `params`:
https://github.com/NatLibFi/Annif/blob/master/annif/backend/mixins.py#L23

Is the intention that `limit` should not be overriden by `--backend-param`, or can that be changed to come from the passed `params`? 

Overall, is passing `params` variable necessary, as updating the object's `params` attribute would seem simpler? E.g. at line https://github.com/NatLibFi/Annif/blob/master/annif/backend/backend.py#L44

osma:
You found some genuine inconsistencies there :)

A few thoughts about how parameter handling in backends should be done, and a comparison with the current (somewhat broken) situation.

* in backend classes, `self.params` should contain the parameters defined in `projects.cfg` (as is already the case) - perhaps this field could be renamed to something a bit more specific, e.g. `self.config_params`
* `suggest`, `train` and `learn` methods should take a `params` attribute (only `suggest` has it currently); this would be used to pass parameters set via CLI options
* the backends should not use `self.params` (or `self.config_params` if renamed) directly; instead, the passed `params` attribute should override parameters from the config file (as is done [here](https://github.com/NatLibFi/Annif/blob/master/annif/backend/backend.py#L42) - but perhaps there could be a helper method that does this)
* backend code should not assume that a particular parameter always exists; for example, `params['limit']` should be changed to `params.get('limit', self.DEFAULT_LIMIT)`; see #273 for some thoughts on this 

More generally, I think it would be helpful if backend classes explicitly defined the backend-specific (hyper)parameters they need. This information could be used e.g. for hyperparameter optimization (see #240). [fasttext](https://github.com/NatLibFi/Annif/blob/master/annif/backend/fasttext.py#L19), [vw_multi](https://github.com/NatLibFi/Annif/blob/master/annif/backend/vw_multi.py#L19) and [vw_ensemble](https://github.com/NatLibFi/Annif/blob/master/annif/backend/vw_ensemble.py#L26) already sort of do this in their own way, but the parameter definitions should be standardized and probably hyperopt would need more information about the parameters - for example the minimum and maximum values. This can be left for later, but it's probably useful to keep in mind when parameter handling code is modified.

juhoinkinen:
Usage example (with debugging to show passed param values): 
```
$ annif train fasttext-fi tests/corpora/archaeology/documents_10_lines.tsv -b fasttext.dim=1 --backend-param fasttext.lr=42.1 --backend-param fasttext.epoch=0 -v DEBUG
debug: creating app with configuration annif.default_config.Config
debug: CLI option overriding parameter for backend fasttext: dim=1
debug: CLI option overriding parameter for backend fasttext: lr=42.1
debug: CLI option overriding parameter for backend fasttext: epoch=0
...
Backend fasttext: creating fastText model
debug: Backend fasttext: Model parameters: {'dim': 1, 'lr': 42.1, 'epoch': 0, 'loss': 'hs'}
Read 0M words
Number of words:  56
Number of labels: 15
Progress: 100.0% words/sec/thread:       0 lr:  0.000000 loss:  0.000000 ETA:   0h 0m
```

This PR introduces also the following:
- Simple validation of the passed backend-params, for example: 
```
$ annif train fasttext-fi tests/corpora/archaeology/documents_10_lines.tsv -b tfidf.dim=1 
Error: Error: The backend tfidf in CLI option "-b tfidf.dim=1" not matching the project backend fasttext.
```

- Raising `NotSupportedException` when trying to train ensemble backend:
```
$ annif train ensemble-fi ~/annif-projects/Annif-corpora/training/yso-finna-fi.tsv.gz 
Error: Error: Training ensemble model is not possible.
```

osma:
Let's get back to this after PR #322 has been merged. Then the changes from `master` need to be merged into this branch.

juhoinkinen:
> This pull request **introduces 1 alert** when merging [ba9edb8](https://github.com/NatLibFi/Annif/commit/ba9edb81034d45fea91a78484f37d24ab6547283) into [06f5221](https://github.com/NatLibFi/Annif/commit/06f52218b10e8bc43847a7ddd2ee4e9590add70d) - [view on LGTM.com](https://lgtm.com/projects/g/NatLibFi/Annif/rev/pr-093a8c9792de25d88b8b26afbb766fa34d7be749)
> 
> **new alerts:**
> 
>     * 1 for Non-iterable used in for loop

This alert seems like [the reported false positive](https://github.com/Semmle/ql/issues/2137) (the fix for that is merged to Semmle/ql master, but I could not find out which version of Semmle is run in lgtm.com).

PR #292, opened 2019-07-05
Issue290 upgrade fasttext

### LINKED ISSUE #290: Upgrade fastText back to official PyPI version
We've been using [fasttextmirror](https://pypi.org/project/fasttextmirror/) because the official PyPI [fasttext](https://pypi.org/project/fasttext/) package was not being maintained for a long time.

But now there is finally a new version (0.9.1) available. We should switch over to that.

mvsjober:
Upgrade fastText back to official PyPI version, and enable the pretrained vectors feature.

Closes #290


osma:
Thanks for the PR!

It seems that the fastText tests are not running under Travis CI, which leads to low test coverage. I'm guessing that the reason is that in `.travis.yml`, `fasttextmirror` is [still being installed](https://github.com/NatLibFi/Annif/blob/master/.travis.yml#L26). Can you try changing that too?

mvsjober:
I fixed the installation in `.travis.yml`. BTW, also this page needs to be updated in the wiki: https://github.com/NatLibFi/Annif/wiki/Optional-features-and-dependencies#fasttext-backend - what is the process for that?

osma:
Thanks, much better now! The tests seem to be running fine after the change.

The wiki page can simply be edited by anyone, including you. There is a bit of a synchronization challenge here - ideally the wiki would be changed at the same time the PR is merged. Perhaps we could simply drop the alternative recipe (`pip install fasttextmirror` or `pip install fasttext`) and recommend `pip install annif[fasttext]` instead - I think it should do the right thing (install either `fasttextmirror` or `fasttext`, depending on which version of Annif has been installed) but I'm a little bit unsure whether it uses the dependency information from the locally installed Annif or the one available on PyPI, which could be an older or newer version.

Do you think it would be easy to add a unit test for the pretrained vectors feature? It would require adding a file with some pretrained vectors and then adding a test (in `tests/test_backend_fasttext.py`) that uses it. There is some toy data under `tests/corpora/archaeology` that could perhaps be used to create the vectors. Preferably the vectors file should be small (at most a few MB) so it doesn't bloat the repo. The idea would be to verify that it loads without error - quality of classification results is not important.

mvsjober:
I thought about making a test but dismissed the idea as I would have to add a huge file somehow, but a toy data test I can definitely do.

mvsjober:
There is still one problem that needs to be addressed before we can merge.  With some test sets I get a ValueError:

```
  File "/homeappl/home/mvsjober/appl_taito/Annif/annif/backend/fasttext.py", line 119, in _suggest_chunks
    chunktexts, project, limit)
  File "/homeappl/home/mvsjober/appl_taito/Annif/annif/backend/fasttext.py", line 114, in _predict_chunks
    for chunktext in chunktexts])), limit)
  File "/homeappl/home/mvsjober/.local/share/virtualenvs/Annif-YopBXEME/lib/python3.5/site-packages/fasttext/FastText.py", line 151, in predict
    result_as_pair = np.array(predictions, dtype=dt)
ValueError: setting an array element with a sequence
```

This appears to be a bug in the upstream fastText module.  The line `result_as_pair = np.array(predictions, dtype=dt)` expects predictions to be a list of of predictions for each text, each item should be another list of _equal_ length.  However as we pass both k=1000 and threshold=0.0 to multilinePredict() it can happen that it sometimes returns less that k=1000 items if there are less than k over the threshold.

mvsjober:
Bug reported to fastText: https://github.com/facebookresearch/fastText/issues/846

mvsjober:
Seems the bug has now been fixed upstream, but we should probably wait until there's an official release of fastText with the bugfix included.

osma:
Agree that we should wait for a PyPI release.
There are some merge conflicts that need to be fixed. I bet these are related to pinning specific versions of libraries. If that's the case they should be trivial to fix.

osma:
fastText 0.9.2 has been released on PyPI: https://pypi.org/project/fasttext/

Meanwhile some merge conflicts have emerged, I will try to sort them out so that we can merge this for the next release 0.48.

osma:
I couldn't figure out how to update this PR. I tried to push to the original PR branch on CSCfi/Annif but got a "permission denied" error, even though the PR seems to allow pushing by upstream users. Anyway, I created a new PR #409 instead.

Issue #291, opened 2019-07-03
Pin down more specific versions of dependencies


osma:
I'm using an Annif installation (on the annif-kk machine) that was set up some months ago. When I use `tfidf` models I get errors like this:

```
/srv/Annif/.venv/lib/python3.6/site-packages/sklearn/externals/joblib/__init__.py:15: DeprecationWarning: sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.
/srv/Annif/.venv/lib/python3.6/site-packages/sklearn/base.py:306: UserWarning: Trying to unpickle estimator TfidfTransformer from version 0.20.0 when using version 0.21.2. This might lead to breaking code or invalid results. Use at your own risk.
/srv/Annif/.venv/lib/python3.6/site-packages/sklearn/base.py:306: UserWarning: Trying to unpickle estimator TfidfVectorizer from version 0.20.0 when using version 0.21.2. This might lead to breaking code or invalid results. Use at your own risk.
```

The problem here is that the model (including transformer and vectorizer) was built and saved using an older version of sklearn. Similar problems could happen when gensim, fasttext or vw are updated, because old serialized models may not work properly with the new version.

I think we should pin down the versions of dependencies a bit more precisely. I don't want to depend on specific versions (at least not yet, Annif is still in 0.x) but I think that for the libraries whose data structures are being serialized to disk we should be a bit more conservative, otherwise models may stop working unexpectedly. In the current situation, I think it would make sense to

- pin sklearn to 0.21.x
- pin fasttext(mirror) to the current available minor version (see #290)
- pin gensim to 3.7.x
- pin vowpalwabbit probably to ~~8.5.x~~ 8.7.x

osma:
In future releases, we could then do a more controlled upgrade to newer versions of these dependencies.

juhoinkinen:
Vowpal Wabbit version [`8.7.0.post1`](https://github.com/VowpalWabbit/vowpal_wabbit/releases/tag/8.7.0)  was released 12.7.2019 in Pypi. Should that be used instead of 8.5? The new version needs some new dependencies, see https://github.com/NatLibFi/Annif/pull/296

osma:
Yes, we should use VW 8.7 if there are no known issues with that version.
[...]

PR #296, opened 2019-07-16
Use vw version 8.7

juhoinkinen:
Vowpalwabbit version [`8.7.0.post1`](https://github.com/VowpalWabbit/vowpal_wabbit/releases/tag/8.7.0) was released 12.7.2019 in Pypi. It requires a few new dependencies to be installed, and change on the linking fix of libboost. The [VW wikipage](https://github.com/NatLibFi/Annif/wiki/Optional-features-and-dependencies) has been updated according to these.

However, issue #291 states VW should be pinned to 8.5. Is that still valid now when newer version is released (note that 8.6 is not in Pypi)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ addons:
     - libboost-program-options-dev
     - libboost-python-dev
     - zlib1g-dev
+    - cmake
+    - libboost-system-dev
+    - libboost-thread-dev
+    - libboost-test-dev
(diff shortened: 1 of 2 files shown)


osma:
Looks good to me. I think the Travis Python 3.5 build for PR #300 is currently failing because of a VW version mismatch so it would be a good idea to merge this PR first.

PR #299, opened 2019-07-17
Issue281 handle invalid lines in training data tsv file

### LINKED ISSUE #281: Validate TSV training input
Maybe Annif could check the input tsv files for trivial flaws such as excessive newlines not containing any tab character and give out a warning. Right now you receive an error from some Python method instead which might not be very transparent for users that are not proficient in Python (and even for those who are).

juhoinkinen:
1. Handle lines that are missing tab in training data `.tsv` by giving warning for such line and continuing:
```
$ annif train tfidf-en trainingdata_with_invalid_lines.tsv 
creating vectorizer
warning: Skipping invalid line (missing tab): ""
warning: Skipping invalid line (missing tab): "A line without tabs"
Backend tfidf: creating similarity index
```

~~2. Hide full traceback for missing file:~~

Edit: missing file case is handled by issue #318.

This closes #281.


annif/corpus/document.py
```diff
@@ -60,13 +63,26 @@ def opener(path):
...
+            raise AnnifException(str(err))
+
+    @staticmethod
+    def _parse_tsv_line(line):
+        try:
+            text, uris = line.split('\t', maxsplit=1)
+            subjects = [annif.util.cleanup_uri(uri)
+                        for uri in uris.split()]
+            yield Document(text=text, uris=subjects, labels=[])
+        except ValueError as err:
```
osma: Two things I dislike in this `try..except` block:

1. It contains 3 lines of code, involving several calls to external code, but the only actually covered problem happens on the first line. The two other lines should be moved outside the block.
2. Testing the specific type of exception using `'not enough values to unpack' in str(err)` seems unnecessarily fragile. A future Python version could change the exception message.

Would it be better to use a test like `if '\t' in line` instead of relying on ValueErrors to catch the problematic case?

tests/test_corpus.py
```diff
@@ -201,6 +203,33 @@ def test_docfile_plain(tmpdir):
...
+def test_docfile_plain_invalid_lines(tmpdir, caplog, capsys):
+    docfile = tmpdir.join('documents_invalid.tsv')
+    docfile.write("""Läntinen\t<http://www.yso.fi/onto/yso/p2557>
+
+        Oulunlinnan\t<http://www.yso.fi/onto/yso/p7346>
+        A line with no tabs
+        Harald Hirmuinen\t<http://www.yso.fi/onto/yso/p6479>""")
+    docs = annif.corpus.DocumentFile(str(docfile))
+    assert len(list(docs.documents)) == 3
+    # Need to capture both logs and stderr: when running this test individually
```
osma: Would it be cleaner to use [pytest-capturelog](https://pypi.org/project/pytest-capturelog/) to catch and check the log messages?

annif/corpus/document.py
juhoinkinen: Yes, should have thought about those points! Now added the if condition

tests/test_corpus.py
juhoinkinen: The `capsys` was used because when full test suite was run, the line [`click_log.basic_config(logger)`](https://github.com/NatLibFi/Annif/blob/master/annif/cli.py#L21) made some changes to the logger, which in turn made the warnings intended to be captured in this test go to `stderr` and capturing them with `caplog` failed: 
```
------------------------------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------------------------------
warning: Skipping invalid line (missing tab): ""
warning: Skipping invalid line (missing tab): "        A line with no tabs"
```

The change due to that line in `logger.__dict__` is from 
`{'filters': [], 'name': 'annif', 'level': 0, 'parent': <RootLogger root (WARNING)>, 'propagate': True, 'handlers': [], 'disabled': False, 'manager': <logging.Manager object at 0x7fa66c0e9860>}`
to 
`{'filters': [], 'name': 'annif', 'level': 0, 'parent': <RootLogger root (WARNING)>, 'propagate': False, 'handlers': [<ClickHandler (NOTSET)>], 'disabled': False, 'manager': <logging.Manager object at 0x7fa66c0e9860>}`

By switching `'propagate': False` back to `True` makes the log capture [work in this test](https://github.com/NatLibFi/Annif/pull/299/commits/9041165f67de5a964b7004f1e1427ded63f7bf41). But I'm not so sure if it should be touched for this purpose, or would there be some way to test this?

juhoinkinen: Discussion on [Handling of loggers with propagate=False](https://github.com/pytest-dev/pytest/issues/3697).

annif/corpus/document.py
```diff
@@ -61,15 +64,24 @@ def opener(path):
...
-                            for uri in uris.split()]
-                yield self._create_document(text=text,
-                                            uris=subjects,
-                                            labels=[])
+        try:
+            with opener(self.path) as tsvfile:
+                for line in tsvfile:
+                    yield from self._parse_tsv_line(line)
+        except FileNotFoundError as err:
+            raise AnnifException(str(err))
```
osma: I think it would be better to define a new AnnifException subclass for this purpose. AnnifException is meant to be a common superclass of all Annif exceptions, not to be directly instantiated. (In fact it could, and maybe should, be turned into an abstract base class, which would prevent it being used like this)

juhoinkinen: Would it be actually better to let Click handle the nonexistent paths in issue #318 and remove this try block?

osma: Yes, that's better!

Issue #148, opened 2018-05-23
Initialization failure under mod_wsgi


osma:
I get this error when running under mod_wsgi:

```
 Traceback (most recent call last):
   File "/var/www/Annif/annif.wsgi", line 6, in <module>
     application = create_app()
   File "/opt/Annif/annif/__init__.py", line 30, in create_app
     annif.project.initialize_projects(cxapp.app)
   File "/opt/Annif/annif/project.py", line 223, in initialize_projects
     projects_file, datadir, init_projects)
   File "/opt/Annif/annif/project.py", line 214, in _create_projects
     projects[project_id].initialize()
   File "/opt/Annif/annif/project.py", line 63, in initialize
     analyzer = self.analyzer
   File "/opt/Annif/annif/project.py", line 112, in analyzer
     self._analyzer = annif.analyzer.get_analyzer(self.analyzer_spec)
   File "/opt/Annif/annif/analyzer/__init__.py", line 15, in get_analyzer
     match = re.match(r'(\\w+)(\\((.*)\\))?', analyzerspec)
   File "/usr/lib/python3.5/re.py", line 163, in match
     return _compile(pattern, flags).match(string)
 TypeError: expected string or bytes-like object
```

osma:
Probably because there's a project that doesn't define an analyzer.

PR #300, opened 2019-07-17
Issue282 handling syntax errors in projects config

### LINKED ISSUE #282: Handling errors in projects.cfg
Some syntax errors of  `projects.cfg` produce a lengthy generic Python error message, see e.g. https://groups.google.com/d/msg/annif-users/kIvJPbZ-vVU/zXDYFm6PAgAJ 

A more concise/specific error message would be better. Some errors are already handled, for example this exists: 
`"Error: Project 'maui-fi': vocab setting is missing"`

juhoinkinen:
Raise `ConfigurationException` instead of full traceback in following situations:

1. when trying to use (train/suggest) a project without a backend set in `projects.cfg`:
    ```
    $ annif train nobackend -p tests/projects.cfg any_trainingdata.tsv 
    Error: Project 'nobackend': backend setting is missing
    ```
    This is in line with missing vocab error: ` "Error: Project 'maui-fi': vocab setting is missing"`.

2. when trying to use `projects.cfg` that has duplicated sections or entries:
    ```
    $ annif train any_backend -p tests/projects_invalid.cfg any_trainingdata.tsv 
    Error: Error: While reading from 'tests/projects_invalid.cfg' [line 10]: section 'duplicatedproject' already exists
    ```
This closes #282.

### DIFF
--- a/annif/project.py
+++ b/annif/project.py
@@ -127,7 +127,7 @@ class AnnifProject(DatadirMixin):
             if self.analyzer_spec:
                 self._analyzer = annif.analyzer.get_analyzer(
                     self.analyzer_spec)
-            elif self.backend.needs_subject_vectorizer:
+            else:
                 raise ConfigurationException(
                     "analyzer setting is missing (and needed by the backend)",
                     project_id=self.project_id)
--- a/annif/project.py
+++ b/annif/project.py
@@ -87,10 +87,10 @@ class AnnifProject(DatadirMixin):
 
     def _initialize_backend(self):
         logger.debug("Project '%s': initializing backend", self.project_id)
-        if not self.backend:
-            logger.debug("Cannot initialize backend: does not exist")
-            return
         try:
+            if not self.backend:
+                logger.debug("Cannot initialize backend: does not exist")
+                return
             self.backend.initialize()
         except AnnifException as err:
             logger.warning(err.format_message())
@@ -127,6 +127,9 @@ class AnnifProject(DatadirMixin):
     @property
     def backend(self):
         if self._backend is None:
+            if 'backend' not in self.config:
+                raise ConfigurationException(
+                    "backend setting is missing", project_id=self.project_id)
             backend_id = self.config['backend']
             try:
                 backend_class = annif.backend.get_backend(backend_id)
@@ -214,7 +217,7 @@ class AnnifProject(DatadirMixin):
         return {'project_id': self.project_id,
                 'name': self.name,
                 'language': self.language,
-                'backend': {'backend_id': self.config['backend']}
+                'backend': {'backend_id': self.config.get('backend')}
                 }
 
 
--- a/annif/project.py
+++ b/annif/project.py
@@ -233,7 +233,11 @@ def _create_projects(projects_file, datadir, init_projects):
     config = configparser.ConfigParser()
     config.optionxform = lambda option: option
     with open(projects_file, encoding='utf-8') as projf:
-        config.read_file(projf)
+        try:
+            config.read_file(projf)
+        except (configparser.DuplicateOptionError,
+                configparser.DuplicateSectionError) as err:
+            raise ConfigurationException(err)
 
     # create AnnifProject objects from the configuration file
     projects = collections.OrderedDict()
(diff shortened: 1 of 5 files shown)


osma:
Travis Python 3.5 build is failing, I think VW needs to be upgraded first. Maybe merge #296?

osma:
failing build log: https://travis-ci.org/NatLibFi/Annif/jobs/559957266

osma:
Looks good, but there is one further case that could be covered. If `analyzer` is left unspecified, I get this:

```
  File "/home/oisuomin/git/Annif/annif/project.py", line 188, in _create_vectorizer
    tokenizer=self.analyzer.tokenize_words)
AttributeError: 'NoneType' object has no attribute 'tokenize_words'
```

Should this case be covered too, or left for another issue/PR?

juhoinkinen:
> Looks good, but there is one further case that could be covered. If `analyzer` is left unspecified, I get this:
> 
> ```
>   File "/home/oisuomin/git/Annif/annif/project.py", line 188, in _create_vectorizer
>     tokenizer=self.analyzer.tokenize_words)
> AttributeError: 'NoneType' object has no attribute 'tokenize_words'
> ```
> 
> Should this case be covered too, or left for another issue/PR?

I'll check if this can be easily covered too and implement, otherwise I'll make a new issue about this.

juhoinkinen:
> Looks good, but there is one further case that could be covered. If `analyzer` is left unspecified, I get this:
> 
> ```
>   File "/home/oisuomin/git/Annif/annif/project.py", line 188, in _create_vectorizer
>     tokenizer=self.analyzer.tokenize_words)
> AttributeError: 'NoneType' object has no attribute 'tokenize_words'
> ```
> 
> Should this case be covered too, or left for another issue/PR?

Implemented raising `ConfigurationError` for this in the previous commit. For the test I changed the [`[noanalyzer]` project](https://github.com/NatLibFi/Annif/pull/300/commits/f03c320435fb1d460e99bd36144a4ac78a43e734#diff-e889023e3a656e930515ec0cb95ce5b1R36) to use TFIDF backend because dummy backend does not need analyzer and would not raise the `ConfigurationError`. That project was included for issue #148. I did not see any reason for the backend to remain `dummy`, but could @Osma confirm?

annif/project.py
```diff
@@ -120,8 +123,14 @@ def _suggest_with_backend(self, text, backend_params):
 
     @property
     def analyzer(self):
-        if self._analyzer is None and self.analyzer_spec:
-            self._analyzer = annif.analyzer.get_analyzer(self.analyzer_spec)
+        if self._analyzer is None:
+            if self.analyzer_spec:
+                self._analyzer = annif.analyzer.get_analyzer(
+                    self.analyzer_spec)
+            elif self.backend.needs_subject_vectorizer:
```
osma: I don't think this is the right condition to test, maybe not even the right place for it. It happens to work for the case where an analyzer is needed within the AnnifProject class, but there are other situations in backend-specific code where an analyzer is needed as well, irrespective of `needs_subject_vectorizer`:

* [fasttext.py](https://github.com/NatLibFi/Annif/blob/master/annif/backend/fasttext.py#L77)
* [mixins.py](https://github.com/NatLibFi/Annif/blob/master/annif/backend/mixins.py#L21)
* [vw_multi.py](https://github.com/NatLibFi/Annif/blob/master/annif/backend/vw_multi.py#L57)

Alternatively, backends could have a `needs_analyzer` field that indicates that they are making use of the analyzer, then the `if` clause above could check for that.

annif/project.py
```diff
@@ -62,10 +62,13 @@ def _init_access(self):
...
-        analyzer = self.analyzer
-        logger.debug("Project '%s': initialized analyzer: %s",
-                     self.project_id,
-                     str(analyzer))
+        try:
+            analyzer = self.analyzer
+            logger.debug("Project '%s': initialized analyzer: %s",
+                         self.project_id,
+                         str(analyzer))
+        except AnnifException as err:
```
osma: Can we avoid having to catch the exception here? Not sure it's a great idea that simply accessing `self.analyzer` may trigger an exception. Returning None is a bit more elegant, but then of course the missing analyzer case has to be handled by the calling code.

PR #305, opened 2019-08-07
Use data image from quay.io

juhoinkinen:


juhoinkinen:
Updated Wiki for pushing a data image to quay.io instead of uploading `.tar` of the image to Portainer.

Issue #269, opened 2019-04-05
Maui Server backend, including training support


osma:
The [Maui Server API](https://github.com/TopQuadrant/MauiServer/blob/master/API.md) provides all the important Maui functionality over REST. Annif should have a backend that works with this - either by enhancing the `http` backend or by creating a new `maui` backend. The latter is probably easier, though even the `http` backend could be quite easily made to support at least the basic analyze/suggest case just by making the response handling more robust so it will accept Maui Server responses as well.

All the configuration for the Maui Server tagger should go in the Annif `projects.cfg` configuration. There could be a new Annif command `init` (#251) that creates the tagger on the Maui server end, based on the configuration. Possibly it would upload the vocabulary to Maui as well, though then `loadvoc` needs to be executed first. The other (probably better) option would be to add a hook for notifying backends about loading the vocabulary, which would trigger the upload.

osma:
A new version of Maui Server that includes the NatLibFi enhancements to Maui (e.g. support for Finnish and Swedish) is being developed at https://github.com/NatLibFi/MauiServer/

We need to start working on this backend soon.

osma:
One challenge is that Maui requires a vocabulary in SKOS format. Annif does support loading a SKOS vocabulary, but currently it doesn't store the SKOS file at all, it just extracts URIs and prefLabels and saves these into a TSV file.

PR #306, opened 2019-08-07
Issue251 cli init command

### LINKED ISSUE #251: CLI init command for clearing project data
We could have a command like `annif clear tfidf-en` that removes all the data files for a particular project.

juhoinkinen:
The `init` command simply deletes all files under `data/projects/project-to-remove/` (that directory will remain):
```
$ annif init tfidf-fi-copy 
Removed data files for project tfidf-fi-copy.

$ annif init nonexistent-project
No projects found with id 'nonexistent-project'.
```

Stated in the issue comment: "It could wipe all state and reset any configuration settings." 
Is this PR enough for this? 

The issue #269 for creating backend for using Maui server plans using `init` command for "creating tagger on the Maui server end", which is left for future.

This closes #251.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -128,7 +128,7 @@ def run_show_project(project_id):
 @common_options
 def run_init_project(project_id):
     """
-    Remove the data files of a project.
+    Initialize the project to its original, untrained state.
     """
 
     proj = get_project(project_id)
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -132,7 +132,6 @@ def run_clear_project(project_id):
     """
     proj = get_project(project_id)
     proj.remove_model_data()
-    click.echo('Removed model data for project {}.'.format(project_id))
 
 
 @cli.command('loadvoc')
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -10,6 +10,7 @@ import click
 import click_log
 from flask import current_app
 from flask.cli import FlaskGroup, ScriptInfo
+from shutil import rmtree
 import annif
 import annif.corpus
 import annif.eval
@@ -130,19 +131,15 @@ def run_init_project(project_id):
     """
     Initialize the project to its original, untrained state.
     """
-
     proj = get_project(project_id)
-    datadir_path = proj.datadir
-    data_files = os.listdir(datadir_path)
-    if data_files:
-        for data_file in data_files:
-            os.remove(os.path.join(datadir_path, data_file))
-        click.echo('Removed data files for project {}.'.format(project_id))
-    else:
+    datadir_path = proj._datadir_path
+    if not os.path.isdir(datadir_path):
         click.echo(
-            'No data files to remove for project {}.'.format(project_id),
+            'No model data to remove for project {}.'.format(project_id),
             err=True)
         sys.exit(1)
+    rmtree(datadir_path)
+    click.echo('Removed model data for project {}.'.format(project_id))
 
 
 @cli.command('loadvoc')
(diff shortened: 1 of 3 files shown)


annif/cli.py
```diff
@@ -123,6 +123,28 @@ def run_show_project(project_id):
     click.echo(template.format('Access:', proj.access.name))
 
 
+@cli.command('init')
+@click.argument('project_id')
+@common_options
+def run_init_project(project_id):
+    """
+    Remove the data files of a project.
```
osma: Maybe a nitpick, but I'd change this to `Initialize the project to its original, untrained state.` or something like that - it happens to involve clearing the directory in the current implementation, but this could change later.

annif/cli.py
```diff
@@ -123,6 +123,28 @@ def run_show_project(project_id):
...
+    """
+
+    proj = get_project(project_id)
+    datadir_path = proj.datadir
+    data_files = os.listdir(datadir_path)
+    if data_files:
+        for data_file in data_files:
```
osma: Would it be easier and/or cleaner to just remove the entire directory? That would be exactly the same as the original state.

tests/test_cli.py
```diff
@@ -84,6 +84,28 @@ def test_show_project_nonexistent():
...
+    open(fpath, 'a').close()
+
+    assert runner.invoke(
+        annif.cli.cli,
+        ['init', 'dummy-fi']).exit_code == 0
+    assert not os.path.exists(fpath)
+
+
+def test_init_project_nonexistent(testdatadir):
+    runner.invoke(
```
osma: Why run this command in this test? Isn't it enough to run `init` just for `TEMP_PROJECT`?

annif/cli.py
juhoinkinen: True, naturally :) The reason for not-so-straightforward previous approach was that already accessing `proj.datadir` [created the directory](https://github.com/NatLibFi/Annif/blob/master/annif/datadir.py#L14), and raising error message for trying to init an already empty project needed the check of _file_ existence in the directory. 

But now the directory path is got using `proj._datadir_path`, simplifying the necessary logic.

osma: Hmm. I don't like accessing a "private" field like this. I'd rather move the init code to the AnnifProject class and just call it from the CLI code.

tests/test_cli.py
juhoinkinen: I think the idea was to check that `init` raises an error when a project that has entry in `projects.cfg` but no data. Removed that command now

annif/project.py
```diff
@@ -217,6 +218,14 @@ def dump(self):
                 'backend': {'backend_id': self.config['backend']}
                 }
 
+    def remove_model_data(self):
+        """remove the data of this project"""
+        datadir_path = self._datadir_path
+        if not os.path.isdir(datadir_path):
+            raise AnnifException('No model data to remove for project {}.'
```
osma: I think this should emit a warning message but it shouldn't be a fatal error.

juhoinkinen: Implemented showing warning instead of erroring. 

In the test for nonexistent data the command is run for `dummy-fi` project, because when running for a nonexistent project a (different) warning is raised already in [`get_project`](https://github.com/NatLibFi/Annif/blob/master/annif/cli.py#L31). 

The test has `logger.propagate = True` for the same reason as in https://github.com/NatLibFi/Annif/pull/299#discussion_r313433563.

osma:
Still need to document this on the wiki page [Commands](https://github.com/NatLibFi/Annif/wiki/Commands)

PR #317, opened 2019-08-15
Pin VW, Gensim, and Scikit-learn versions

### LINKED ISSUE #291: Pin down more specific versions of dependencies

juhoinkinen:
Pin versions:

- sklearn to 0.21.x
- gensim to 3.8.x
- vowpalwabbit probably to 8.7.x

At the moment fasttext(mirror) is not pinned, maybe wait for resolving #290 by merging PR #292?

Closes #291.


osma:
I think we should pin fastText too, to the current version (`fasttextmirror` version 0.8.22). This way we are at least safe against unintended upgrades if `fasttextmirror` happens to change.

Unfortunately #292 is stuck currently due to a bug in the new fastText Python bindings. When we get around to merging that we should then update the pin to the new version.

Pipfile
```diff
@@ -22,10 +22,11 @@ flask-cors = "*"
 click-log = "*"
 joblib = "*"
```
osma: I think we should pin `joblib` too, since it's used to serialize some data to disk. It would be unfortunate if the version suddenly changed when an upstream version is released, and the serialized files couldn't be loaded anymore.

juhoinkinen: Pinned `joblib==0.13.2`.

juhoinkinen:
Also pinned versions in [Wiki page for fasttext and VW installation with `pip`](https://github.com/NatLibFi/Annif/wiki/Optional-features-and-dependencies/_compare/6bafb2a12388dc8cddfa8706c51b6f93adced6ee...8a77f1209b43acaef74b1d20287d2fd7585004e8).

PR #319, opened 2019-08-22
Fix unsuccessful setting of readonly bash variable by using different name than UID

juhoinkinen:
In [the wiki section](https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker#using-annif-with-gunicorn-nginx-and-maui-backend) for `docker-compose.yml` it was instructed to run command
```
ANNIF_PROJECTS=~/annif-projects UID=${UID} GID=${GID} docker-compose up
```
to start up the services, which set `UID` and `GID` variables which in turn were intended to be used for setting user in containers. However this was actually failing with the warning
```
bash: UID: readonly variable
WARNING: The UID variable is not set. Defaulting to a blank string.
```
and the user remained root in the containers. 

This is fixed by using different name than `UID` in the `docker-compose.yml` and the instruction in Wiki.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     image: quay.io/natlibfi/annif
     volumes:
       - ${ANNIF_PROJECTS}:/annif-projects
-    user: ${UID}:${GID}
+    user: ${MY_UID}:${MY_GID}
     stdin_open: true
     tty: true
     command: bash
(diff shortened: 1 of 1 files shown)


TommiRTVA:
I have noticed that this setting is not needed or even possible to set when using Docker for Windows?

"On Docker for Windows it is not necessary to change uid and gid in your .env file.
Docker for Windows is internally using network shares (SMB) to mount Docker volumes. This does not require to syncronize file and directoriy permissions via uid and gid."

https://devilbox.readthedocs.io/en/latest/howto/uid-and-gid/find-uid-and-gid-on-win.html

PR #320, opened 2019-08-26
Turn AnnifException into an abstract base class and fix message prefix behavior

osma:
I realized that `AnnifException` isn't intended to be instantiated directly but doing so was still possible. This PR changes into an abstract base class, which cannot be instantiated.

While performing the change, I noticed that the prefix mechanism used to generate error messages was broken. This PR fixes it as well.

Example error before the changes:

    Error: Project 'noanalyzer': vocab setting is missing

Example error after the changes:

    Error: Misconfigured project 'noanalyzer': vocab setting is missing

### DIFF
--- a/annif/exception.py
+++ b/annif/exception.py
@@ -1,10 +1,11 @@
 """Custom exceptions used by Annif"""
 
 
+import abc
 from click import ClickException
 
 
-class AnnifException(ClickException):
+class AnnifException(ClickException, metaclass=abc.ABCMeta):
     """Base Annif exception. We define this as a subclass of ClickException so
     that the CLI can automatically handle exceptions."""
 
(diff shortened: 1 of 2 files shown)


osma:
The abstractness isn't actually working though. It's still possible to instantiate AnnifException when defined this way. Apparently inheriting from ClickException somehow cancels the abstractness of the class...

osma:
Something weird is happening here, I posted a [question](https://stackoverflow.com/questions/57658862/making-an-abstract-base-class-that-inherits-from-exception) on StackOverflow because I can't understand it.

juhoinkinen:
[This SO discussion](https://stackoverflow.com/questions/24792881/python-abc-module-extending-both-an-abstract-base-class-and-an-exception-derive) seems to be related.

osma:
Thanks @juhoinkinen , I wonder why I didn't find that discussion myself.

I answered my own SO question and will update this PR according to the workarounds provided in the earlier SO posts.

osma:
Almost forgot to add the tests I wrote.

PR #324, opened 2019-08-29
Issue273 default values for configuration settings

juhoinkinen:
Implemented one initial approach for discussion.

In this good is that default parameters can be included only in the relevant base classes for a backend (e.g. `chunksize` in `ChunkingBackend`). The Fasttext and VW params could be included to `DEFAULT_PARAMS` too (with a bit of work for passing only the backend specific params to the backends). 

Not-so-good is maybe the location of this, should it be in `project.py` instead? 

Another completely different approach could be to use to use [ConfigParser's defaults implementation](https://pymotw.com/3/configparser/#using-defaults).

### DIFF
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -28,7 +28,7 @@ class AnnifBackend(metaclass=abc.ABCMeta):
     def fill_params_with_defaults(self):
         """Set the parameters that are not provided in the projects config file
         with default values defined in the backend class and its bases."""
-        for source_cls in (self.__class__, *self.__class__.__bases__):
+        for source_cls in self.__class__.mro()[:-1]:  # omit the object class
             for default_param, default_value in \
                     source_cls.DEFAULT_PARAMS.items():
                 if default_param not in self.params:
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -20,22 +20,11 @@ class AnnifBackend(metaclass=abc.ABCMeta):
         backend type."""
         self.backend_id = backend_id
         self.datadir = datadir
-        self._params = config_params
+        self.params = self.default_params().copy()
+        self.params.update(config_params)
 
     def default_params(self):
         return self.DEFAULT_PARAMS
-
-    @property
-    def params(self):
-        params_to_set = {param: str(val)
-                         for param, val in self.default_params().items()
-                         if param not in self._params}
-        if params_to_set:
-            self.debug(
-                'all parameters not set, using the following defaults: {}'
-                .format(params_to_set))
-            self._params.update(params_to_set)
-        return self._params.copy()
 
     def train(self, corpus, project):
         """train the model on the given document or subject corpus"""
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -12,6 +12,10 @@ class AnnifBackend(metaclass=abc.ABCMeta):
     needs_subject_index = False
     needs_subject_vectorizer = False
 
+    DEFAULT_PARAMS = {
+        'limit': 100,
+    }
+
     def __init__(self, backend_id, params, datadir):
         """Initialize backend with specific parameters. The
         parameters are a dict. Keys and values depend on the specific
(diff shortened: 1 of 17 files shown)


osma:
I like the way backend classes (including mixins like ChunkingBackend) declare their own default parameters, but the method for looking them up by accessing the class object seems a bit overcomplicated. I'd prefer to use standard inheritance for this.

How about letting backend classes declare a property that, when accessed, returns the list of default parameter values. The implementation of the property method can then decide whether the defaults are inherited from superclasses or not. The simple case (inherited from AnnifBackend) would just return DEFAULT_PARAMS. Something like:

```python
class AnnifBackend:

    DEFAULT_PARAMS = {'limit': 100}

    @property
    def default_params(self):
        return self.DEFAULT_PARAMS


class ChunkingBackend:

    DEFAULT_PARAMS = {'chunksize': 2}


class FastTextBackend(ChunkingBackend, AnnifBackend):

    DEFAULT_PARAMS = {'loss', 'hs'}  # and others
    
    @property
    def default_params(self):
        # Start with AnnifBackend defaults
        params = AnnifBackend.default_params(self).copy()
        # override with ChunkingBackend defaults
        params.update(ChunkingBackend.default_params(self))
        # override with our own defaults
        params.update(self.DEFAULT_PARAMS)
        return params

```

Okay, maybe this isn't the simplest method either, but at least backends have full control over which default parameters they wish to inherit from upper classes.

Note that FastTextBackend already has a `FASTTEXT_PARAMS` field that specifies the types of parameters it supports. VWMultiBackend and VWEnsembleBackend go even further with `VW_PARAMS` that also specifies some default values. Their implementation of the `default_params` property could make use of this information.

osma:
On second thought, maybe `@property` is not helpful here. `default_params` could just be a plain method.

annif/backend/backend.py
```diff
@@ -12,13 +12,30 @@ class AnnifBackend(metaclass=abc.ABCMeta):
...
         backend type."""
         self.backend_id = backend_id
-        self.params = params
         self.datadir = datadir
+        self.params = self.fill_params_with_defaults(params)
+
+    def default_params(self):
+        return self.DEFAULT_PARAMS
+
+    def fill_params_with_defaults(self, params):
```
juhoinkinen: Would the params actually be better to be implemented as a property, and then this would be decorated `@params.setter`?

osma: I like the idea of `params` being a property. It could be implemented in such a way that it transparently fills in default parameters before returning a dict. And it could have a setter used to set manual parameters (which would be stored in an internal field, say `_params`). Actually, the filling in could also be done within the setter - then it would only happen once instead of on every access. Accessing the `params` property would then just return `self._params` or perhaps a `copy()` to be safe.

juhoinkinen:
> This pull request **introduces 1 alert** when merging [91136a6](https://github.com/NatLibFi/Annif/commit/91136a639b0a27fc1b6923a1596ee246fe5565b7) into [e449223](https://github.com/NatLibFi/Annif/commit/e449223a6af45590dffc43a71ff05016d6bef712) - [view on LGTM.com](https://lgtm.com/projects/g/NatLibFi/Annif/rev/pr-d14f7474ce7355ee9407ac7711d87a2c125aecec)
> 
> **new alerts:**
> 
>     * 1 for __init__ method calls overridden method

According to [this thread](https://stackoverflow.com/questions/6858842/is-it-safe-to-call-an-overridden-method-from-init) the alert could be ignored.

osma:
Regarding the LGTM alert `__init__ method calls overridden method`, I think it could be solved like this:

1. __init__ method doesn't touch `self.params`, it just sets `self.config_params = config_params`
2. `params` becomes a `@property` method like this:

```python
    @property
    def params(self):
        params = {}
        params.update(self.default_params())
        params.update(self.config_params)
        return params
```

With the above implementation of `params`, there should be no need to call `.copy()` anywhere.

The benefit of starting from an empty `dict` and using `update`, instead of taking a `.copy()`, is that the `default_params()` methods in backend subclasses may return anything dict-like, it doesn't have to be an actual dict.

annif/backend/dummy.py
```diff
@@ -11,6 +11,10 @@ class DummyBackend(backend.AnnifLearningBackend):
     uri = 'http://example.org/dummy'
     label = 'dummy'
 
+    def default_params(self):
+        params = backend.AnnifBackend.DEFAULT_PARAMS.copy()
```
osma: `.copy()` can be dropped - this could just be `return backend.AnnifBackend.DEFAULT_PARAMS`

PR #331, opened 2019-09-23
Add nn_ensemble backend

### LINKED ISSUE #330: Neural network ensemble backend
I've done some succesful experiments with using a Keras/TensorFlow based neural network to implement an intelligent example in this [Jupyter notebook](https://github.com/osma/Annif-fusion/blob/master/Annif-Fusion-LTR.ipynb). The results have been encouraging, but of course need to be validated in more settings.

I'd like to add a `nn_ensemble` backend to Annif based on this work.

osma:
Neural network ensemble backend based on Keras and TensorFlow. WIP.

Creating the (draft) PR early on so that I get feedback from the QA tools etc.

Fixes #330


osma:
TODO items:

- [ ] make (hyper)parameters configurable (should be coordinated with #324 / #273, probably that work should be merged first)
- [ ] decide if this should remain an optional extra, or a core dependency (keras + tensorflow)
- [ ] pin versions of keras and tensorflow; see if we can get rid of deprecation warnings by choosing compatible versions
- [ ] document this backend in the wiki

annif/backend/nn_ensemble.py
```diff
@@ -0,0 +1,126 @@
...
+        self._model = load_model(model_filename)
+
+    def _merge_hits_from_sources(self, hits_from_sources, project, params):
+        score_vector = np.array([hits.vector * weight
+                                 for hits, weight in hits_from_sources])
+        results = self._model.predict(
+            np.expand_dims(score_vector.transpose(), 0))
+        return VectorSuggestionResult(results[0], project.subjects)
+
+    def _create_model(self, sources, project):
```
osma: Any comments @juhoinkinen or @mvsjober to the Keras model defined here? Would you do something differently?

juhoinkinen: Looks good

Issue #339, opened 2019-10-24
Reduce memory consumption of PAV backend


osma:
Currently the PAV backend collects the subject vectors for each document into memory. This takes up a lot of RAM and limits the number of training documents it can handle. Several improvements could be made in this area:

* limit the precision of subject vectors (float32 instead of float64)
* use sparse arrays for the subject vectors
* when collecting the subject vectors training the PAV backend, store them on disk (e.g. in a temporary file or LMDB database) instead of in RAM

osma:
The `float32` part was implemented in #340; other ideas have been split up into individual issues. Closing this one.

PR #340, opened 2019-10-25
Reduce vector memory usage

osma:
This PR changes the data type of many NumPy arrays used to encode subject and suggestion vectors. Suggestions are now encoded using float32 (instead of float64) and known subjects are encoded using boolean arrays (instead of int8).

The end result is reduced memory usage and in some cases also faster execution. For example the `eval` command on a `tfidf` backend is now noticeably faster (around a third in one test).

Among other improvements this should reduce the memory consumption of the PAV backend during training (#339)


osma:
Drone builds are failing and BCH is complaining about an extra line in a method, but I'm ignoring those as the important tests are all clear.

juhoinkinen:
> Drone builds are failing and BCH is complaining about an extra line in a method, but I'm ignoring those as the important tests are all clear.

Drone builds fixed by #341.

PR #344, opened 2019-10-29
Add Maui Server backend

### LINKED ISSUE #269: Maui Server backend, including training support

osma:
Fixes #269 

This is currently WIP but already working. Unit tests need to be written and I expect that the code has to be refactored, but let's see what the QA tools say.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,6 +13,7 @@ swagger-tester = "*"
 bumpversion = "*"
 sphinx = "*"
 sphinx-rtd-theme = "*"
+responses = "*"
 
 [packages]
 "e1839a8" = {path = ".", editable = true}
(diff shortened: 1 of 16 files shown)


osma:
TODO items:

- [ ] ensure full test coverage
- [ ] more robustness checks, e.g. what happens if suggest is run before train
- [ ] documentation in the wiki

PR #354, opened 2019-11-05
Add F1@5 evaluation metric

osma:
I find that I often want to see the F1 score at K=5, i.e. when the suggestions are limited to the top 5 subjects. Currently this requires using `annif eval --limit 5 ...` but this is inconvenient and I often forget the `--limit` parameter. 

This PR adds F1@5 to the list of evaluation results.

The `filter_pred_top_k` function is a bit awkward - my NumPy array mangling skills weren't enough to implement the logic without resorting to a `for` loop. It can be changed later if necessary, and the unit test verifies that it at least gives the correct result.

### DIFF
--- a/annif/eval.py
+++ b/annif/eval.py
@@ -9,6 +9,19 @@ from sklearn.metrics import label_ranking_average_precision_score
 from annif.exception import NotSupportedException
 
 
+def filter_pred_top_k(preds, limit):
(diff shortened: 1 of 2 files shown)


osma:
I was expecting the QA checks to run, but according to https://www.githubstatus.com/ there are some issues with PR handling at the moment.

PR #366, opened 2019-12-03
Issue343 omikuji backend

### LINKED ISSUE #343: Omikuji backend
[Omikuji](https://github.com/tomtung/omikuji) looks like a very promising algorithm for subject classification. It has performed extremely well in tests performed by CSC (by @jmakoske and @mvsjober). We could add a new backend that integrates Omikuji into Annif.

osma:
This PR adds an Omikuji backend.
Fixes #343


osma:
Still need to refactor the use of TfidfVectorizer. I think this functionality could be extracted into a mixin class that is also used by TfidfBackend - though it's a bit more complex than that since TfidfBackend also converts the train corpus from document-oriented to a subject-oriented view before creating the vectorizer.

osma:
I can't figure out why Codecov doesn't show the `continue` line as covered, even though it clearly is exercised by two different tests and now even has the `# noqa` marker. But I won't spend any more time on that.

I'd like to make it possible to use the AttentionXML like parameters (mentioned in the omikuji [README](https://github.com/tomtung/omikuji/blob/master/README.md#layer-collapsing-for-balanced-shallow-trees)) as well, but it seems impossible because the `collapse_every_n_layers` hyperparameter is not exposed via the Python bindings. Reported it [here](https://github.com/tomtung/omikuji/issues/12).

Documentation still TODO...

osma:
Documentation: https://github.com/NatLibFi/Annif/wiki/Backend%3A-Omikuji
Also added install instructions & linked to the backend page from the appropriate places.

PR #370, opened 2019-12-13
Support Python 3.8

osma:
This PR attempts to enable support for Python 3.8 and have the Travis CI tests run also under Python 3.8


osma:
Some observations:
- upgraded scikit-learn to 0.22.* and joblib to 0.14.*
- disabled installing tensorflow (for nn_ensemble) and omikuji for Python 3.8 on Travis, as these are not currently available as pre-built wheels

The apparent drop in coverage is a bit mysterious. I'm pretty sure those code lines are being executed but Codecov doesn't agree. In any case they are not caused by the changes in this PR.

PR #375, opened 2020-01-14
Enable https in Swagger

juhoinkinen:
`annif.dev.finto.fi` was updated to use `https` instead of `http`, and this change allows Swagger use it.

### DIFF
--- a/annif/swagger/annif.yaml
+++ b/annif/swagger/annif.yaml
@@ -4,6 +4,7 @@ info:
   version: v1
 schemes:
   - http
+  - https
 basePath: /v1
 parameters:
   project_id:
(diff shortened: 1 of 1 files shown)


osma:
The problem with this change (adding `https` to `schemes`) is that it promises `https` support, even though it may not actually be available, depending on the installation. I think it would be better to omit the `schemes` setting. The [OpenAPI 2.0 spec](https://swagger.io/docs/specification/2-0/api-host-and-base-path/) says:

> If schemes are not specified, the scheme used to serve the API specification will be used for API calls.

I think that's more in line with what we want, as it supports both `http` and `https` but doesn't promise either one, only that the API is available on the same scheme as the OpenAPI profile, which is already the case when using Connexion.

PR #381, opened 2020-01-28
Use LMDB and sparse vectors in nn_ensemble backend

### LINKED ISSUE #363: Use LMDB to store vectors in nn_ensemble
When the `nn_ensemble` backend is trained, it sends all documents through the source projects and aggregates their suggestion vectors in memory. This can take up a significant amount of RAM. For example, with three YSO based source projects, I could only train a NN ensemble with 8k documents on a machine with 16GB RAM - any larger training set leads to an out of memory situation.

If we instead streamed the vectors to a LMDB database, and then read them back from the LMDB in batches, the backend could scale to much larger training data sets. An additional benefit would be that the LMDB could be retained on disk, so that another training run could be made using the same documents but different hyperparameters (this would require implementing the `--cached` option - see #342), without having to process the documents again, so it would be much faster.

LMDB seems to be ideal for this as it is very fast and supports streaming style operations both for reading and writing. It will introduce an additional dependency, though.

osma:
Before this PR, the nn_ensemble backend collected score and label (target) vectors in memory, which took a lot of RAM when there are thousands of documents, multiple source projects and a large vocabulary such as YSO. This PR makes the nn_ensemble backend instead spool those vectors into an on-disk, memory-mapped LMDB database instead (implemented as a keras.utils.Sequential subclass). The vectors are also compressed by representing them as SciPy sparse vectors.

The LMDB training data can also be reused in later training rounds by using the recently introduced `--cached` option, although this functionality is not yet implemented in the initial commit.

Fixes #363

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
 - pip install --upgrade pytest
 - pipenv install --dev --skip-lock
 - travis_wait 30 python -m nltk.downloader punkt
-# Install the optional neural network dependencies (Keras and TensorFlow)
+# Install the optional neural network dependencies (TensorFlow and LMDB)
 # - except for one Python version (3.8) so that we can test also without them
 # (tensorflow 2.0.0 isn't available for python 3.8 anyway)
 - if [[ $TRAVIS_PYTHON_VERSION != '3.8' ]]; then pip install .[nn]; fi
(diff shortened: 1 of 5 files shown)


osma:
Still need to make sure the LMDB dependencies are included in Docker images.

PR #383, opened 2020-01-31
Issue274 retain subject ids when loading vocabulary over existing one

### LINKED ISSUE #274: Retain subject IDs when loading vocabulary over existing one
Currently if you load a new version of a vocabulary over an existing one, what likely happens is that the integer subject IDs used internally within Annif will change. This will cause mismatches with backends that use the subject IDs (e.g. tfidf, fasttext, vw_multi) within their models, so you have to train them again.

Instead we could try to match URIs in the new vocabulary with old subject IDs and reuse them as much as possible. Then at least loading a vocabulary with additional concepts would not break existing models. If concepts have disappeared from the new version of the vocabulary, they should be marked as nonexistent and filtered from suggestion results.

juhoinkinen:
Closes #274.

### DIFF
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -49,6 +49,13 @@ class SubjectIndex:
     def __getitem__(self, subject_id):
         return (self._uris[subject_id], self._labels[subject_id])
 
+    def append(self, uri, label):
+        subject_id = len(self._uris)
+        self._uris.append(uri)
+        self._labels.append(label)
+        self._uri_idx[uri] = subject_id
+        self._label_idx[label] = subject_id
+
     def by_uri(self, uri):
         """return the subject index of a subject by its URI"""
         try:
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -17,7 +17,7 @@ class SubjectFileTSV:
     def subjects(self):
         with open(self.path, encoding='utf-8-sig') as subjfile:
             for line in subjfile:
-                uri, label = line.strip().split(None, 1)
+                uri, _, label = line.strip().partition('\t')
                 clean_uri = annif.util.cleanup_uri(uri)
                 yield Subject(uri=clean_uri, label=label, text=None)
 
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -88,6 +88,12 @@ class SubjectIndex:
                 for subject_id in (self.by_label(label) for label in labels)
                 if subject_id is not None]
 
+    def empty_labels(self):
+        """return subject indices of subjects with empty labels"""
+
+        return [subject_id for subject_id, label in enumerate(self._labels)
+                if label == '']
+
     def save(self, path):
         """Save this subject index into a file."""
 
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -56,6 +56,9 @@ class SubjectIndex:
         self._uri_idx[uri] = subject_id
         self._label_idx[label] = subject_id
 
+    def contains_uri(self, uri):
+        return uri in self._uris
+
     def by_uri(self, uri):
         """return the subject index of a subject by its URI"""
         try:
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -91,8 +91,8 @@ class SubjectIndex:
                 for subject_id in (self.by_label(label) for label in labels)
                 if subject_id is not None]
 
-    def empty_labels(self):
-        """return subject indices of subjects with empty labels"""
+    def deprecated_ids(self):
+        """return indices of deprecated subjects"""
 
         return [subject_id for subject_id, label in enumerate(self._labels)
                 if label == '']
--- a/annif/corpus/subject.py
+++ b/annif/corpus/subject.py
@@ -31,17 +31,18 @@ class SubjectIndex:
     """An index that remembers the associations between integers subject IDs
     and their URIs and labels."""
 
-    def __init__(self, corpus):
+    def __init__(self, corpus=None):
         """Initialize the subject index from a subject corpus."""
         self._uris = []
         self._labels = []
         self._uri_idx = {}
         self._label_idx = {}
-        for subject_id, subject in enumerate(corpus.subjects):
-            self._uris.append(subject.uri)
-            self._labels.append(subject.label)
-            self._uri_idx[subject.uri] = subject_id
-            self._label_idx[subject.label] = subject_id
+        if corpus is not None:
+            for subject_id, subject in enumerate(corpus.subjects):
+                self._uris.append(subject.uri)
+                self._labels.append(subject.label)
+                self._uri_idx[subject.uri] = subject_id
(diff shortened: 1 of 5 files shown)


juhoinkinen:
Demo run: 
```
rm data/vocabs/yso-fi/subjects
annif loadvoc tfidf-fi tests/corpora/archaeology/subjects.tsv
annif train tfidf-fi tests/corpora/archaeology/documents.tsv 
cat removed_subjects.txt | annif suggest tfidf-fi
annif loadvoc tfidf-fi tests/corpora/archaeology/subjects_updated.tsv 
cat removed_subjects.txt | annif suggest tfidf-fi
```
In `subjects_updated.tsv` the following subject labels have been removed and are in `removed_subjects.txt`:
```
mykeneläinen kulttuuri
fosfaattianalyysi
kykladinen kulttuuri
rahalöydöt
aarrelöydöt
neoliittinen kausi
hylyt
foinikialainen kulttuuri
laivalöydöt
vedenalainen arkeologia
```
Also, the label `meriarkeologia` is renamed to `suurten vesien arkeologia` in the updated vocabulary.

The suggestion results with original vocabulary:
```
<http://www.yso.fi/onto/yso/p10295>	kykladinen kulttuuri	0.33801543712615967
<http://www.yso.fi/onto/yso/p8867>	meriarkeologia	0.2033890336751938
<http://www.yso.fi/onto/yso/p8993>	hylyt	0.19928325712680817
<http://www.yso.fi/onto/yso/p8888>	foinikialainen kulttuuri	0.18877731263637543
<http://www.yso.fi/onto/yso/p8869>	laivalöydöt	0.17182369530200958
<http://www.yso.fi/onto/yso/p8868>	vedenalainen arkeologia	0.16133852303028107
<http://www.yso.fi/onto/yso/p9285>	neoliittinen kausi	0.15398944914340973
<http://www.yso.fi/onto/yso/p10073>	mykeneläinen kulttuuri	0.14665688574314117
<http://www.yso.fi/onto/yso/p10415>	rahalöydöt	0.1455041617155075
<http://www.yso.fi/onto/yso/p7429>	aigeialainen kulttuuri	0.10021767020225525
```
After loading updated vocabulary, the suggestion resuls are:
```
<http://www.yso.fi/onto/yso/p8867>	suurten vesien arkeologia	0.2033890336751938
<http://www.yso.fi/onto/yso/p7429>	aigeialainen kulttuuri	0.10021767020225525
<http://www.yso.fi/onto/yso/p12484>	paleoliittinen kausi	0.08510526269674301
<http://www.yso.fi/onto/yso/p4739>	Komsan kulttuuri	0.0794997438788414
<http://www.yso.fi/onto/yso/p12651>	Kundan kulttuuri	0.07667402923107147
<http://www.yso.fi/onto/yso/p19258>	Suomusjärven kulttuuri	0.07460267096757889
<http://www.yso.fi/onto/yso/p20471>	mesoliittinen kausi	0.06100035458803177
<http://www.yso.fi/onto/yso/p10826>	Askolan kulttuuri	0.051676034927368164
<http://www.yso.fi/onto/yso/p6780>	Indus-kulttuuri	0.05137034133076668
<http://www.yso.fi/onto/yso/p7428>	minolainen kulttuuri	0.05129198357462883
```

annif/vocab.py
```diff
@@ -25,6 +25,24 @@ def _create_subject_index(self, subject_corpus):
         self._subjects = annif.corpus.SubjectIndex(subject_corpus)
         annif.util.atomic_save(self._subjects, self.datadir, 'subjects')
 
+    def _update_subject_index(self, subject_corpus):
+        old_subjects = self.subjects
+        new_subjects = annif.corpus.SubjectIndex(subject_corpus)
+        updated_subjects = annif.corpus.SubjectIndex.load(os.path.devnull)
```
juhoinkinen: It's maybe a bit stupid to create a new variable for updated subjects (at least this way). Another way could be to add `__setitem__` in `SubjectIndex` class, which could be used to update the labels in `old_subjects`. Somehow this now implemented way still seemed simpler to me. What would be better?

annif/corpus/subject.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, path):
     def subjects(self):
         with open(self.path, encoding='utf-8-sig') as subjfile:
             for line in subjfile:
-                uri, label = line.strip().split(None, 1)
+                uri, _, label = line.strip().partition('\t')
                 clean_uri = annif.util.cleanup_uri(uri)
                 yield Subject(uri=clean_uri, label=label, text=None)
```
juhoinkinen: This part is different in #385.

annif/corpus/subject.py
```diff
@@ -49,6 +49,16 @@ def __len__(self):
     def __getitem__(self, subject_id):
         return (self._uris[subject_id], self._labels[subject_id])
 
+    def append(self, uri, label):
+        subject_id = len(self._uris)
+        self._uris.append(uri)
+        self._labels.append(label)
+        self._uri_idx[uri] = subject_id
+        self._label_idx[label] = subject_id
```
juhoinkinen: This method duplicates much of the code in `__init__(...)`, so if this is kept, maybe there should be a helper method `_append(self, uri, label)` that is used in both? (Also, #385 adds `notation` along `uri` and `label` here.)

annif/vocab.py
osma: I think this method of generating the new vocabulary is fine.

I think it's a bit awkward to have to use `os.path.devnull` here. How about making the subject_corpus parameter optional in the SubjectIndex constructor? Then the above line would become just:
    updated_subjects = annif.corpus.SubjectIndex()

annif/suggestion.py
```diff
@@ -175,12 +181,11 @@ def vector(self):
 
     def filter(self, limit=None, threshold=0.0):
         hits = sorted(self.hits, key=lambda hit: hit.score, reverse=True)
+        hits = [hit for hit in hits if
```
osma: The `hits` variable gets reassigned here. I think it would be better to use a different variable name. The first one could be e.g. `sorted_hits`.

annif/corpus/subject.py
```diff
@@ -81,6 +91,12 @@ def labels_to_uris(self, labels):
                 for subject_id in (self.by_label(label) for label in labels)
                 if subject_id is not None]
 
+    def empty_labels(self):
```
osma: I'm not sure I like the method name since it's so low level and doesn't even return labels. How about `deprecated_ids`?

annif/suggestion.py
```diff
@@ -136,11 +136,17 @@ def vector(self):
 
     def filter(self, limit=None, threshold=0.0):
         mask = (self._vector > threshold)
+        empty_labels = self._subject_index.empty_labels()
```
osma: Same comment as for the method name - something like `deprecated_ids` would be a better name for the variable

annif/corpus/subject.py
osma: Sounds like a reasonable refactoring to extract the common code into a method.

Issue #391, opened 2020-03-09
Evaluation per-subject


Veldhoen:
A model can be evaluated from the perspective of documents (e.g. how many correct terms were suggested for this document?) and from the perspective of subjects (e.g. how often was this subject suggested correctly?). To compute metrics, true and false positives and negatives are counted. For determining micro average, the overall counts are the same for both perspectives, but computing a macro average for documents is different from a macro average for subjects.

I propose two enhancements:
- When reporting the evaluation metrics, specify the interpretation of each metric and add weighted subject average as a metric. 
- Provide an option to output non-aggregated results per subject, to allow for a qualitative evaluation of the model (e.g. a model may be good at suggesting a certain range of topics, whereas it fails on other topics). With non-aggregated results, one can explore the relation between the frequency of subjects in the dataset and corresponding model performance.

osma:
Thanks for the suggestion. Feel free to make PRs. It might be best to make a separate PR for each specific feature if possible.

Veldhoen:
Thanks. I've created separate PRs as you suggested. I am not very familiar with all the OS processes, so chances are I did something suboptimal. Any suggestions are welcome :)

I see that some automated tests failed, I will look into that later this week.

osma:
Thanks for the PRs! I've reviewed them both. #392 looks very good, with only minor whitespace issues. #393 is a bit more complicated, but I've suggested some ways to improve it and fix the test failures.

Veldhoen:
PR #393 just got approved and merged. Closing this issue now.

PR #392, opened 2020-03-09
evaluate samples: specify interpretation of metrics (per subject), ad…

Veldhoen:
#391 addressing the first suggestion: "When reporting the evaluation metrics, specify the interpretation of each metric and add weighted subject average as a metric."
Only changes in annif.eval.py

### DIFF
--- a/annif/eval.py
+++ b/annif/eval.py
@@ -98,9 +98,7 @@ class EvaluationBatch:
         self._samples.append((hits, gold_subjects))
 
     def _evaluate_samples(self, y_true, y_pred, metrics='all'):
-
         y_pred_binary = y_pred > 0.0
-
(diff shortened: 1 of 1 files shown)


annif/eval.py
```diff
@@ -98,7 +98,9 @@ def evaluate(self, hits, gold_subjects):
         self._samples.append((hits, gold_subjects))
 
     def _evaluate_samples(self, y_true, y_pred, metrics='all'):
+
```
osma: Very minor nitpick: adding or deleting blank lines in PRs for no obvious reason is bad practice. Please change only what's necessary.

osma:
Thanks a lot Sara! Merged.

Veldhoen:
Cool, thanks!

PR #393, opened 2020-03-09
Issue391 evaluation per subject resultsfile

Veldhoen:
#391  Addressing the second suggestion: writing results per subject to file (filename given to CLI as an argument to eval command)


annif/eval.py
```diff
@@ -145,10 +145,43 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
+
+        r = len(y_true)
+
+        with open(results_file, 'w') as f:
+            f.write('\t'.join(['URI', 'Label', 'Support', 'True_positives',
+                               'False_positives', 'False_negatives',
+                               'Precision', 'Recall', 'F1_score']) + '\n')
+            zipped = zip(self._subject_index._uris,       # URI
+                         self._subject_index._labels,     # Label
+                         [sum(tp[i]) + sum(fn[i])
```
osma: I think it should be possible to use NumPy [ndarray.sum](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.sum.html#numpy.ndarray.sum) operations here, instead of using a for loop. This would be more efficient and less verbose

annif/eval.py
```diff
@@ -145,10 +145,43 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
+        y_pred = y_pred.T > 0.0
+        y_true = y_true.T > 0.0
+
+        tp = (y_true & y_pred)
+        fp = (~y_true & y_pred)
+        fn = (y_true & ~y_pred)
+
+        r = len(y_true)
+
+        with open(results_file, 'w') as f:
```
osma: The result file should probably be opened in UTF-8 mode? Also, please avoid too short variable names. 

If you use a click.File then you don't have to open it here and can just use `results_file` as a file object.

annif/eval.py
```diff
@@ -145,10 +145,43 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
+                         [sum(tp[i]) for i in range(r)],  # True_positives
+                         [sum(fp[i]) for i in range(r)],  # False_positives
+                         [sum(fn[i]) for i in range(r)],  # False_negatives
+                         [precision_score(y_true[i], y_pred[i], zero_division=0)
+                          for i in range(r)],             # Precision
+                         [recall_score(y_true[i], y_pred[i], zero_division=0)
+                          for i in range(r)],             # Recall
+                         [f1_score(y_true[i], y_pred[i], zero_division=0)
+                          for i in range(r)])             # F1
+            f.write('\n'.join('\t'.join(str(e) for e in row)
```
osma: This is probably a matter of style, but I think it would be clearer to use a more traditional for loop and a print command, something like:

```python
for row in zipped:
    print('\t'.join((str(e) for e in row)), file=f)
```

annif/eval.py
```diff
@@ -145,10 +145,43 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
         return results
 
-    def results(self, metrics='all'):
+    def output_result_per_label(self, y_true, y_pred, results_file):
+        """Write results per label (non-aggregated) to outputfile"""
+
+        y_pred = y_pred.T > 0.0
+        y_true = y_true.T > 0.0
+
+        tp = (y_true & y_pred)
```
osma: I'd prefer more verbose variable names like `true_positives` or `true_pos`

annif/eval.py
Veldhoen: Good point, I did this for the simple summations. For the precision and recall computations, I kept the list comprehensions in order to use sklearn metrics per row. Do you agree?

osma: Yes, fine to keep the for loops for those. Much better now!

annif/eval.py
```diff
@@ -145,10 +145,48 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
 
         return results
 
-    def results(self, metrics='all'):
+    def output_result_per_label(self, y_true, y_pred, results_file):
```
osma: I'd like to avoid talking of labels (in this sense) in the codebase since it already has a different meaning (like in SKOS) - could you rename this to `output_result_per_subject`? And the docstring as well.

annif/eval.py
```diff
@@ -145,10 +145,48 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
+                      for i in range(r)])                     # F1
+        for row in zipped:
+            print('\t'.join((str(e) for e in row)), file=results_file)
+
+    def results(self, metrics='all', results_file=None):
         """evaluate a set of selected subjects against a gold standard using
-        different metrics. The set of metrics can be either 'all' or
-        'simple'."""
+        different metrics. The set of metrics can be either 'all' or 'simple'.
+        If results_file (file object) is given, write results per label to it"""
```
osma: label -> subject here as well

annif/eval.py
Veldhoen: You're right, I fixed this

Veldhoen:
Sorry about the pep8 , I still need to get into the routine of running autopep before committing (or at least pushing) changes. 

Last thing on my todo list was to write a unit test, which is new territory for me. 
However, I can't get the pytest module (as is) to run to start with. Apparently, there are lots of dependencies. I am running pytest-3 (regular pytest seemed outdated). 
I am using pipenv (I ran pipenv shell), I thought that should solve dependencie issues, right? Am I missing something? I hope you can help me. 

This is the traceback:

  File "/usr/lib/python3/dist-packages/_pytest/config.py", line 328, in _getconftestmodules
    return self._path2confmods[path]
KeyError: local('<Annif-repo>/tests')

During handling of the above exception, another exception occurred:
  File "/usr/lib/python3/dist-packages/_pytest/config.py", line 359, in _importconftest
    return self._conftestpath2mod[conftestpath]
KeyError: local('<Annif-repo>/tests/conftest.py')

During handling of the above exception, another exception occurred:
  File "/usr/lib/python3/dist-packages/_pytest/config.py", line 365, in _importconftest
    mod = conftestpath.pyimport()
  File "/usr/lib/python3/dist-packages/py/_path/local.py", line 668, in pyimport
    __import__(modname)
  File "/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py", line 213, in load_module
    py.builtin.exec_(co, mod.__dict__)
  File "<Annif-repo>/tests/conftest.py", line 8, in <module>
    import annif
  File "<Annif-repo>/annif/__init__.py", line 6, in <module>
    import connexion
ModuleNotFoundError: No module named 'connexion'
ERROR: could not load <Annif-repo>/tests/conftest.py

juhoinkinen:
> Sorry about the pep8 , I still need to get into the routine of running autopep before committing (or at least pushing) changes.
> 
> Last thing on my todo list was to write a unit test, which is new territory for me.
> However, I can't get the pytest module (as is) to run to start with. Apparently, there are lots of dependencies. I am running pytest-3 (regular pytest seemed outdated).
> I am using pipenv (I ran pipenv shell), I thought that should solve dependencie issues, right? Am I missing something? I hope you can help me.
> 
> This is the traceback:
> 
> Traceback (most recent call last):
> File "/usr/lib/python3/dist-packages/_pytest/config.py", line 328, in _getconftestmodules
> return self._path2confmods[path]
> KeyError: local('/tests')
> 
> During handling of the above exception, another exception occurred:
> Traceback (most recent call last):
> File "/usr/lib/python3/dist-packages/_pytest/config.py", line 359, in _importconftest
> return self._conftestpath2mod[conftestpath]
> KeyError: local('/tests/conftest.py')
> 
> During handling of the above exception, another exception occurred:
> Traceback (most recent call last):
> File "/usr/lib/python3/dist-packages/_pytest/config.py", line 365, in _importconftest
> mod = conftestpath.pyimport()
> File "/usr/lib/python3/dist-packages/py/_path/local.py", line 668, in pyimport
> **import**(modname)
> File "/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py", line 213, in load_module py.builtin.exec_(co, mod.**dict**)
> File "/tests/conftest.py", line 8, in
> import annif
> File "/annif/**init**.py", line 6, in
> import connexion
> ModuleNotFoundError: No module named 'connexion'
> ERROR: could not load /tests/conftest.py


Hi Sara, I think to make pytest work also the [dev-packages](https://github.com/NatLibFi/Annif/blob/master/Pipfile#L6) should be [installed with pipenv](https://pipenv-fork.readthedocs.io/en/latest/basics.html#pipenv-install) (by running `pipenv install --dev`), have you checked that?

Veldhoen:
I ran pipenv install -dev now, but get the same results.. 
I'll give it another try on Monday, as I am quite tired by now :')

Veldhoen:
1. If the results-file is in a nonexistent directory, it raises an exception. 
2. I'm not too happy about the warning that is raised in normal execution: "Reading empty file", but I couldn't figure out where it was raised and how I could filter it locally. Passing "errors='ignore'" to click.File didn't make a change.. 
3. I created unit tests for the results-file, also for the case where it is in a nonexistent directory and execution fails. Please review them: as I mentioned earlier writing tests is new for me. 
4. There were some unrelated pytest failures that I ignored.

osma:
Thanks Sara!

Travis tests are failing:
```__________________________________ PEP8-check __________________________________

/home/travis/build/NatLibFi/Annif/tests/test_cli.py:533:80: E501 line too long (80 > 79 characters)

            'eval', '--results-file', 'newdir/test_output_file.txt', 'dummy-en',
```

Simply breaking/shortening the line should fix this. `autopep8` can't always fix this kind of errors.
See [build log](https://travis-ci.org/github/NatLibFi/Annif/jobs/667716360)

I'll review the other changes but wanted to report this obvious issue. This was the only test failure I saw so if you fix it then Travis tests should pass.

annif/eval.py
```diff
@@ -145,10 +145,49 @@ def _evaluate_samples(self, y_true, y_pred, metrics='all'):
...
+        y_pred = y_pred.T > 0.0
+        y_true = y_true.T > 0.0
+
+        true_pos = (y_true & y_pred)
+        false_pos = (~y_true & y_pred)
+        false_neg = (y_true & ~y_pred)
+
+        r = len(y_true)
+
+        print('\t'.join(['URI',
```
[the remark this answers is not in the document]
Veldhoen: I moved the header and body writing to helper functions. 
I considered using the header creation to check for writability, but I wanted to keep the output_results_per_subject together and at the same time I found that the check for the file being write-able should be done before heavy stuff like loading the model and evaluation. Therefore, I resolved that in a different way.

annif/cli.py
```diff
@@ -269,22 +269,39 @@ def run_index(project_id, directory, suffix, force,
...
+
     project = get_project(project_id)
     backend_params = parse_backend_params(backend_param, project)
 
     hit_filter = SuggestionFilter(limit=limit, threshold=threshold)
     eval_batch = annif.eval.EvaluationBatch(project.subjects)
 
+    if results_file:
+        try:
+            print('', file=results_file)
```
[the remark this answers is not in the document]
Veldhoen: You're right, I changed it to really write nothing to the results file (while still checking it's writable)

tests/test_cli.py
```diff
@@ -477,6 +477,35 @@ def test_eval_param(tmpdir):
...
+    tmpdir.join('doc1.key').write('dummy')
+    tmpdir.join('doc2.txt').write('doc2')
+    tmpdir.join('doc2.key').write('none')
+    tmpdir.join('doc3.txt').write('doc3')
+    result = runner.invoke(
+        annif.cli.cli, [
+            'eval', '--results-file', 'test_file.txt', 'dummy-en',
+            str(tmpdir)])
+    assert not result.exception
+    assert result.exit_code == 0
```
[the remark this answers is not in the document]
Veldhoen: I extended the test to cover these items. It may have grown a bit too large, but since it's all testing the contents of the same output file I kept it like this. 
I hope it's ok that the test is in test_cli rather than test_eval. I realised this might not be ideal only after I'd finished writing it, and moving it around wasn't trivial.

tests/test_cli.py
```diff
@@ -508,6 +508,78 @@ def test_eval_param(tmpdir):
...
 
+def test_eval_resultsfile(tmpdir):
+    tmpdir.join('doc1.txt').write('doc1')
+    tmpdir.join('doc1.key').write('dummy')
+    tmpdir.join('doc2.txt').write('doc2')
+    tmpdir.join('doc2.key').write('none')
+    tmpdir.join('doc3.txt').write('doc3')
+    result = runner.invoke(
+        annif.cli.cli, [
+            'eval', '--results-file', 'test_file.txt', 'dummy-en',
```
osma, line 519: This test leaves behind (and clobbers if it exists) a file called `test_file.txt`. It would be better to use the `tmpdir` fixture also for this purpose, then the file would be automatically deleted afterwards:

* Create a [py.path.local](https://py.readthedocs.io/en/latest/path.html#py-path-local-local-file-system-path) object like this: `resultfile = tmpdir.join('results.tsv')`
* You can get the actual file path using `str(resultfile)` (for the `--results-file` parameter value)
* You can open it using `resultfile.open()`, or just call `.readlines()` on it

osma:
Oh, one more thing - I think it would be good from a UX perspective if the CLI command printed a message like "Writing per subject evaluation results to <filename>". This would give the user some feedback that the option actually did something.

osma:
Excellent, thanks for the final touches Sara! I think this can be merged now - unless you have anything on your mind that you still want to add or change?

Veldhoen:
Yay, glad to hear! Thanks for the help and patience :)

osma:
So - good to merge from your POV?

Veldhoen:
Yes, I would say so!

osma:
So be it :) Thanks again!

PR #396, opened 2020-03-23
Issue395 Maui and Omikuji suggest with empty input 

### LINKED ISSUE #395: Suggest with empty input
A minor oddity with Omkuji backends and ensemble backends including them is that with empty input document they produce some suggestions, which is confusing. 

Another inconvenience in case of empty input to Maui backend (Mauiserver) is that it makes Annif log the following: 
```
Backend maui: HTTP request failed: 400 Client Error:  for url: http://mauiserver:8080/mauiserver/maui-fi/suggest
```

Mauiserver itself logs:
```
13 Mar 2020 14:00:55  INFO RootServlet - POST /maui-en/suggest 400, json: "Empty text"
```

juhoinkinen:
Fixes #395.

### DIFF
--- a/annif/backend/maui.py
+++ b/annif/backend/maui.py
@@ -150,6 +150,8 @@ class MauiBackend(backend.AnnifBackend):
                                                       self.project.subjects)
 
     def _suggest(self, text, params):
+        if len(text.strip()) == 0:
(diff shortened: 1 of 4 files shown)


osma:
I think this could be merged now, @juhoinkinen ?

PR #397, opened 2020-03-24
Show notations in web UI results list

### LINKED ISSUE #390: Show notations in web UI results list
Annif now stores concept notations and returns them in the CLI and REST API, but they are not shown in the Web UI. The [Vue component](https://github.com/NatLibFi/Annif/blob/master/annif/templates/home.html#L125) for results should be changed so that notations are shown as well.

juhoinkinen:
Fixes #390. 

Now notations are just added after label, for example: 
image

Should notations be somehow more distinctly separated from the labels, e.g. `[42.42]`? 

Or maybe it would be better to show them before the label? Like this:
image


osma:
Usually the notation comes before the label. See e.g. http://finto.fi/ykl/fi/page/42.42
So I suggest switching around.

Issue #403, opened 2020-04-02
Upgrade to TensorFlow 2.2 and require Python 3.6+


osma:
TensorFlow 2.2 is almost out, 2.2.0-rc2 has been released:
https://github.com/tensorflow/tensorflow/releases/tag/v2.2.0-rc2

This is the first TF version that supports Python 3.8. Once we upgrade to it, all Annif functionality will be available under Python 3.8. We could then drop support for Python 3.5 and still have full support for three Python releases (3.6, 3.7, 3.8).

PR #405, opened 2020-04-02
Upgrade to TensorFlow 2.2, Python 3.6+, drop Pipenv

osma:
This PR does several things:
* Drop the use of Pipenv which seems to cause all kinds of dependency management issues
* Upgrade to TensorFlow 2.2 (which supports Python 3.8)
* Bump up required Python version to 3.6+

Making this as a draft PR for now since TensorFlow 2.2.0 final is not yet out.


osma:
Thanks @juhoinkinen ! Will fix the comment, then merge

juhoinkinen:
This closed #403.

PR #414, opened 2020-05-15
Implement hyperparameter optimization of ensemble weights

### LINKED ISSUE #321: Automate results in optional weights in ensemble/PAV/VW
There is a lot of options when using optional weights in ensemble/PAV/VW. it would be good if there could be some way to automate testing of eval in these different weights, for example in optimize-command Annif could go through different weigh and suggest the best combination to user.

 backend1:1,backend2:1,backend2:1 -> backend1:1,backend2:1,backend2:2 -> backend1:1,backend2:2,backend2:2 etc.

osma:
Fixes #321

This is very rough and there are no unit tests yet, but creating a draft PR to get some feedback from QA tools.

### DIFF
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -69,7 +69,7 @@ class EnsembleOptimizer(hyperopt.HyperparameterOptimizer):
                          in annif.util.parse_sources(
                              backend.config_params['sources'])]
 
-    def _prepare(self, n_jobs):
+    def _prepare(self, n_jobs=1):
         self._gold_subjects = []
         self._source_hits = []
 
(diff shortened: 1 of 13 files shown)


osma:
Still TODO before merging:

- [ ] Test with larger models and validation data
- [ ] Measure memory consumption to see if there are problems
- [ ] Check if it's possible to optimize parallel runs - at the moment the speedup seems minimal
- [ ] Final check of QA tool complaints

osma:
I merged all the recent work from `master` into the PR branch, sorting out the conflicts.

Still some more work to do before this is merge-ready.

Issue #329, opened 2019-09-18
Store data about training a project to show by list-projects/show-project command


juhoinkinen:
It could be valuable to store some info about training of a project, which could also be shown by the CLI commands.

Currently `show-project` outputs the following:
```
Project ID:         tfidf-fi
Project Name:       TF-IDF Finnish
Language:           fi
Access:             public
```

Maybe some of the following data could be added:
- training state: trained or not-trained
- vocabulary state: loaded or not-loaded
- timestamp of the training run
- duration of the training run
- timestamp(s) of learning run(s)
- version of Annif that has been used in training/learning
- used backend parameters 
- number of examples in the training data
- some more details on the training data (what?)
- something that now goes to the debug log?

These data for could be stored e.g. in a metadata file `data/projects/the_project/annif_metadata` along with the model file(s).

osma:
These are all good ideas. Also it could be useful to have a "failed" status for projects/backends where the initialization fails for some reason - but that means making the initialization more careful.

I think it makes sense to start small and implement status features one (or two) at a time in separate PRs. For example trained vs. not-trained would be a good start.

Marking the issue as Long term because there are so many ideas here, but that doesn't mean we could implement some of them very soon.

osma:
Tagging with 0.48 to indicate that some work in this direction (but not everything) should be done in that release

psmukhopadhyay:
Does this element "number of examples in the training data" mean the number of records with which the model is trained? It will be very helpful if available in the show-project output. Another thought - the 'eval' command may store major metrics like F1@5, NDCG etc in the annif_metadata file every time it runs, otherwise (not run once,) F1@5 or NDCG may display 'NA'.

juhoinkinen:
> Does this element "number of examples in the training data" mean the number of records with which the model is trained? It will be very helpful if available in the show-project output. 

Yes, I actually just modified that item in the list to "number of documents used to train the project" (for clarity) and added "name(s) of the file(s) via which the documents was/were given when training the project" (I have needed that information often; however, the full path to the file(s) should **not** be stored for security reasons), and reordered the list slightly to.

PR #415, opened 2020-05-18
Show project train state and modification time

juhoinkinen:
Implements some features of #329.

Adds 
- `Trained` column to `list-projects` CLI command
- `Trained` and `Modification time` rows to `show-project` CLI command 
- `is_trained` and `modification_time` fields to the REST API responses showing project info

### DIFF
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -12,6 +12,25 @@ class EnsembleBackend(backend.AnnifBackend):
     """Ensemble backend that combines results from multiple projects"""
     name = "ensemble"
 
+    @property
+    def is_trained(self):
+        sources_trained = self._get_sources_attribute('is_trained')
+        return all(sources_trained)
+
+    @property
+    def modification_time(self):
+        mtimes = self._get_sources_attribute('modification_time')
+        return max(filter(None, mtimes), default=None)  # TODO: Or should ensemble mtime be None if any mtime of sources is None?
+
+    def _get_sources_attribute(self, attr):
+        params = self._get_backend_params(None)
+        sources = annif.util.parse_sources(params['sources'])
+        attrs = []
+        for project_id, _ in sources:
+            source_project = annif.project.get_project(project_id)
+            attrs.append(getattr(source_project, attr))
+        return attrs
+
     def _normalize_hits(self, hits, source_project):
         """Hook for processing hits from backends. Intended to be overridden
         by subclasses."""
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -20,7 +20,6 @@ class EnsembleBackend(backend.AnnifBackend):
     @property
     def modification_time(self):
         mtimes = self._get_sources_attribute('modification_time')
-        # TODO: Should ensemble mtime be None if mtime of any source is None?
         return max(filter(None, mtimes), default=None)
 
     def _get_sources_attribute(self, attr):
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -20,7 +20,8 @@ class EnsembleBackend(backend.AnnifBackend):
     @property
     def modification_time(self):
         mtimes = self._get_sources_attribute('modification_time')
-        return max(filter(None, mtimes), default=None)  # TODO: Or should ensemble mtime be None if any mtime of sources is None?
+        # TODO: Should ensemble mtime be None if mtime of any source is None?
+        return max(filter(None, mtimes), default=None)
 
     def _get_sources_attribute(self, attr):
         params = self._get_backend_params(None)
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -26,11 +26,8 @@ class EnsembleBackend(backend.AnnifBackend):
     def _get_sources_attribute(self, attr):
         params = self._get_backend_params(None)
         sources = annif.util.parse_sources(params['sources'])
-        attrs = []
-        for project_id, _ in sources:
-            source_project = annif.project.get_project(project_id)
-            attrs.append(getattr(source_project, attr))
-        return attrs
+        return [getattr(annif.project.get_project(project_id), attr)
+                for project_id, _ in sources]
 
     def _normalize_hits(self, hits, source_project):
         """Hook for processing hits from backends. Intended to be overridden
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -24,7 +24,7 @@ class EnsembleBackend(backend.AnnifBackend):
     def _get_sources_attribute(self, attr):
         params = self._get_backend_params(None)
         sources = annif.util.parse_sources(params['sources'])
-        return [getattr(annif.project.get_project(project_id), attr)
+        return [getattr(self.project.registry.get_project(project_id), attr)
                 for project_id, _ in sources]
 
     def _normalize_hits(self, hits, source_project):
--- a/annif/backend/http.py
+++ b/annif/backend/http.py
@@ -13,33 +13,30 @@ class HTTPBackend(backend.AnnifBackend):
 
     @property
     def is_trained(self):
-        response = self._get_project_info()
-        if 'is_trained' in response:
-            return response['is_trained']
-        else:
-            return None
+        return self._get_project_info('is_trained')
 
     @property
     def modification_time(self):
-        response = self._get_project_info()
-        if 'modification_time' in response:
-            return response['modification_time']
-        else:
-            return None
+        return self._get_project_info('modification_time')
 
-    def _get_project_info(self):
+    def _get_project_info(self, key):
         params = self._get_backend_params(None)
         try:
             req = requests.get(params['endpoint'].replace('/suggest', ''))
             req.raise_for_status()
         except requests.exceptions.RequestException as err:
             self.warning("HTTP request failed: {}".format(err))
-            return {}
+            return None
         try:
-            return req.json()
+            response = req.json()
         except ValueError as err:
             self.warning("JSON decode failed: {}".format(err))
-            return {}
+            return None
+
+        if key in response:
+            return response[key]
+        else:
+            return None
 
     def _suggest(self, text, params):
         data = {'text': text}
--- a/annif/backend/http.py
+++ b/annif/backend/http.py
@@ -11,6 +11,36 @@ from . import backend
 class HTTPBackend(backend.AnnifBackend):
     name = "http"
 
+    @property
+    def is_trained(self):
+        response = self._get_project_info()
+        if 'is_trained' in response:
+            return response['is_trained']
+        else:
+            return None
+
+    @property
+    def modification_time(self):
+        response = self._get_project_info()
+        if 'modification_time' in response:
+            return response['modification_time']
+        else:
+            return None
+
+    def _get_project_info(self):
+        params = self._get_backend_params(None)
+        try:
+            req = requests.get(params['endpoint'].replace('/suggest', ''))
+            req.raise_for_status()
+        except requests.exceptions.RequestException as err:
+            self.warning("HTTP request failed: {}".format(err))
+            return {}
+        try:
+            return req.json()
+        except ValueError as err:
+            self.warning("JSON decode failed: {}".format(err))
+            return {}
+
     def _suggest(self, text, params):
         data = {'text': text}
         if 'project' in params:
--- a/annif/backend/http.py
+++ b/annif/backend/http.py
@@ -5,6 +5,7 @@ and returns the results"""
 import requests
 import requests.exceptions
 from annif.suggestion import SubjectSuggestion, ListSuggestionResult
+from annif.exception import OperationFailedException
 from . import backend
 
 
@@ -25,13 +26,13 @@ class HTTPBackend(backend.AnnifBackend):
             req = requests.get(params['endpoint'].replace('/suggest', ''))
             req.raise_for_status()
         except requests.exceptions.RequestException as err:
-            self.warning("HTTP request failed: {}".format(err))
-            return None
+            msg = f"HTTP request failed: {err}"
+            raise OperationFailedException(msg) from err
         try:
             response = req.json()
         except ValueError as err:
-            self.warning("JSON decode failed: {}".format(err))
-            return None
+            msg = f"JSON decode failed: {err}"
+            raise OperationFailedException(msg) from err
 
         if key in response:
             return response[key]
--- a/annif/backend/maui.py
+++ b/annif/backend/maui.py
@@ -18,6 +18,26 @@ class MauiBackend(backend.AnnifBackend):
 
     TRAIN_FILE = 'maui-train.jsonl'
 
+    @property
+    def is_trained(self):
+        params = self._get_backend_params(None)
+        try:
+            resp = requests.get(self.tagger_url(params))
+            resp.raise_for_status()
+        except requests.exceptions.RequestException as err:
+            self.warning("HTTP request failed: {}".format(err))
+            return None
+        try:
+            return resp.json()['is_trained']
+        except ValueError as err:
+            self.warning("JSON decode failed: {}".format(err))
+            return None
+
+    @property
+    def modification_time(self):
+        # TODO Should mtime of Maui project always be None or how to get it?
+        return None
+
     def endpoint(self, params):
         try:
             return params['endpoint']
--- a/annif/backend/maui.py
+++ b/annif/backend/maui.py
@@ -26,7 +26,8 @@ class MauiBackend(backend.AnnifBackend):
     @property
     def modification_time(self):
         mtime = self._get_project_info('end_time')
-        return datetime.strptime(mtime, '%Y-%m-%dT%H:%M:%S.%fZ')
+        if mtime is not None:
+            return datetime.strptime(mtime, '%Y-%m-%dT%H:%M:%S.%fZ')
 
     def _get_project_info(self, key):
         params = self._get_backend_params(None)
--- a/annif/backend/maui.py
+++ b/annif/backend/maui.py
@@ -6,6 +6,7 @@ import os.path
 import json
 import requests
 import requests.exceptions
+from datetime import datetime
 from annif.exception import ConfigurationException
 from annif.exception import NotSupportedException
 from annif.exception import OperationFailedException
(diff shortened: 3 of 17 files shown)


juhoinkinen:
Now any file present in a project's datadir is used as an indicator to the project to be trained, and the modification time of the newest file is taken as the project's modification time.

juhoinkinen:
Marked this ready now as I did not see a good simple way to use `is_trained` info for catching the case of suggesting with not-trained project, because ensemble BE should be usable without training. Deleting `_suggest` method from the ensemble class and using condition for both `is_trained` and whether a BE has `_suggest` defined (not inherited) seemed too complex.

annif/cli.py
```diff
@@ -122,12 +122,14 @@ def run_list_projects():
     List available projects.
     """
 
-    template = "{0: <25}{1: <45}{2: <8}"
-    header = template.format("Project ID", "Project Name", "Language")
+    template = "{0: <25}{1: <45}{2: <10}{3: <10}"
+    header = template.format(
+        "Project ID", "Project Name", "Language", "Is trained")
```
osma, line 127: Maybe just "Trained" ?

annif/cli.py
```diff
@@ -122,12 +122,14 @@ def run_list_projects():
...
     click.echo(header)
     click.echo("-" * len(header))
     for proj in annif.project.get_projects(min_access=Access.private).values():
-        click.echo(template.format(proj.project_id, proj.name, proj.language))
+        click.echo(template.format(
```
osma, line 131: f-string here too?

annif/cli.py
```diff
@@ -144,6 +146,8 @@ def run_show_project(project_id):
     click.echo(template.format('Project Name:', proj.name))
     click.echo(template.format('Language:', proj.language))
     click.echo(template.format('Access:', proj.access.name))
+    click.echo(template.format('Is trained:', proj.is_trained))
```
osma, line 149: "Trained" ?
And f-strings could be used for all these echos...

annif/project.py
```diff
@@ -153,6 +155,16 @@ def vocab(self):
     def subjects(self):
         return self.vocab.subjects
 
+    @property
+    def is_trained(self):
```
osma, line 159: I see a couple of issues in implementing is_trained on the project level:
1. How to deal with ensembles (as you mentioned) - in my view, an ensemble could be considered trained when all its sources have been trained, but that requires peeking into the internals of the ensemble
2. External services such as Maui and another instance of Annif used via the `http` backend - they (probably) don't have any local state

I'd rather implement is_trained in the individual backends. The default implementation could be like this, but ensembles and remote service backends could define their own implementation.

annif/project.py
```diff
@@ -153,6 +155,16 @@ def vocab(self):
...
+        return bool(glob(os.path.join(self._datadir_path, '*')))
+
+    @property
+    def modification_time(self):
```
osma, line 163: Same comment as for `is_trained` above - I think it makes sense to implement this in individual backends instead of here on the project level, for the same reasons.

annif/cli.py
```diff
@@ -122,12 +122,15 @@ def run_list_projects():
...
+    template = "{0: <25}{1: <45}{2: <10}{3: <7}"
+    header = template.format(
+        "Project ID", "Project Name", "Language", "Trained")
     click.echo(header)
     click.echo("-" * len(header))
     for proj in annif.project.get_projects(min_access=Access.private).values():
-        click.echo(template.format(proj.project_id, proj.name, proj.language))
+        click.echo(
+            f'{proj.project_id: <25}{proj.name: <45}{proj.language: <10}'
+            f'{str(proj.is_trained): <7}')
```
juhoinkinen, line 133: I didn't find a better way to use the same field widths as in the header but this. Formatting to the template could have been better?

juhoinkinen:
@osma could take a look on this again? I start to feel that this is not really so useful as I thought... 

One issue is that if train state of a project is not available, e.g. when connection to MauiServer fails, `list-projects` is muddled with warnings:

```
Project ID               Project Name                                 Language  Trained
---------------------------------------------------------------------------------------
fasttext                 FastText                                     fi        True   
warning: Backend http: HTTP request failed: HTTPConnectionPool(host='localhost', port=5000): Max retries exceeded with url: /v1/projects/fasttext (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b387a668>: Failed to establish a new connection: [Errno 111] Connection refused'))
http                     HTTP English                                 en        None   
yso-fi                   YSO NN ensemble Finnish                      fi        True   
warning: Backend maui: HTTP request failed: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /mauiserver/maui-fi (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b387a3c8>: Failed to establish a new connection: [Errno 111] Connection refused'))
yso-ensemble-fi          YSO ensemble Finnish                         fi        False  
yso-pav-fi               YSO PAV Finnish                              fi        False  
yso-sv                   YSO NN Ensemble Swedish                      sv        False  
yso-en                   YSO NN Ensemble English                      en        False  
warning: Backend maui: HTTP request failed: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /mauiserver/maui-fi (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b387f550>: Failed to establish a new connection: [Errno 111] Connection refused'))
yso-maui-fi              YSO Maui Finnish                             fi        None   
warning: Backend maui: HTTP request failed: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /mauiserver/maui-fi (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b387feb8>: Failed to establish a new connection: [Errno 111] Connection refused'))
yso-maui-arch            YSO Maui Finnish                             fi        None   
warning: Backend maui: HTTP request failed: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /mauiserver/maui-sv (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b403aa58>: Failed to establish a new connection: [Errno 111] Connection refused'))
yso-maui-sv              YSO Maui Swedish                             sv        None   
warning: Backend maui: HTTP request failed: HTTPConnectionPool(host='localhost', port=8080): Max retries exceeded with url: /mauiserver/maui-en (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1b3876320>: Failed to establish a new connection: [Errno 111] Connection refused'))
yso-maui-en              YSO Maui English                             en        None   
yso-parabel-fi           YSO Omikuji Parabel Finnish                  fi        True   
yso-parabel-sv           YSO Omikuji Parabel Swedish                  sv        False

File: annif/swagger/annif.yaml [hunk 7]
```diff
@@ -143,6 +143,12 @@ definitions:
         example: en
       backend:
         $ref: '#/definitions/ProjectBackend'
+      is_trained:
+        type: boolean
+        example: true
+      modification_time:
+        type: string
```
osma, line 150: you could add `format: date-time` here to indicate that the string is in datetime format

annif/cli.py
osma: You're right, using the template is nicer because then you avoid declaring the column widths twice. Maybe I got carried away with my suggestion to use f-strings...

annif/backend/ensemble.py
```diff
@@ -12,6 +12,26 @@ class EnsembleBackend(backend.AnnifBackend):
...
+    def modification_time(self):
+        mtimes = self._get_sources_attribute('modification_time')
+        # TODO: Should ensemble mtime be None if mtime of any source is None?
+        return max(filter(None, mtimes), default=None)
+
+    def _get_sources_attribute(self, attr):
+        params = self._get_backend_params(None)
+        sources = annif.util.parse_sources(params['sources'])
+        attrs = []
+        for project_id, _ in sources:
```
osma, line 30: Maybe write this as a list comprehension instead? Something like
```python
return [getattr(annif.project.get_project(project_id), attr)
        for project_id, _ in sources]
```

annif/backend/ensemble.py
```diff
@@ -12,6 +12,26 @@ class EnsembleBackend(backend.AnnifBackend):
...
 
+    @property
+    def is_trained(self):
+        sources_trained = self._get_sources_attribute('is_trained')
+        return all(sources_trained)
+
+    @property
+    def modification_time(self):
+        mtimes = self._get_sources_attribute('modification_time')
+        # TODO: Should ensemble mtime be None if mtime of any source is None?
```
osma, line 23: I think it's OK to ignore None values here, as you've done

annif/project.py
```diff
@@ -153,9 +153,32 @@ def vocab(self):
     def subjects(self):
         return self.vocab.subjects
 
+    @property
+    def is_trained(self):
+        try:
+            return self.backend.is_trained
+        except AnnifException as err:
```
osma, line 160: Why catch this exception here instead of letting it bubble up to the caller?

annif/project.py
```diff
@@ -153,9 +153,32 @@ def vocab(self):
...
+            logger.warning(err.format_message())
+        except AttributeError:
+            return None
+
+    @property
+    def modification_time(self):
+        try:
+            return self.backend.modification_time
+        except AnnifException as err:
```
osma, line 169: Same as above: why catch this here?

annif/project.py
```diff
@@ -153,9 +153,32 @@ def vocab(self):
...
+            return None
+
     def suggest(self, text, backend_params=None):
         """Suggest subjects the given text by passing it to the backend. Returns a
         list of SubjectSuggestion objects ordered by decreasing score."""
+        if not self.is_trained:
+            if self.is_trained is None:
+                logger.warn('Could not get train state information.')
+            else:
+                raise ConfigurationException('Project is not trained.')
```
osma, line 181: I think NotInitializedException would be more appropriate here

annif/backend/maui.py
```diff
@@ -18,6 +18,26 @@ class MauiBackend(backend.AnnifBackend):
...
+            return None
+        try:
+            return resp.json()['is_trained']
+        except ValueError as err:
+            self.warning("JSON decode failed: {}".format(err))
+            return None
+
+    @property
+    def modification_time(self):
+        # TODO Should mtime of Maui project always be None or how to get it?
```
osma, line 38: You can get the training time (as well as `is_trained`) from the `train` endpoint, e.g. `/mauiserver/maui-en/train` which returns JSON like this:

```json
{"service_status":"ready",
 "completed":true,
 "documents":693,
 "skipped":0,
 "start_time":"2020-05-26T15:35:03.680+03:00",
 "runtime_millis":137488,
 "end_time":"2020-05-26T15:37:21.168+03:00",
 "is_trained":true,
 "training_status":"ready"}
```

I think the `end_time` is the most relevant.

annif/project.py
juhoinkinen: The reason for catching is that if project definition is missing backend, `list-projects` or `show-project` CLI commands crash. There is such a project (`nobackend`) in `tests/projects.cfg`. Maybe it would be better to catch in CLI methods (or remove that project from test projects)... 

I'll check if there are merge conflicts from #417 and look at this then again.

juhoinkinen:
I made some progress here, can you comment @osma?


> Regarding the ugly warnings - those only happen when Maui (or the HTTP endpoint) is down, so I don't see that as a major issue. But perhaps it would make sense to swallow the warnings in some way, or use DEBUG level instead which is not normally shown.

For `list-projects` the default log-level is now ERROR, so no warnings are shown at all, but if wanted they can be seen by passing `-v WARNING`. For `show-project` warnings are still by default visible. 

Codebeat complains about code quality decline, but this seems to be inevitable:

> 1 namespace has degraded:
> • annif.project.AnnifProject B > F

And lgtm-bot complains about 
> First argument to super() is not enclosing class

from e.g. nn_ensemble's
```python
@property
def is_trained(self):
    return super(ensemble.EnsembleBackend, self).is_trained
```
but this seems good for this case, to access the method from grandparent class (AnnifBackend), which is needed there instead of Ensemble backend's `is_trained` method.

juhoinkinen:
One more thing: when an ensemble project specifies a source project that does not exist and `list-projects` is run, the ValueError thrown from [get_project()](https://github.com/NatLibFi/Annif/blob/master/annif/project.py#L299) is not caught anywhere but a crash results. 

It could be caught in [when processing sources](https://github.com/NatLibFi/Annif/blob/project-train-state-and-modification-time/annif/backend/ensemble.py#L27), however for the same reason a crash results also when `suggest` is run on such ensemble or nn_ensemble project.

annif/backend/http.py
```diff
@@ -11,6 +11,33 @@
...
+        return self._get_project_info('is_trained')
+
+    @property
+    def modification_time(self):
+        return self._get_project_info('modification_time')
+
+    def _get_project_info(self, key):
+        params = self._get_backend_params(None)
+        try:
+            req = requests.get(params['endpoint'].replace('/suggest', ''))
```
osma, line 25: Having to do this URL modification here is probably an indication that the `endpoint` parameter should be pointing at a root resource of the API (either the root of the API, or the project level), instead of directly at the `/suggest` method. That would be a backwards incompatible change though. Maybe that could be fixed in a separate PR, since this one is already quite complex.

annif/backend/http.py
```diff
@@ -11,6 +11,33 @@
...
+            req.raise_for_status()
+        except requests.exceptions.RequestException as err:
+            self.warning("HTTP request failed: {}".format(err))
```
osma, line 28: Should this be changed into raising an OperationFailedException, like the `maui backend does? I know that it's not used in the `suggest` method of the `http` backend, but perhaps it would be better to use it also there.

annif/backend/nn_ensemble.py
```diff
@@ -97,6 +97,14 @@ class NNEnsembleBackend(
     # defaults for uninitialized instances
     _model = None
 
+    @property
+    def is_trained(self):
+        return super(ensemble.EnsembleBackend, self).is_trained
```
osma, line 102: Is there any reason to redefine this property here instead of just inheriting it from EnsembleBackend? Or is the problem that it is being inherited from AnnifBackend (via AnnifLearningBackend) instead? Then maybe the order of superclass declarations could be changed?

annif/backend/nn_ensemble.py
```diff
@@ -97,6 +97,14 @@ class NNEnsembleBackend(
...
+
+    @property
+    def modification_time(self):
+        return super(ensemble.EnsembleBackend, self).modification_time
```
osma, line 106: same as above

annif/backend/pav.py
```diff
@@ -26,6 +26,14 @@ class PAVBackend(ensemble.EnsembleBackend):
 
     DEFAULT_PARAMETERS = {'min-docs': 10}
 
+    @property
+    def is_trained(self):
+        return super(ensemble.EnsembleBackend, self).is_trained
```
osma, line 31: same as above

annif/backend/pav.py
```diff
@@ -26,6 +26,14 @@ class PAVBackend(ensemble.EnsembleBackend):
...
+
+    @property
+    def modification_time(self):
+        return super(ensemble.EnsembleBackend, self).modification_time
```
osma, line 35: same as above

annif/backend/ensemble.py
```diff
@@ -11,6 +11,22 @@ class EnsembleBackend(backend.AnnifBackend):
...
+
+    @property
+    def modification_time(self):
+        mtimes = self._get_sources_attribute('modification_time')
+        return max(filter(None, mtimes), default=None)
+
+    def _get_sources_attribute(self, attr):
+        params = self._get_backend_params(None)
+        sources = annif.util.parse_sources(params['sources'])
+        return [getattr(annif.project.get_project(project_id), attr)
```
osma, line 27: Now that #417 has been merged into `master`, you should change from `annif.project.get_project` to `self.project.registry.get_project` so you don't need to go through Flask machinery to access the source projects

tests/test_backend_http.py
```diff
@@ -130,3 +130,40 @@ def test_http_suggest_unexpected_json(project):
             project=project)
         result = http.suggest('this is some text')
         assert len(result) == 0
+
+
+def test_http_is_trained(project):
```
osma, line 135: Should you test also the cases where the HTTP operation fails for some reason?

tests/test_backend_maui.py
```diff
@@ -287,3 +288,19 @@ def test_maui_suggest_unexpected_json(maui, project):
     result = maui.suggest('this is some text')
     assert len(result) == 0
     assert len(responses.calls) == 1
+
+
+@responses.activate
+def test_maui_is_trained(maui, project):
```
osma, line 294: Same here, test also for the unsuccessful cases

tests/test_backend_nn_ensemble.py
```diff
@@ -105,6 +114,24 @@ def test_nn_ensemble_train_and_learn_params(registry, tmpdir, capfd):
...
+def test_nn_ensemble_is_trained(app_project):
+    nn_ensemble_type = annif.backend.get_backend('nn_ensemble')
+    nn_ensemble = nn_ensemble_type(
+        backend_id='nn_ensemble',
+        config_params={'sources': 'dummy-en'},
+        project=app_project)
+    assert nn_ensemble.is_trained
+
+
+def test_pav_modification_time(app_project):
```
osma, line 126: should this function be named `test_nn_ensemble_modification_time` instead?

annif/backend/http.py
juhoinkinen: I changed the `_get_project_info()` of http and Maui backends to raise the `OperationFailedException` instead of only warn in case http or json failures. The raised exception is then handled in `_get_info()` of project module (to allow `list-projects` command to continue after a failing http project).

annif/backend/nn_ensemble.py
juhoinkinen: The (re)definition is needed because the property that would be inherited from EnsembleBackend is not applicable here as it does not check the local state (files present in datadir), and I think the NN Ensemble (and PAV) should be considered trained only if they have the model files in datadir. Same goes for the modification time.

The intention here is to use the (default) property from AnnifBackend. Instead of the `super()` call the default implementations could be simply duplicated in here, at least then the lgmt bot would not complain.

juhoinkinen: If the `modification_time` in AnnifBackend wasn't a property but regular method, this could be ofcourse used:
```python
def is_trained(self):
    return backend.AnnifBackend.is_trained(self)
```
As in shown [here](https://stackoverflow.com/questions/18117974/calling-a-parents-parents-method-which-has-been-overridden-by-the-parent) but with property this doesn't work. :/

juhoinkinen: Actually for accessing the property from AnnifBackend is possible with `__get__()`, like:
```python
@property
def is_trained(self):
    return backend.AnnifBackend.is_trained.__get__(self)
```

Issue #65, opened 2018-03-20
Evaluate documents in parallel


osma:
We could perhaps speed up the evaldir command by making use of multiple CPUs, using the multiprocessing module. There would be a pool of workers (as many as there are CPU cores) and documents would be handed to the workers for evaluation.

osma:
Logging is a bit of a challenge, but there is https://pypi.python.org/pypi/multiprocessing-logging/ that might help

kinow:
For me the evaluation of a directory took a while, but I suspect over 80% of the time was loading the vectorizer from disk.

Is the idea here to perhaps first pre-load the model, and then evaluate the documents in parallel?

osma:
@kinow Yes, the initialization time (loading vectorizer, models etc) tends to dominate when you evaluate with a small amount of documents. Possibly some of this initialization could be parallelized as well...

Anyways, the idea of this feature was to load the model first, then evaluate the documents in parallel. With a large enough set of documents (thousands?) there should be a significant reduction in overall time spent on evaluation.

osma:
The main thing holding this back currently is that Annif projects are too tightly tied with the Flask current_app object. This is especially problematic for ensemble backends (and vw_multi) that need to access other projects. They have to do that via current_app. This works fine in the main thread, but not in subprocesses launched by e.g. multiprocessing or joblib.Parallel.

I'm working on decoupling the Annif internals from Flask and will prepare a PR on that soon. I also have a mostly-working implementation on parallel evaluation of documents but it needs a bit more work (and perhaps first some more refactorings to avoid passing around large objects).

PR #417, opened 2020-05-28
Separate Annif projects from Flask current_app

osma:
The beginnings of implementing #65. At the moment this draft PR only contains commits that separate the Annif project handling from the Flask current_app object. Making a PR to get early feedback from QA tools.

### DIFF
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -2,7 +2,6 @@
 
 
 import annif.suggestion
-import annif.project
 import annif.util
 from . import backend
 from annif.exception import NotSupportedException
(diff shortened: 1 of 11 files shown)


osma:
Would you like to take a look @juhoinkinen before I merge this?

The main change here is to introduce the AnnifRegistry class which tracks all the available projects and stores them (somewhat unconventionally) in a static variable. Projects now keep a reference to the registry and projects/backends can access other backends via the registry, instead of using the utility function `annif.project.get_project` which relies on the Flask app context. That function still exists but it is only used to access projects when there is no direct reference to the registry available, for example in the CLI, REST API and some unit tests.

PR #418, opened 2020-06-05
Multiprocessing in eval command

### LINKED ISSUE #65: Evaluate documents in parallel

osma:
This PR makes it possible to use multiprocessing to speed up the `eval` command.

The majority of the changes are actually refactorings to decouple the subject index from the SuggestionResult classes. After the changes, SuggestionResult instances no longer keep a reference to SubjectIndex. Instead a SubjectIndex is passed as a parameter to individual SuggestionResult methods as necessary. Since properties cannot take parameters, the `hits` property has been changed to the `as_list` method and the `vector` property has been changed to the `as_vector` method.

The actual multiprocessing implementation is still very rough and further changes are needed:

- [x] make it possible to select the number of parallel jobs (e.g. using a `--jobs` CLI argument)
- [x] don't use multiprocessing module if jobs=1
- [x] (try to) optimize VectorSuggestionResult.filter() so it won't return large vectors which cause serialization/deserialization overhead
- [x] test with different backends to make sure that they perform well (measure CPU time, wall time, peak memory) and don't have race conditions or other similar issues

Fixes #65

### DIFF
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -34,7 +34,7 @@ def create_app(script_info=None, config_name=None):
     CORS(cxapp.app)
 
     if cxapp.app.config['INITIALIZE_PROJECTS']:
-        annif.project.initialize_projects(cxapp.app)
+        annif.registry.initialize_projects(cxapp.app)
 
     # register the views via blueprints
     from annif.views import bp
--- a/annif/backend/dummy.py
+++ b/annif/backend/dummy.py
@@ -25,8 +25,7 @@ class DummyBackend(backend.AnnifLearningBackend):
         return ListSuggestionResult([SubjectSuggestion(uri=self.uri,
                                                        label=self.label,
                                                        notation=notation,
-                                                       score=score)],
-                                    self.project.subjects)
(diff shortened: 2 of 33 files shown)


osma:
Rebased on current `master` and force-pushed.

osma:
After performing some testing, I'm fairly confident that this feature works with at least most types of backend (tested tfidf, omikuji, fasttext, maui and simple ensemble), but the speedup is not very big - the parallelization overhead is quite significant and the initialization of models and postprocessing of results, which cannot be parallelized, take up significant chunks of time too. In practice, with two parallel jobs, the evaluation takes around the same time as with one job, and to get any improvement in overall evaluation time, you need to use jobs=4 or more. 

I changed the default to jobs=1 so that parallel evaluation is only performed when requested by the user.

I will open another issue on implementing a parallel optimize command.

Still some more refactoring, then this can be merged I think.

PR #419, opened 2020-06-11
WEB-UI: Remove empty entry from list of projects.

mo-fu:
Remove an empty entry in the Vue.js UI. With the change, the first element of project list is selected by default. Therefore the user does not need to select a project manually.


osma:
Thanks @mo-fu!

PR #422, opened 2020-06-25
Assign first retrieved project to selected variable.

mo-fu:
My last PR selected the first project only visually. When trying to analyze text in the web UI, before selecting another project than the first, the user was asked to select a project. The only way to select the first project was to select another project and then again select the first project from the drop down menu. Sorry for the mess.


osma:
Thanks for the follow-up fix @mo-fu !

PR #426, opened 2020-07-08
Issue424 New style for web UI

### LINKED ISSUE #424: New style for Web UI
Now that the new [annif.org](https://annif.org) website has been published, it would make sense to also update the Annif Web UI to match the style. So things like

- [x] upgrade to latest Bootstrap and jQuery versions
- [x] install the Jost font
- [x] put the new Annif logo in place
- [x] adapt the form styling and functionality (clear button etc)
- [x] adapt the style of the rest of the page (e.g. font styling, backgrounds/gradients...)

The form part might be a bit challenging because the Annif Web UI uses Vue.js, while the Annif website uses just jQuery. So the code cannot just be copied, it needs to be adapted for Vue as well.

juhoinkinen:
Fixes #424. 

Adapts web UI to the new Annif style:
image

In contrast to annif.org:
- The second Annif logo next to "Get suggestions" is left out, as in Web UI the logo in heading can be seen all the time.
- The "Get suggestions" button is left enabled also when text box is empty (with empty input the following error message is shown:  `Please correct the following error(s): You need to enter the text to get suggestions for`).
- The "Get suggestions" button is disabled when query is running (this avoids getting duplicated suggestions if the button is clicked second time during query). A spinner is shown inside the button, not in the suggestions list area.


juhoinkinen:
Actually, should there be "Web UI" instead of "REST API" in the header?

osma:
Looks very good!

Regarding the REST API header, I agree it's misleading. Perhaps it would be better to have a separate section/box for REST things, similar to the [Finto AI](https://ai.finto.fi) layout, and call this the Web UI.

SonarCloud found a few HTML/CSS bugs, will need to check out those.

I will try to continue from here and get this merged in the next few days so we can release 0.49.

osma:
@kinow Want to take a look by any chance? You've done previous work on the Web UI, e.g. the Vue.js switch...

annif/templates/home.html
```diff
@@ -112,20 +133,51 @@ <h2 v-if="problem.hasTitle()"><% problem.title %></h2>
   },
   template: '<div>\
   <div class="form-group">\
-    <label for="project">Project (vocabulary and language):</label>\
+    <label for="project">Project (vocabulary and language)</label>\
+    <div class="select-wrapper">\
     <select class="form-control" id="project" @input="handleProjectSelected">\
       <option v-for="project in projects" v-bind:value="project.project_id"><% project.name %></option>\
+    </div>\
```
kinow, line 140: I think the `div` is closing the `select`. Probably need to move one line down?

annif/static/css/style.css
```diff
@@ -1,2 +1,105 @@
...
+a:visited { color: #536f9f }
+h1 { font-size: 2rem; }
+#form {
+  padding: 2rem;
+  background-color: #f3f3f6;
+  background: -moz-linear-gradient(top, #ffffff 0%, #d9dfe3 1%, #f3f3f6 1%, #f3f3f6 99%, #d9dfe3 99%, #ffffff 100%);
+  background: -webkit-linear-gradient(top, #ffffff 0%, #d9dfe3 1%, #f3f3f6 1%, #f3f3f6 99%, #d9dfe3 99%, #ffffff 100%);
+  background: linear-gradient(left, #ffffff 0%, #d9dfe3 1%, #f3f3f6 1%, #f3f3f6 99%, #d9dfe3 99%, #ffffff 100%);
+  background: linear-gradient(to right, #ffffff 0%, #d9dfe3 1%, #f3f3f6 1%, #f3f3f6 99%, #d9dfe3 99%, #ffffff 100%);
+  background: linear-gradient(to bottom, #ffffff 0%, #d9dfe3 1%, #f3f3f6 1%, #f3f3f6 99%, #d9dfe3 99%, #ffffff 100%);
```
kinow, line 27: I think the first two `-moz-` and `-webkit` are browser-specific. Not too sure about the other three. Aren't they replacing each other, and only the last one is in effect?

annif/static/css/fonts.css
```diff
@@ -0,0 +1,21 @@
...
+
+/* jost-500 - latin-ext_latin */
+@font-face {
+  font-family: 'Jost';
+  font-style: normal;
+  font-weight: 500;
+  src: local(''),
+       url('../fonts/Jost-500-Medium.otf') format('opentype'),
+       url('../fonts/Jost-500-Medium.ttf') format('truetype');
+}
```
kinow, line 21: Missing newline at the end of the file

osma:
Thanks @kinow, good points!

annif/templates/home.html
osma: Good catch! The order was a bit off

annif/static/css/style.css
osma: You're right. The browser-specific ones are hardly needed anymore and the others are just unnecessary duplicates

annif/static/css/fonts.css
osma: Good catch!

osma:
The "Get suggestions" button isn't toggled as disabled/enabled like the one on annif.org, but other than that, I think it's close enough. We can do additional improvements in subsequent PRs as necessary

PR #438, opened 2020-08-21
Lexical STWFSAPY Backend

mo-fu:
Here is the first shot at adding the lexical backend developed at ZBW to Annif. Please suggest changes and missing tests.

*The algorithm works best for the English language or languages that are not morphologically rich.
*Currently it works only with short texts. My current understanding is that this is due to the lack of global features, i.e, every match by the finite state automaton is transformed into features but does not take into account other matches. If there are many matches in a text, all of them will get reported and scored individually. This will drive down precision.
*I tested the algorithm by using the tutorial data sets. I removed the small sample of titles from the larger one to have a training and test set. In the YSO case I also needed to manually edit the ontology because one label of https://finto.fi/yso/en/page/p37741 is malformated. [This breaks the automaton construction](https://github.com/zbw/stwfsapy/issues/23).

I used the following config:

```
[stwfsapy-yso] 
name=STWFSAPY YSO 
language=en
backend=stwfsapy
vocab=yso  
concept_type_uri=http://www.w3.org/2004/02/skos/core#Concept
sub_thesaurus_type_uri=http://www.w3.org/2004/02/skos/core#Collection
thesaurus_relation_type_uri=http://www.w3.org/2004/02/skos/core#member 
thesaurus_relation_is_specialisation=True
simple_english_plural_rules=True
graph_path=/home/fuer/Annif-tutorial/data-sets/yso-nlf/yso-skos.rdf
```

And got the following results:
Metric | Value
---|---:
Precision (doc avg):     |            0.3539953809523809 
Recall (doc avg):         |      0.15510352654022555  
F1 score (doc avg):    |         0.19221541052157337
Precision (subj avg):   |        0.14931225045169869 
Recall (subj avg):         |     0.07734734983512383
F1 score (subj avg):      |       0.08407773848545441
Precision (weighted subj avg): |  0.45994831973792344 
Recall (weighted subj avg):   |   0.12177820581040807
F1 score (weighted subj avg):  |  0.16307901726109417
Precision (microavg):       |     0.42555643541955374 
Recall (microavg):        |       0.12177820581040807 
F1 score (microavg):     |        0.1893667795628981
F1@5:                  |          0.19213065080346786
NDCG:                |            0.1993514066034938
NDCG@5:             |             0.22009559912085086
NDCG@10:           |              0.20175143141259658
Precision@1:           |          0.39159 
Precision@3:            |         0.35656666666666664
Precision@5:          |           0.354183
LRAP:                    |        0.1399838756074559
True positives:           |       60189
False positives:          |       81247
False negatives:        |         434062
Documents evaluated:        |     100000

It is a little better for the ZBW data set (document average F1 ~0.25). For our internal data set I got an improvement from 0.18 to 0.25 document average F1 when compared to the default configuration of Maui Server


osma:
Thanks for the PR!

The eval results are a bit hard to read, any chance you could reformat the block, for example by indenting it?

Also I see that stwfsapy is on PyPI, but there is no dependency declared in setup.py. Am I correct that the implementation is pure Python i.e. doesn't require any compiled extensions? In that case, I think it could be added as a core dependency, not an optional one.

mo-fu:
> 
> 
> Thanks for the PR!
> 
> The eval results are a bit hard to read, any chance you could reformat the block, for example by indenting it?
> 
> Also I see that stwfsapy is on PyPI, but there is no dependency declared in setup.py. Am I correct that the implementation is pure Python i.e. doesn't require any compiled extensions? In that case, I think it could be added as a core dependency, not an optional one.

Just forgot the setup.py.
Yes, it is a pure python implementation. I can change it to a default backend if  you want.
The original post should also have better readability now.

juhoinkinen:
I did some test runs with this backend, evaluating on kirjaesittelyt set (book presentations, 130 short texts [like this](https://www.wsoy.fi/kirja/liisa-jokinen/live-like-a-finn/9789510438169)) while training on kirjaesittelyt, JYU theses or Finna metadata sets.

Here are the F1@5 scores:
|train set | F1@5|
|-----|---|
|kirjaesittelyt|0.1697
|JYU theses|0.1536
|Finna|0.1714

All these scores are better than for Maui (0.1472) that has been trained on full-text documents from different collections.

osma:
Apologies @mo-fu for taking so long to review this. As @juhoinkinen already noted, this is giving promising results - it appears to be somewhat better than Maui on short documents, as you said.

Some quick findings:

# rdflib version compatibility

stwfsapy seems to [require rdflib 4.2.*](https://github.com/zbw/stwfsapy/blob/master/setup.py#L28), while Annif currently doesn't depend on a specific version. In practice when you install Annif from scratch, you will most likely end up with rdflib 5.0.0 which is the most recent version currently. This causes a version conflict if you then try to install stwfsapy and needs to be resolved manually (at least for me) by downgrading rdflib.

I think the best option here would be for both stwfsapy and Annif to depend on the most recent version (e.g. 5.0.*). In my experience upgrading to rdflib 5.0.0 shouldn't cause any major compatibility issues for most projects that use rdflib (though I did have to do some fixes in Skosify as it did some special tricks with namespaces).

# Fragile handling vocabulary terms

At least with YSO (until we fixed it), and also with a snapshot version of the NAL Thesaurus I tried, some special characters (esp. unmatched braces) seem to be causing parse errors in stwfsapy. Would it be possible to make it more robust?

# Matching all upper case letters

I happened to notice a [regular expression for matching upper case characters](https://github.com/zbw/stwfsapy/blob/master/stwfsapy/text_features.py#L83) in the stwfsapy code base. It hardcodes the set as `[A-ZÄÖÜ]`. This seems a bit limited, since there are many other upper case letters as well (e.g. Å in Swedish and Õ in Estonian). I know stwfsapy currently targets only English, but these characters may appear in names (and you already include `ÄÖÜ`). Unfortunately the standard library `re` module is a bit limited, but it would be possible to use the `regex` module from PyPI (for this an other solutions, see [this SO thread](https://stackoverflow.com/questions/36187349/python-regex-for-unicode-capitalized-words))

# Backend name

We talked about this already - I'd hope to see a shorter name for the backend, as `stwfsapy` is a bit unwieldy (think about recording a tutorial video about it and having to pronounce it many times!). Maybe the backend could have a shorter name - just shortening it to `stwfsa` would already be an improvement.

annif/backend/stwfsapy.py
```diff
@@ -0,0 +1,119 @@
...
+            raise NotSupportedException(
+                'Training stwfsapy project from cached data not supported.')
+        if corpus.is_empty():
+            raise NotSupportedException(
+                'Cannot train stwfsapy project with no documents.')
+        self.debug("Transforming training data.")
+        X = [doc.text for doc in corpus.documents]
+        y = [doc.uris for doc in corpus.documents]
+        graph = Graph()
+        graph_path = params[_KEY_GRAPH_PATH]
```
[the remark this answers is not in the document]
osma: That's good.

annif/backend/stwfsapy.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _train(self, corpus, params):
+        if corpus == 'cached':
+            raise NotSupportedException(
+                'Training stwfsapy project from cached data not supported.')
+        if corpus.is_empty():
+            raise NotSupportedException(
+                'Cannot train stwfsapy project with no documents.')
+        self.debug("Transforming training data.")
+        X = [doc.text for doc in corpus.documents]
```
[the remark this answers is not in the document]
osma: Now fixed, thanks!

osma:
Ah, forgot this one:

>Regarding the CI. I can make it a default dependency or add it to the CI depending on the python version and use import_or_skip like the other optional backends. Your decision.

Since this backend and stwfsapy only introduce pure Python dependencies, and in my understanding they support all the Python versions Annif currently supports (3.6-3.8), I think adding it as a default dependency would be OK.

mo-fu:
I can probably work on the issues starting next week, when our annual review has started. Assuming it goes smoothly.

mo-fu:
PR for more robus bracket handling: https://github.com/zbw/stwfsapy/pull/27 I could add only ZBW organization members as reviewers, but please have a look.

mo-fu:
Next PR for stwfsapy with changes for better upper case detection is here: https://github.com/zbw/stwfsapy/pull/29 Please request any changes necessary for adding stwfsa to Annif.

mo-fu:
After changing the rdflib dependency version (https://github.com/zbw/stwfsapy/commit/a654264d9315c3edc3a183b4bdfd00c0dbee8c4e ) I renamed the backend to stwfsa. There is a conflict in `setup.py`.
Do you prefer merges or rebases for resolving conflicts?

osma:
> I renamed the backend to stwfsa. 

Great!

> Do you prefer merges or rebases for resolving conflicts?

Whatever is easiest for you...there were some dependency updates recently, which is probably the cause for the conflict.

mo-fu:
The requested changes should all be done now. Let me know, if I missed something or further changes are needed.

osma:
Thanks @mo-fu , I'll take a look soon!

One thing that certainly needs to be done is to create a wiki page for the new backend. Like the other similar pages it should briefly explain what the backend does, how to set it up, and what the configuration parameters are.

...which brings up the issue of config parameters. The backend seems to take quite a few parameters, based on the config example you gave above. Are all of them necessary? Are there any default values? I think it would be helpful if most of the parameters had sane defaults that work with a typical, simple SKOS thesaurus.

You may want to take a look at the new MLLM backend, which aims to become a replacement for Maui. I just opened a draft PR #462 with some initial code. There's a tiny bit of overlap there with this PR since MLLM also needs the `as_graph` functionality - I simply copied that code from this PR. This may cause a merge conflict down the line but it should be trivial to resolve - don't worry about it now.

tests/corpora/archaeology/yso-archaeology.rdf
```diff
@@ -1119,4 +1119,140 @@
     <skos:altLabel xml:lang="sv">sigillvetenskap</skos:altLabel>
     <skos:prefLabel xml:lang="sv">sigillografi</skos:prefLabel>
   </skos:Concept>
+    <isothes:ConceptGroup rdf:about="http://www.yso.fi/onto/yso/p26593">
```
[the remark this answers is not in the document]
osma: Do you still need to introduce this ConceptGroup into yso-archaeology.rdf for the rest of the PR to work, or can this change now be dropped?

annif/backend/stwfsa.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def initialize(self):
+        if self._model is None:
+            path = os.path.join(self.datadir, self.MODEL_FILE)
+            self.debug(f'Loading STWFSA model from {path}.')
+            if os.path.exists(path):
+                self._model = StwfsapyPredictor.load(path)
+                self.debug('Loaded model.')
+            else:
+                raise NotInitializedException(
```
osma, line 71: This line 71 was not covered by the tests. I suggest that you add a test that checks this case, similar to [test_omikuji_suggest_no_vectorizer](https://github.com/NatLibFi/Annif/blob/267cc8518e392530a543995cef6df7c0caf6273b/tests/test_backend_omikuji.py#L28)

annif/backend/stwfsa.py
```diff
@@ -0,0 +1,121 @@
...
+                if key in self.STWFSA_PARAMETERS
+            }
+        new_params.pop(_KEY_INPUT_LIMIT)
+        p = StwfsapyPredictor(
+            graph=graph,
+            langs=frozenset([params['language']]),
+            **new_params)
+        p.fit(X, y)
+        self._model = p
+        p.store(os.path.join(self.datadir, self.MODEL_FILE))
```
osma, line 102: You could use `annif.util.atomic_save` here (like many other backends do) but no big deal if you don't. The idea of `atomic_save` is to first save the model (or whatever) into a temporary file, then rename it to the final name. If the program crashes during the save, it won't leave behind a broken model file.

setup.py
```diff
@@ -34,7 +34,8 @@ def read(fname):
         'rdflib',
```
osma, line 34: Can you please add here the same rdflib version declaration that stwfsapy uses? Currently it [seems to be](https://github.com/zbw/stwfsapy/blob/master/setup.py) `rdflib>=4.2,<6.0`
This way we can avoid version conflicts, for example when rdflib 6.0 is eventually released.

annif/backend/stwfsa.py
```diff
@@ -0,0 +1,121 @@
...
+_KEY_INPUT_LIMIT = 'input_limit'
+
+
+class StwfsaBackend(backend.AnnifBackend):
+
+    name = "stwfsa"
+    needs_subject_index = True
+
+    STWFSA_PARAMETERS = {
+        _KEY_CONCEPT_TYPE_URI: str,
```
osma, line 31: I think this should default to skos:Concept

osma (changes requested):
I did a quick check of the code and gave some comments. I think this is getting really close to merging, basically we just need to work out the best default values (and corresponding documentation) plus a few other small details.

osma:
A couple more, pretty minor issues:

1. Code Climate has [some complaints](https://codeclimate.com/github/NatLibFi/Annif/pull/438) about the PR. You can ignore the duplicate code warnings (the fasttext backend has similar code for handling defaults, but that's OK), but would it be easy to reduce the cyclomatic complexity of the `_train` method? For example, the transforming of the corpus to the `X` and `y` arrays could perhaps be extracted to a separate (static?) helper method or function, and/or introducing the `graph` variable could be avoided by just calling `as_graph` inline.
2. I see that you've improved the [wiki page for the backend](https://github.com/NatLibFi/Annif/wiki/Backend:-STWFSA), great! Now that the concept type defaults to `skos:Concept`, I think that those settings can be dropped from the example configurations?

mo-fu:
> 
> 
> A couple more, pretty minor issues:
> 
>     1. Code Climate has [some complaints](https://codeclimate.com/github/NatLibFi/Annif/pull/438) about the PR. You can ignore the duplicate code warnings (the fasttext backend has similar code for handling defaults, but that's OK), but would it be easy to reduce the cyclomatic complexity of the `_train` method? For example, the transforming of the corpus to the `X` and `y` arrays could perhaps be extracted to a separate (static?) helper method or function, and/or introducing the `graph` variable could be avoided by just calling `as_graph` inline.
Should be fixed now
>     2. I see that you've improved the [wiki page for the backend](https://github.com/NatLibFi/Annif/wiki/Backend:-STWFSA), great! Now that the concept type defaults to `skos:Concept`, I think that those settings can be dropped from the example configurations?
I kept it for the second configuration example and added some more details to the description of the second example.

osma (approved):

osma:
Excellent! Merging this now. :rocket:

osma:
Since this is now merged, I added links to the backend wiki page from the wiki front page and the sidebar.

PR #440, opened 2020-08-26
Selectable language for suggestion labels 

juhoinkinen:
Adds a menu for selecting the language of the labels of the suggestions. Default selection is "text language", in which case the labels are not changed, but the ones coming from http://ai.finto.fi (Annif) are retained. If a specific language is selected, then the URIs of the suggestions are used to get the labels from http://api.finto.fi with the selected language. If something fails in the query, an alert window is shown.


juhoinkinen:
At least this case remains to be solved: if there is no term for a concept in a language, like for [Pohjanmaa in English](https://finto.fi/yso-paikat/en/page/?clang=en&uri=p94344), that label is not shown in the list of suggestions (or an empty string is shown):

image

ai.finto.fi/static/js/i18n.js
```diff
@@ -50,6 +50,11 @@ jQuery(function() {
             "nav-subject-indexing": "Subject indexing",
             "project": "Vocabulary and text language",
             "limit": "Maximum # of suggestions",
+            "label-language": "Labels language",
```
osma, line 53: "Term language" could be better?

ai.finto.fi/static/js/i18n.js
```diff
@@ -50,6 +50,11 @@ jQuery(function() {
...
+            "label-language-option-project": "project",
```
osma: "text language" ?

ai.finto.fi/static/js/i18n.js
```diff
@@ -72,6 +77,11 @@ jQuery(function() {
             "nav-subject-indexing": "Innehållsbeskrivning",
             "project": "Vokabulär och textens språk",
             "limit": "Maximalt antal förslag",
+            "label-language": "Förslag språk",
```
osma, line 80: "Termernas språk"

ai.finto.fi/static/js/i18n.js
```diff
@@ -72,6 +77,11 @@ jQuery(function() {
...
+            "label-language-option-project": "projekt",
```
osma: "textens språk"

ai.finto.fi/static/js/i18n.js
```diff
@@ -95,6 +105,11 @@ jQuery(function() {
             "nav-subject-indexing": "Sisällönkuvailu",
             "project": "Sanasto ja tekstin kieli",
             "limit": "Ehdotusten enimmäismäärä",
+            "label-language": "Termien kieli",
+            "label-language-option-project": "projekti",
```
osma: "tekstin kieli"

osma (changes requested):
Looks good! Some suggestions for the translations below

osma:
> At least this case remains to be solved: if there is no term for a concept in a language, like for Pohjanmaa in English, that label is not shown in the list of suggestions (or an empty string is shown):

Maybe you could do the same that Finto/Skosmos does in similar situations: use the untranslated value, with the language in parentheses. So this would become `Pohjanmaa (fi)`

PR #441, opened 2020-09-29
Sparse subjects

mo-fu:
I observed that by using sparse matrices for the gold subjects, evaluation time can be reduced. 
 Note that ranking metrics (when they use argsort) are not compatible with sparse matrices. Sparse matrices also do not support boolean operators and computing the elementwise inversion of a boolean sparse matrix, would probaly lead to a non sparse result.
Here is a simple test using an omikuji project with different sized prefixes of the `yso-finna-small.tsv` from the tutorial repo.
All measurements are from single runs so take with a grain of salt and try to reproduce it.
| size of prefix | time without sparse matrices| time with sparse matrices|
|--------------|-----------------------------:|--------------------------:|
1000|` 83.73s user 6.59s system 137% cpu 1:05.47 total`|`69.17s user 4.75s system 151% cpu 48.648 total`|
2000|`166.30s user 12.25s system 138% cpu 2:09.16 total `|`142.23s user 10.75s system 154% cpu 1:38.92 total`|
5000|`447.16s user 25.58s system 142% cpu 5:30.73 total`|`350.88s user 18.83s system 158% cpu 3:52.88 total`|
10000|`868.23s user 50.75s system 141% cpu 10:47.59 total`|`706.70s user 33.58s system 160% cpu 7:41.34 total`|
100000|`9125.41s user 1792.14s system 133% cpu 2:16:42.95 total`|`7306.73s user 863.42s system 151% cpu 1:30:04.99 total`|

All times are including data loading and prediction.
In my opinion the numbers show that it never hurts to compute the sparse matrix.

For non zero thresholds it may make sense to also compute a sparse variant of the predictions after the cut-off.
I have not checked, whether it makes a difference to first compute the sparse matrix and then convert it to a dense one.
The approach in this PR seemed to be the low hanging fruit.


osma:
Thanks, looks very promising indeed! Also, good catch eliminating the unused NDCG metrics from the `optimize` command (not the `eval` command - the commit message was a bit confusing at first). That should speed up the `optimize` command as well.

We will try to replicate the comparison.

osma:
I tried this on a few different data sets and the performance increase seems to be consistent. Memory usage does not increase. I think this is good enough for merging - we will be doing some more evaluation runs in the near future. It's unlikely that there would be any adverse effects, but that should catch them.

Thanks a lot - and if you can find any further optimization opportunities, those are very welcome!

PR #444, opened 2020-10-27
Commit transaction and reopen LMDB read-only for NN ensemble training

### LINKED ISSUE #443: Allow use of cached data after cancelled training of nn_ensemble backend
When cancelling training of the `nn_ensemble` backend after the data transformation was completed, the backend is not able to use the cached data written to the lmdb file.
A quick fix for me was to disable locking when opening the lmdb file but that opens another can of worms.
Another possible solution could be to close the db after writing all the data and reopening it in readonly mode. Have not tested yet whether this would do the trick.

osma:
Fixes #443

Since the LMDB is closed before starting the TensorFlow model training, it doesn't matter if the training gets aborted. The LMDB can still be reused for another training run.

I'd appreciate if you could test this @mo-fu but I understand if you don't have time

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -184,14 +184,16 @@ class NNEnsembleBackend(
 
     def _fit_model(self, corpus, epochs):
         env = self._open_lmdb(corpus == 'cached')
-        with env.begin(write=True, buffers=True) as txn:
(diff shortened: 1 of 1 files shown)


mo-fu:
With the in this PR, I was able to reuse the converted data when canceling training after neural network training has started.

osma:
Thanks @mo-fu !

PR #446, opened 2020-10-29
Add a backend paratemer to limit input characters in suggest

juhoinkinen:
As noted in a CSC's recent presentation on new possible algorithms it can be advantageous to make the suggestions based only on the beginning part of a text (it typically includes an abstract and/or introduction). Also quick tests on JYU thesis showed quite big improvements of F1@5 scores when using only the first 5000 characters of each document.

This PR adds a backend parameter `input_limit` that can be used to pass only N first characters of a document to a backend for making suggestions. With the default value zero the limiting is not performed (maybe `None` would have been a more proper value for this but using zero kept code a bit more simpler). Could there be a better name for the parameter? Maybe `input_character_limit`?

I noted that the default parameters from [AnnifBackend](https://github.com/NatLibFi/Annif/blob/master/annif/backend/backend.py#L17) do not propagate as intended to PAV and nn_ensemble backends, because PAV [defines its own default parameters](https://github.com/NatLibFi/Annif/blob/master/annif/backend/pav.py#L27) without merging/updating the ones from AnnifBackend, and with [nn_ensemble there is similar case](https://github.com/NatLibFi/Annif/blob/master/annif/backend/nn_ensemble.py#L100). This is why the relevant tests now need the `input_limit` in `config_params`.

### DIFF
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -14,7 +14,8 @@ class AnnifBackend(metaclass=abc.ABCMeta):
     name = None
     needs_subject_index = False
 
-    DEFAULT_PARAMETERS = {'limit': 100}
+    DEFAULT_PARAMETERS = {'limit': 100,
+                          'input_limit': 0}
 
     def __init__(self, backend_id, config_params, project):
         """Initialize backend with specific parameters. The
@@ -80,6 +81,8 @@ class AnnifBackend(metaclass=abc.ABCMeta):
         represented as a list of SubjectSuggestion objects."""
         beparams = self._get_backend_params(params)
         self.initialize()
+        if int(beparams['input_limit']) != 0:
+            text = text[:int(beparams['input_limit'])]
         return self._suggest(text, params=beparams)
 
     def debug(self, message):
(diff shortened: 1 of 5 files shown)


juhoinkinen:
Ok, I make a bug fix PR for the default parameters and then come back to this...

annif/backend/backend.py
```diff
@@ -80,6 +81,8 @@ def suggest(self, text, params=None):
         represented as a list of SubjectSuggestion objects."""
         beparams = self._get_backend_params(params)
         self.initialize()
+        if int(beparams['input_limit']) != 0:
```
osma, line 84: What happens if `input_limit` is set to zero in the configuration file? The string `0` is different from 0, so the `if` clause is true. Will the input then be truncated to zero length?

osma (commented):
Needs documentation in the wiki. Other than the above comment, looks good to me!

annif/backend/backend.py
juhoinkinen: That's no problem, the `input_limit` is converted to `int`, so even if it is string from the configuration file, the condition works as expected.

juhoinkinen: So no truncation is performed.

osma: You're right

juhoinkinen:
Added mentions for `input_limit` in these wiki pages:
- [Project configuration](https://github.com/NatLibFi/Annif/wiki/Project-configuration) 
- [Achieving good results](https://github.com/NatLibFi/Annif/wiki/Achieving-good-results#4-configure-and-train-some-basic-projectsmodelsbackends).

PR #447, opened 2020-10-29
Fix missing default params in pav and nn ensemble

juhoinkinen:
As noted in #446 there was a bug concerning the default parameters in PAV and nn_ensemble. 

> > and with nn_ensemble there is similar case.
> 
> I don't see the problem there? The default parameters are combined from the superclass and the current class.

For some reason the default parameters from AnnifBackend were not present in NNEnsembleBackend (for now only the `limit` parameter). Seems like even if the "right method" i.e. `default_params(self)` from AnnifBackend was in use via `super()`, the `self` in this case was the object of NNEnsembleBackend class, which has its own `DEFAULT_PARAMETERS` field.

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -98,8 +98,7 @@ class NNEnsembleBackend(
     _model = None
 
     def default_params(self):
-        params = {}
-        params.update(super().default_params())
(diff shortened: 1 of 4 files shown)


osma (approved):
Looks good to me! Good catch on nn_ensemble, it wasn't obvious that there was a problem.

PR #448, opened 2020-11-02
Upgrade dependencies

juhoinkinen:
Upgrades multiple dependencies:
```
joblib 0.14.* -> 0.17.0
scikit-learn 0.22.* -> 0.23.2
scipy 1.4.1 -> 1.5.3
numpy 1.17.* -> 1.18.*
optuna 1.5.* -> 2.2.0
vowpalwabbit 8.7.* -> 8.8.1
tensorflow-cpu 2.3.0 -> 2.3.1
```
There is still the question of pytest (and pep8), if the latest pytest version `6.1.2` is used, the following error occurs:
```
_____________________________________________ ERROR collecting test session ______________________________________________
Direct construction of Pep8Item has been deprecated, please use Pep8Item.from_parent.
See https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent for more details.
```
This is due to the [pytest-pep8 plugin](https://pypi.org/project/pytest-pep8/#history). Options seem to be:
- Pin pytest to `5.4.*`
- Switch from `pytest-pep8` to some other linter tool, e.g. to [`pytest-flake8`](https://pypi.org/project/pytest-flake8/), as done [here](https://github.com/larray-project/larray/issues/889) for example (`pytest-flake8` seemed to work quite straightforwardly with Annif too in quick test.)


juhoinkinen:
One additional "dependency" control that could be added is a requirement for Python version in `setup.py`, with e.g. `python_requires='>=3.6'`. This would stop installation attempts on a not-supported Python version without delay: 
```
$ pip install .[dev]
Processing /home/jmminkin/git/Annif
ERROR: Package 'annif' requires a different Python: 3.5.2 not in '>=3.6'
```
(Noted demand for this at kj-kk.)

osma:
> One additional "dependency" control that could be added is a requirement for Python version in setup.py

Sounds like a good idea!

osma:
> Switch from pytest-pep8 to some other linter tool, e.g. to pytest-flake8

I think that would be much better than relying on unmaintained pytest-pep8 and an obsolete version of pytest (5.x), so +1 for switching, if it's straightforward

osma (approved):
The changes look good, +1 for merging.

My only concern is how the library upgrades will affect existing trained models. Based on earlier experiences I would guess that at least the tfidf vectorizers used by the `tfidf` and `omikuji` backends will cause warnings when they are loaded with the newer scikit-learn version. This should be investigated (test if old models still work or not, and note any warning messages) and explained in the release notes for the next release.

juhoinkinen:
> The changes look good, +1 for merging.
> 
> My only concern is how the library upgrades will affect existing trained models. Based on earlier experiences I would guess that at least the tfidf vectorizers used by the `tfidf` and `omikuji` backends will cause warnings when they are loaded with the newer scikit-learn version. This should be investigated (test if old models still work or not, and note any warning messages) and explained in the release notes for the next release.

I ran `eval` with Omikuji and nn_ensemble models made using earlier Annif version, and the results do not change when switching to the upgraded package versions. As expected an (unwarranted) warning is printed out due to scikit upgrade:
```
/srv/experiments/check-annif-models-on-upgrade/Annif/venv-upgrade2/lib/python3.6/site-packages/sklearn/base.py:334: UserWa
rning: Trying to unpickle estimator TfidfTransformer from version 0.22.2.post1 when using version 0.23.2. This might lead 
to breaking code or invalid results. Use at your own risk.

**Comment by juhoinkinen** [comment]
So old modes seem to work with but upgraded libraries, but I noticed something weird. 

When installing `pip install .[nn]` in a new venv (Python 3.6.9, at annif-kk) I got errors messages:
```
...
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

google-auth 1.23.0 requires setuptools>=40.3.0, but you'll have setuptools 39.0.1 which is incompatible.
tensorboard 2.3.0 requires setuptools>=41.0.0, but you'll have setuptools 39.0.1 which is incompatible.
tensorflow-cpu 2.3.1 requires numpy<1.19.0,>=1.16.0, but you'll have numpy 1.19.4 which is incompatible.
Successfully installed absl-py-0.11.0 annif-0.50.0.dev0 astunparse-1.6.3 cachetools-4.1.1 gast-0.3.3 google-auth-1.23.0 google-auth-oauthlib-0.4.2 google-pasta-0.2.0 grpcio-1.33.2 h5py-2.10.0 keras-preprocessing-1.1.2 lmdb-1.0.0 markdown-3.3.3 oauthlib-3.1.0 opt-einsum-3.3.0 protobuf-3.13.0 pyasn1-0.4.8 pyasn1-modules-0.2.8 requests-oauthlib-1.3.0 rsa-4.6 tensorboard-2.3.0 tensorboard-plugin-wit-1.7.0 tensorflow-cpu-2.3.1 tensorflow-estimator-2.3.0 termcolor-1.1.0 wheel-0.35.1 wrapt-1.12.1
```

The worrying part is is the 
```
tensorflow-cpu 2.3.1 requires numpy<1.19.0,>=1.16.0, but you'll have numpy 1.19.4 which is incompatible
```
but still that package (and v2.3.1 of it) is successfully installed, and tests for nn_ensemble pass. This actually happens also in [Travis run](https://travis-ci.com/github/NatLibFi/Annif/jobs/426982498). Seems weird at least. 

I got rid of the messages about setuptools by upgrading it to 50.3.2. 

(Also, I checked what happens when using the `--use-feature=2020-resolver` option as recommended in the message, and then the installation fails:
```
Collecting gunicorn
  Downloading gunicorn-0.16.0.tar.gz (130 kB)
     |████████████████████████████████| 130 kB 37.2 MB/s 
    ERROR: Command errored out with exit status 1:
     command: /srv/experiments/check-annif-models-on-upgrade/Annif/venv-upgrade2/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-nt_tr9iz/gunicorn_6c381e3be78342189e379a2a53ac6f74/setup.py'"'"'; __file__='"'"'/tmp/pip-install-nt_tr9iz/gunicorn_6c381e3be78342189e379a2a53ac6f74/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-a1c03m8y
         cwd: /tmp/pip-install-nt_tr9iz/gunicorn_6c381e3be78342189e379a2a53ac6f74/
    Complete output (5 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-nt_tr9iz/gunicorn_6c381e3be78342189e379a2a53ac6f74/setup.py", line 43, in <module>
        with open(fname) as f:
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-install-nt_tr9iz/gunicorn_6c381e3be78342189e379a2a53ac6f74/requirements_dev.txt'
```
However this seems to be unrelated (and why here `gunicorn-0.16.0.tar.gz` is downloaded, but gunicorn 20.0.4 is installed when not using the `2020-resolver` option...).)

osma:
I think while we are upgrading dependencies, we should do it in a way that works also with the new pip algorithm.

> tensorflow-cpu 2.3.1 requires numpy<1.19.0,>=1.16.0, but you'll have numpy 1.19.4 which is incompatible

How about keeping numpy at a slightly older version (1.18.x probably)? Even if 1.19.4 happens to work for us, I don't think it's a good idea to upgrade to a version that is not officially supported by the TensorFlow version we are using.

I have no idea about the gunicorn related issue, seems weird...

juhoinkinen:
Well I can't reproduce the `gunicorn` error again, could have been due to old `setuptools`, and in any case it arised only when using that option. Merging now.

PR #452, opened 2020-11-17
Apply the input_limit backend parameter to texts in train & learn 

juhoinkinen:
…learn

Would this approach be suitable? Another way would be to apply the limit when each backend uses the document texts.

Now I see the limit would need to be applied also in [`DocumentList` class](https://github.com/NatLibFi/Annif/blob/input-limit-for-corpus-document-texts/annif/corpus/document.py#L91) as it is used by REST learn method.

### DIFF
--- a/tests/test_backend_tfidf.py
+++ b/tests/test_backend_tfidf.py
@@ -35,7 +35,7 @@ def test_tfidf_train(datadir, document_corpus, project):
     assert datadir.join('tfidf-index').size() > 0
 
 
-def test_tfidf_train_input_limited(datadir, document_corpus, project):
+def test_tfidf_train_input_limited(document_corpus, project):
     tfidf_type = annif.backend.get_backend("tfidf")
     tfidf = tfidf_type(
         backend_id='tfidf',
@@ -48,7 +48,7 @@ def test_tfidf_train_input_limited(datadir, document_corpus, project):
         in str(excinfo)
 
 
-def test_tfidf_train_negative_input_limit(datadir, document_corpus, project):
+def test_tfidf_train_negative_input_limit(document_corpus, project):
     tfidf_type = annif.backend.get_backend("tfidf")
     tfidf = tfidf_type(
         backend_id='tfidf',
--- a/tests/test_backend_tfidf.py
+++ b/tests/test_backend_tfidf.py
@@ -1,5 +1,6 @@
 """Unit tests for the TF-IDF backend in Annif"""
 
+import pytest
 import annif
 import annif.backend
 import annif.corpus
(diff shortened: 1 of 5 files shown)


osma:
Not sure I like setting `corpus.text_limit` from outside. It seems wrong to change the corpus after it has been passed to the project.

How about a wrapper class, similar to `DocumentList`, that wraps an existing `Document*` class and applies the limit? Something like

```python
class TruncatingDocumentCorpus(DocumentCorpus):
    """A document corpus that wraps another document corpus
       but truncates the documents to a given length"""

    def __init__(self, documents, limit):
        self._documents = documents
        self._limit = limit

    @property
    def documents(self):
        # left as an exercise for the reader ;)
```

Then in the `train` and `learn` methods this could be applied like:

```python
    if int(beparams['input_limit']) != 0:
        corpus = TruncatingDocumentCorpus(corpus, int(beparams['input_limit']))
```

This way the corpus itself doesn't need to be mutated, and the limiting only has to be implemented in one place (the wrapper class).

tests/test_backend_tfidf.py
```diff
@@ -33,6 +34,19 @@ def test_tfidf_train(datadir, document_corpus, project):
     assert datadir.join('tfidf-index').size() > 0
 
 
+def test_tfidf_train_input_limited(datadir, document_corpus, project):
```
osma, line 37: This test is a bit funny...I thik I understand what happens here (the training docs get truncated to 1 character, thus there are no words to train on since the minimum length is 3 characters) but maybe it could be explained in a comment

osma (approved):
Looks good to me. The only worry I have is what happens if a negative `input_limit` is configured? In my understanding, documents will then be truncated from the end, so that only the tail remains. Is this desirable? I think there are two possibilities here:

1. Check if `input_limit` is negative and fail with a `ConfigurationException`
2. Document (e.g. in the TruncatingDocumentCorpus docstrings) that if a negative `input_limit` is set, the document will be truncated from the end instead of the beginning

Either way is fine with me.

juhoinkinen:
> Looks good to me. The only worry I have is what happens if a negative `input_limit` is configured? In my understanding, documents will then be truncated from the end, so that only the tail remains. Is this desirable? I think there are two possibilities here:
> 
>     1. Check if `input_limit` is negative and fail with a `ConfigurationException`
> 
>     2. Document (e.g. in the TruncatingDocumentCorpus docstrings) that if a negative `input_limit` is set, the document will be truncated from the end instead of the beginning
> 
> 
> Either way is fine with me.

When the `input_limit` is negative, its value sets the number of characters that are dropped (in contrast to the num. of chars that are retained). Not sure how useful that can be, but this is now explained in the docstring.

osma:
> When the input_limit is negative, its value sets the number of characters that are dropped (in contrast to the num. of chars that are retained). Not sure how useful that can be, but this is now explained in the docstring.

Right - now that you put it this way, I don't think it can be very useful. It would be perhaps slightly more useful to leave the tail part consisting of `abs(input_limit)` characters, like this: `text[-input_limit:]`

juhoinkinen:
> > When the input_limit is negative, its value sets the number of characters that are dropped (in contrast to the num. of chars that are retained). Not sure how useful that can be, but this is now explained in the docstring.
> 
> Right - now that you put it this way, I don't think it can be very useful. It would be perhaps slightly more useful to leave the tail part consisting of `abs(input_limit)` characters, like this: `text[-input_limit:]`

I switched to raising the `ConfigurationException` for negative `input_limit` values, now for suggest calls too.

PR #458, opened 2020-12-09
Improved handling of source weights in NN ensemble

### LINKED ISSUE #457: NN ensemble gives poor results with weighted sources
The NN ensemble tends to perform quite well when given source projects that have default weights (1.0), like this:

```
sources=yso-maui-fi,yso-fasttext-fi,yso-bonsai-fi
```

When evaluated on 11 different data sets, this type of NN ensemble achieved a F1@5 score of 0.4418.

However, when the sources are assigned weights, and in particular small weights that add up to 1 (for example generated using `annif hyperopt` with a plain ensemble, e.g. `sources=yso-maui-fi:0.2607,yso-fasttext-fi:0.5544,yso-bonsai-fi:0.1849`), the NN ensemble gives much poorer results (0.3758 for the same 11 data sets). This is also much worse than a plain weighted ensemble (0.4246 using the same weights) so the neural network makes the results **worse** instead of improving them, compared to a plain ensemble.

The problem seems to be the simplistic way that the NN ensemble averages the scores coming in from the source backends. A better averaging method that is scale-invariant (i.e. not sensitive to the scale of the given weights) would be needed.

PR with fixes coming up soon...

osma:
This PR contains fixes aiming at improving the way the NN ensemble handles sources that have been assigned non-default weights.

1. The weights assigned to source backends are now always normalized so that they add up to 1 (L1 normalization).
2. The NN ensemble multiplies incoming score vectors by the assigned weights. Since the normalized weights are now generally smaller than they used to be (e.g. not 1:1:1 but 0.33:0.33:0.33), they are multiplied - outside the neural network - by the number of sources to compensate for this change.
3. While researching this, I discovered that the NN ensemble performs better when the incoming weights are slightly larger numerically. Taking the square root of the scores (which increases the scores, since sqrt(x) > x when 0<x<1) before feeding them to the neural network seems to improve F1@5 scores by an additional 1-2 percentage points!

Fixes #457

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -115,7 +115,7 @@ class NNEnsembleBackend(
         self._model = load_model(model_filename)
 
     def _merge_hits_from_sources(self, hits_from_sources, params):
-        score_vector = np.array([hits.as_vector(subjects) * weight
+        score_vector = np.array([hits.as_vector(subjects) * weight * len(hits_from_sources)
                                  for hits, weight, subjects
                                  in hits_from_sources],
                                 dtype=np.float32)
@@ -168,7 +168,7 @@ class NNEnsembleBackend(
             for source_project, weight in sources:
                 hits = source_project.suggest(doc.text)
                 doc_scores.append(
-                    hits.as_vector(source_project.subjects) * weight)
+                    hits.as_vector(source_project.subjects) * weight * len(sources))
             score_vector = np.array(doc_scores,
                                     dtype=np.float32).transpose()
             subjects = annif.corpus.SubjectSet((doc.uris, doc.labels))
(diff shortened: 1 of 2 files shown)


osma:
There's also the question what happens if you use an old model, trained with the current/old NN ensemble code, while running the new code. I tested that with the example model from #457 that originally achieved an average F1@5 score of 0.4418; when evaluated with the new code, the score was 0.4330, which is a bit worse, but not catastrophically so - and in fact for some of the individual data sets the scores had improved.

I think it's enough to state in the release notes that retraining NN ensemble models is recommended, but not absolutely necessary.

juhoinkinen (approved):

osma:
I verified that this PR doesn't change the results of a simple ensemble. (The `annif.util.parse_sources` that was modified in this PR is used by all ensembles)

Still need to double-check that PAV isn't affected

osma:
I verified this once more with the Annif tutorial data sets. The results improved significantly for the STW data set, but declined slightly for the YSO one (though depends on metric).

Double-checked that PAV is unaffected. Merging this now.

Issue #224, opened 2018-12-07
Suggestions for new terms


hekl:
Is their a possibility to get suggestions for terms that are important in a document or a set of documents, but that are not present in the supported vocabulary?

osma:
Thanks, it's a great suggestion, but in its current form Annif is quite focused on recognizing existing concepts from a predefined vocabulary, not suggesting new ones to be added to the vocabulary.

I think this could be added to the Maui backend, because it is based on a lexical approach and has to compare terms in the vocabulary to terms that appear in the text. For example the MauiService REST API could have a method for suggesting new terms/concepts. I seem to remember that such tools have previously been built over Maui's predecessor Kea (or Kea++).

hekl:
I have used Rake, and it can suggest the most interesting keywords in a document, but this did not connect to a vocabulary.

osma:
I wasn't avare of RAKE and that there's a Python implementation available, thanks for the hint. This could be a valuable addition for the future, so I'm reopening the issue.

hekl:
Of course both MAUI and Rake have been used by [Alyona Medelyan](https://www.airpair.com/nlp/keyword-extraction-tutorial). I found Rake working quite good, considering its elementary setup.

osma:
Superseded by #566

PR #461, opened 2021-01-12
YAKE backend

juhoinkinen:
This PR adds a new backend to Annif by integrating the [YAKE](https://github.com/LIAAD/yake) library. 

YAKE performs unsupervised automatic keyword extraction, and in the Annif backend the keywords found by YAKE are searched from the SKOS vocabulary labels, and the matches are returned as subject suggestions. The search can be targeted to prefLabels, altLabels and/or hiddenLabels as set in project configuration. 

The YAKE backend is based on lexical principle, but does not perform as well as the other lexical backends (MLLM, STWFSA or Maui) as measured by evaluation results. However, the (free) keyword extraction operation offers a possibility to add new features in Annif, especially the feature for suggesting new terms for a vocabulary (the keywords not found in the vocabulary), see #224. Also the unsupervised approach can be useful in some cases: there is no need for training data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,4 +133,7 @@ Zenodo DOI:
 
 The code in this repository is licensed under Apache License 2.0, except for the
 dependencies included under `annif/static/css` and `annif/static/js`,
-which have their own licenses. See the file headers for details. Using the optional Yake backend may change the licence of Annif to GPLv3, because [YAKE](https://github.com/LIAAD/yake) is licensed under GPLv3.
+which have their own licenses, see the file headers for details.
+Enabling the optional Yake backend may result in [GPLv3](https://www.gnu.org/licenses/gpl-3.0.txt)
+terms to cover the application, because [YAKE](https://github.com/LIAAD/yake)
+is licensed under GPLv3.
--- a/README.md
+++ b/README.md
@@ -134,6 +134,10 @@ Zenodo DOI:
 The code in this repository is licensed under Apache License 2.0, except for the
 dependencies included under `annif/static/css` and `annif/static/js`,
 which have their own licenses, see the file headers for details.
-Enabling the optional Yake backend may result in [GPLv3](https://www.gnu.org/licenses/gpl-3.0.txt)
-terms to cover the application, because [YAKE](https://github.com/LIAAD/yake)
-is licensed under GPLv3.
+Please note that the [YAKE](https://github.com/LIAAD/yake) library is licended
+under [GPLv3](https://www.gnu.org/licenses/gpl-3.0.txt), while Annif is
+licensed under the Apache License 2.0. The licenses are compatible, but
+depending on legal interpretation, the terms of the GPLv3 (for example the
+requirement to publish corresponding source code when publishing an executable
+application) may be considered to apply to the whole of Annif+Yake if you
+decide to install the optional Yake dependency.
--- a/README.md
+++ b/README.md
@@ -133,4 +133,4 @@ Zenodo DOI:
 
 The code in this repository is licensed under Apache License 2.0, except for the
 dependencies included under `annif/static/css` and `annif/static/js`,
-which have their own licenses. See the file headers for details.
+which have their own licenses. See the file headers for details. Using the optional Yake backend may change the licence of Annif to GPLv3, because [YAKE](https://github.com/LIAAD/yake) is licensed under GPLv3.
--- a/annif/backend/yake.py
+++ b/annif/backend/yake.py
@@ -48,14 +48,16 @@ class YakeBackend(backend.AnnifBackend):
         if type(self.params['label_types']) == str:  # Label types set by user
             label_types = [lt.strip() for lt
                            in self.params['label_types'].split(',')]
-            valid_types = ('prefLabel', 'altLabel', 'hiddenLabel')
-            for lt in label_types:
-                if lt not in valid_types:
-                    raise ConfigurationException(
-                        f'invalid label type {lt}', backend_id=self.backend_id)
+            self._validate_label_types(label_types)
         else:
             label_types = self.params['label_types']  # The defaults
         return [getattr(SKOS, lt) for lt in label_types]
+
+    def _validate_label_types(self, label_types):
+        for lt in label_types:
+            if lt not in ('prefLabel', 'altLabel', 'hiddenLabel'):
+                raise ConfigurationException(
+                    f'invalid label type {lt}', backend_id=self.backend_id)
 
     @property
     def graph(self):
--- a/annif/backend/yake.py
+++ b/annif/backend/yake.py
@@ -74,8 +74,8 @@ class YakeBackend(backend.AnnifBackend):
         return self._graph
 
     def _create_index(self):
-        # TODO Should index creation be done on loadvoc command?
-        # TODO American to British labels?
+        # TODO Should index creation & saving be done on loadvoc command?
+        # Or saving at all? It takes about 1 min to create the index
         index = defaultdict(set)
         for predicate in [SKOS.prefLabel, SKOS.altLabel, SKOS.hiddenLabel]:
             for concept in self.graph.subjects(RDF.type, SKOS.Concept):
@@ -104,8 +104,7 @@ class YakeBackend(backend.AnnifBackend):
         with open(path, 'r', encoding='utf-8') as indexfile:
             for line in indexfile:
                 label, uris = line.strip().split('\t')
-                uris = uris.split()
-                index[label] = uris
+                index[label] = uris.split()
         return index
 
     def _sort_phrase(self, phrase):
@@ -140,10 +139,7 @@ class YakeBackend(backend.AnnifBackend):
     def _keyphrase2uris(self, keyphrase):
         keyphrase = self._lemmatize_phrase(keyphrase)
         keyphrase = self._sort_phrase(keyphrase)
-        uris = []
-        uris.extend(self._index.get(keyphrase, []))
-
-        return uris
+        return self._index.get(keyphrase, [])
 
     def _transform_score(self, score):
         # TODO if score<0:
--- a/annif/backend/yake.py
+++ b/annif/backend/yake.py
@@ -65,12 +65,8 @@ class YakeBackend(backend.AnnifBackend):
     @property
     def graph(self):
         if self._graph is None:
-            # TODO use as_graph() that is now available
-            # self._graph = vocab.as_graph()
-            self._graph = rdflib.Graph()
-            path = os.path.join(self.project.vocab.datadir, 'subjects.ttl')
-            self.info('Loading graph from {}'.format(path))
-            self._graph.load(path, format=rdflib.util.guess_format(path))
+            self.info('Loading graph')
+            self._graph = self.project.vocab.as_graph()
         return self._graph
 
     def _create_index(self):
--- a/annif/backend/yake.py
+++ b/annif/backend/yake.py
@@ -194,9 +194,13 @@ class YakeBackend(backend.AnnifBackend):
                 combined_suggestions[uri] = score
             else:
                 old_score = combined_suggestions[uri]
-                combined_suggestions[uri] = self._conflate_scores(
+                combined_suggestions[uri] = self._combine_scores(
                     score, old_score)
         return list(combined_suggestions.items())
 
-    def _conflate_scores(self, score1, score2):
-        return score1 * score2 / (score1 * score2 + (1-score1) * (1-score2))
+    def _combine_scores(self, score1, score2):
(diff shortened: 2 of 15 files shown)


.travis.yml
```diff
@@ -26,9 +26,9 @@ install:
 # Install the optional neural network dependencies (TensorFlow and LMDB)
 # - except for one Python version (3.7) so that we can test also without them
 - if [[ $TRAVIS_PYTHON_VERSION != '3.7' ]]; then pip install .[nn]; fi
-# Install the optional Omikuji dependency
+# Install the optional Omikuji and Yake dependencies
 # - except for one Python version (3.7) so that we can test also without them
-- if [[ $TRAVIS_PYTHON_VERSION != '3.7' ]]; then pip install .[omikuji]; fi
+- if [[ $TRAVIS_PYTHON_VERSION != '3.7' ]]; then pip install .[omikuji] .[yake]; fi
```
osma, line 31: I think this could be expressed as `.[omikuji,yake]` but whatever...

README.md
```diff
@@ -133,4 +133,4 @@ Zenodo DOI:
 
 The code in this repository is licensed under Apache License 2.0, except for the
 dependencies included under `annif/static/css` and `annif/static/js`,
-which have their own licenses. See the file headers for details.
+which have their own licenses. See the file headers for details. Using the optional Yake backend may change the licence of Annif to GPLv3, because [YAKE](https://github.com/LIAAD/yake) is licensed under GPLv3.
```
osma, line 136: This doesn't sound right...we need to figure out the exact wording to use but importing Yake doesn't really *change* the license of Annif - what it may do is affect the terms under which the combination of Annif and Yake may be used and distributed.

annif/backend/yake.py
```diff
@@ -0,0 +1,196 @@
+"""Annif backend using Yake keyword extraction"""
+# TODO Mention GPLv3 license also here?
```
osma, line 2: Hmm...maybe it should be mentioned. At least we can't leave it as a TODO forever :)

annif/backend/yake.py
```diff
@@ -0,0 +1,196 @@
...
+
+    DEFAULT_PARAMETERS = {
+        'max_ngram_size': 4,
+        'deduplication_threshold': 0.9,
+        'deduplication_algo': 'levs',
+        'window_size': 1,
+        'num_keywords': 100,
+        'features': None,
+        'default_label_types': ['pref', 'alt'],
+        'remove_specifiers': False
```
osma, line 34: I think `remove_qualifiers` would be a better name for the option, or perhaps `remove_parentheses`.

annif/backend/yake.py
```diff
@@ -0,0 +1,196 @@
...
+        for word in phrase.split():
+            normalized.append(
+                self.project.analyzer.normalize_word(word).lower())
+        return ' '.join(normalized)
+
+    def _sort_phrase(self, phrase):
+        words = phrase.split()
+        return ' '.join(sorted(words))
+
+    def _suggest(self, text, params):
```
osma, line 139: The `params` passed here don't seem to be used for anything else than the `limit` value. Doesn't this mean that the `--backend-param` option cannot be used to set the parameters at runtime? Would it be better to pass `params` around explicitly between the methods and not rely on `self.params` which are determined by the configuration file?

annif/backend/yake.py
```diff
@@ -0,0 +1,196 @@
...
+        for kp, score in keyphrases:
+            uris = self._keyphrase2uris(kp)
+            for uri in uris:
+                suggestions.append(
+                    (uri, self._transform_score(score)))
+            if not uris:
+                not_matched.append((kp, self._transform_score(score)))
+        # Remove duplicate uris, conflating the scores
+        suggestions = self._combine_suggestions(suggestions)
+        self.debug('Keyphrases not matched:\n' + '\t'.join(
```
osma, line 168: In a future version, I think these non-matched keyphrases should be propagated back to the user as well, but it could be done in a subsequent PR as it requires a lot more scaffolding.

setup.py
```diff
@@ -44,6 +44,7 @@ def read(fname):
         'vw': ['vowpalwabbit==8.8.1'],
         'nn': ['tensorflow-cpu==2.3.1', 'lmdb==1.0.0'],
         'omikuji': ['omikuji==0.3.*'],
+        'yake': ['yake @ git+https://github.com/LIAAD/yake@v0.4.3'],
```
osma, line 47: Yake has just (again) been released on PyPI so this could be changed to `yake==0.4.3` or similar. See https://github.com/LIAAD/yake/issues/42

annif/backend/yake.py
```diff
@@ -0,0 +1,196 @@
...
+        return index
+
+    def _create_index(self):
+        index = defaultdict(set)
+        for concept in self.graph.subjects(RDF.type, SKOS.Concept):
+            if (concept, OWL.deprecated, rdflib.Literal(True)) in self.graph:
+                continue
+            uri = str(concept)
+            for label_type in self.label_types:
+                for label in self.graph.objects(concept, label_type):
```
osma, line 113: Using `graph.preferredLabel` could perhaps be a bit more elegant here, it can be given the list of properties (label types) and the language. See e.g. [this example](https://github.com/NatLibFi/Annif/pull/462/files#diff-231863667a713a02bb9fc65f1bda5432e5a897cc1c97260ed381443be039a82cR181) in the MLLM backend

osma (commented):
I gave some comments on some possible improvements. Overall I think this is looking very promising.

How to express the licensing information (GPLv3) needs some more thought. It shouldn't be terribly complicated but needs a different frame of thinking so I mostly just looked at the code now :)

annif/backend/yake.py
juhoinkinen: I made a helper method `_get_concept_labels` not using the `graph.preferredLabel` because that always returns only one label (the first match that is encountered when comparing against the given label types in `labelProperties`). And I think here it's better to have all the requested label types, so it seems that looping over label types would be needed even with `graph.preferredLabel`.

annif/backend/yake.py
juhoinkinen: The Yake object is now created inside `suggest` and it is instantiated with the explicitly passed `params`; this allows setting those parameters on runtime. However the index is still created or loaded in `initialize`, which doesn't have access to the passed `params`, so switching to a different labels set is not possible on runtime. That would be quite convenient, but I think it would necessitate having separate indices/index files for pref, alt and hiddenLabels.

annif/backend/yake.py
osma: Ah, you're right, `graph.preferredLabel` wouldn't work in this case.

juhoinkinen:
Rebased & force-pushed

annif/backend/yake.py
```diff
@@ -0,0 +1,191 @@
...
+            path = os.path.join(self.datadir, self.INDEX_FILE)
+            if os.path.exists(path):
+                self._index = self._load_index(path)
+                self.info(
+                    f'Loaded index from {path} with {len(self._index)} labels')
+            else:
+                self.info('Creating index')
+                self._index = self._create_index()
+                self._save_index(path)
+                self.info(f'Created index with {len(self._index)} labels')
```
juhoinkinen, line 77: Now the index is created on a suggest call only when the index does not yet exist (it takes about 1 min to create it, a bit too long to create it on every call?). With existing index, if a project's vocabulary is updated, the update does not change the Yake backend index. The user should reset the project e.g. by  `annif clear <project>` to get the vocab update to come into effect, which is a different to the way that trainable backends need (i.e. retraining).

Here could be a condition to check if the vocabulary is newer than the index file, and  recreate the index if so. But perhaps it is better to implement this with/after issue #485, as probably some common functionalities can be used.

juhoinkinen: Related to recreating the index: if a user changes the project configuration related to index (labeltypes, `remove_parentheses` boolean) the change does not have the expected effect if the user does not also clear the project.

annif/vocab.py
```diff
@@ -55,6 +55,18 @@ def subjects(self):
...
         return self._subjects
 
+    @property
+    def skos(self):
+        """return the subject vocabulary from SKOS file"""
+        if self._skos_vocab is None:
+            path = os.path.join(self.datadir, 'subjects.ttl')
+            if os.path.exists(path):
+                logger.debug(f'loading graph from {path}')
+                self._skos_vocab = annif.corpus.SubjectFileSKOS(path, None)
```
juhoinkinen, line 65: Here the language for the SKOS vocabulary is set to `None` as the language is not available from within the `AnnifVocabulary` and it's currently not needed. But should the language be made available by passing it from `AnnifProject` when instantiating `AnnifVocabulary`?

juhoinkinen:
I updated the PR description and put evaluation results in the [results table](https://docs.google.com/spreadsheets/d/1kssU1w9xXbBcp_Wco_El4G827-OergyuWH5pF1vDciQ/edit#gid=688744452) for comparison with Maui and MLLM. 

Still to do: a Wiki page for the backend. 

There are some inline comments/questions in the PR from me.  @osma can you take a look at this again?

juhoinkinen:
A draft Wiki page: https://github.com/NatLibFi/Annif/wiki/Backend:-YAKE

Another page edit to do in Wiki: add installation instructions to [Optional features and dependencies](https://github.com/NatLibFi/Annif/wiki/Optional-features-and-dependencies).

README.md
```diff
@@ -133,4 +133,7 @@ Zenodo DOI:
 
 The code in this repository is licensed under Apache License 2.0, except for the
 dependencies included under `annif/static/css` and `annif/static/js`,
-which have their own licenses. See the file headers for details.
+which have their own licenses, see the file headers for details.
+Enabling the optional Yake backend may result in [GPLv3](https://www.gnu.org/licenses/gpl-3.0.txt)
```
osma, line 137: How about this instead:

Please note that the Yake library is licended under [GPLv3](https://www.gnu.org/licenses/gpl-3.0.txt), while Annif is licensed under the Apache License 2.0. The licenses are compatible, but depending on legal interpretation, the terms of the GPL (for example the requirement to publish corresponding source code when publishing an executable application) may be considered to apply to the whole of Annif+Yake if you decide to install the optional Yake dependency.

annif/backend/yake.py
osma: I think the current code is good; this just needs to be documented on the wiki page of the backend.

annif/backend/yake.py
```diff
@@ -0,0 +1,191 @@
...
+                self._index = self._load_index(path)
+                self.debug(
+                    f'Loaded index from {path} with {len(self._index)} labels')
+            else:
+                self.info('Creating index')
+                self._index = self._create_index()
+                self._save_index(path)
+                self.info(f'Created index with {len(self._index)} labels')
+
+    def _save_index(self, path):
```
osma, line 79: Perhaps `annif.util.atomic_save` could be used here, to ensure that the index won't be corrupted if the process terminates in the middle of writing it?

annif/backend/yake.py
```diff
@@ -0,0 +1,191 @@
...
+        return dict(index)
+
+    def _normalize_label(self, label):
+        label = str(label)
+        if annif.util.boolean(self.params['remove_parentheses']):
+            label = re.sub(r' \(.*\)', '', label)
+        lemmatized_label = self._lemmatize_phrase(label)
+        return self._sort_phrase(lemmatized_label)
+
+    def _lemmatize_phrase(self, phrase):
```
osma, line 113: I think this could be renamed to `_normalize_phrase` because not all analyzers perform lemmatization (only Voikko does currently). Snowball does stemming but that's a bit different...

annif/backend/yake.py
osma: OK - this could perhaps be documented on the wiki page

annif/vocab.py
osma: I guess that would be better...I'm a bit nervous about None values

osma (changes requested):
Looks good. I gave a few suggestions for small changes. You can decide if you want to address them or not, then it's OK to merge this.

I will write a separate comment about the wiki documentation

osma:
Regarding the wiki page:

* there were a couple of comments above about how to handle vocabulary changes etc. (the need for `annif clear`)
* the parameters are not very well explained currently; the descriptions could be expanded a bit (currently only default values are mentiones) and perhaps there could be a paragraph or two explaining how to configure this in common situations
* the [Optional features and dependencies](https://github.com/NatLibFi/Annif/wiki/Optional-features-and-dependencies) page needs a section on Yake - this could include the same note about GPL as in the README

PR #462, opened 2021-01-15
New lexical backend MLLM

osma:
This (draft) PR adds a new Annif backend called MLLM, which stands for "Maui-like Lexical Matching". This is a Python reimplementation of the Maui algorithm (the parts related to matching using a controlled vocabulary), but with some important differences. For example, the initial matching between subject labels and input text is done using sets of tokens, one sentence at a time, instead of n-grams as Maui does. The intent is that this backend would eventually replace the use of Maui, since it's a lot easier to set up and use a built-in algoritm instead of an external Maui Server instance.

The code is in an early draft state and a number of things need to be improved before this can be merged, but I'm opening the PR now to get feedback from automated tools. Here are some TODO items:

- [x] support `--cached` option for training with previously preprocessed training data
- [x] refactor by moving some classes to a separate module (e.g. `annif.lexical`)
- [x] add support for semantic features based on vocabulary structure (broader, narrower, related)
- [x] add support for collection membership feature
  - [x] verify that this works also without SKOS Collections in the vocabulary 
- [x] support hyperparameter optimization using the `annif hyperopt` command
- [x] complete unit tests
- [x] fix QA tool warnings
- [x] wiki documentation


osma:
Rebased and force-pushed.

osma:
I've now tackled some QA tool issues, though complaints still remain. Anyway I think this is ready for wider testing now.

osma:
I wrote some wiki documentation: https://github.com/NatLibFi/Annif/wiki/Backend:-MLLM
This page has not yet been linked to from the Home page or the sidebar. The links should be added when a version of Annif with MLLM is released.

osma:
There are still some QA tool warnings, but it's not easy to fix them. I would like to merge this pretty soon, unless there are objections.

annif/lexical/mllm.py
```diff
@@ -0,0 +1,215 @@
...
+            doc_count += 1
+
+        # precalculate idf values for all candidate subjects
+        self._idf = self._calculate_idf(subject_ids, doc_count)
+
+        return (np.vstack([self._candidates_to_features(candidates)
+                           for candidates in train_x]), np.array(train_y))
+
+    def _create_classifier(self, params):
+        return BaggingClassifier(
```
juhoinkinen, line 187: `BaggingClassifier` has `n_jobs` parameter for parallelization but I don't know could it speed much the overall fitting (or predictions), depending on if candidate generation is much slower.

juhoinkinen (approved):
I looked through the code, I could comment only about `n_jobs` parameter, and probably that would not be very useful (at least before candidate generation is parallelized, which is stated as TODO). 

Overall it is great to have MLLM in Annif! :)

annif/lexical/mllm.py
osma: You are right. The [default value](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.BaggingClassifier.html) for `n_jobs`in BaggingClassifier is None, which in practice means no parallelism. Using -1 instead would use all CPU cores for both `fit` and `predict` operations. My guess is that the impact would be marginal, but I'm not really sure, so maybe I'll try. My hunch is that candidate generation will dominate, but this may be different for hyperparameter optimization at least, since in that case the candidates are generated just once.

osma: I tested this using the kirjaesittelyt corpus, on a laptop with 4 cores (8 threads). Results were not very encouraging.

# Wall clock time in minutes:seconds before / after setting n_jobs=-1

* train on maui-train set: 2:36 / 2:40
* eval on test set (non-parallel): 2:37 / 3:10
* eval on test set (jobs=4): 1:19 / 1:27 with warning `UserWarning: Loky-backed parallel loops cannot be called in a multiprocessing, setting n_jobs=1`
* hyperopt on validate set (trials=20, jobs=4): 18:33 / 17:52

For train and eval, the processing time actually increased with n_jobs=-1; parallel evaluation caused additional warnings (it's not possible to use multiprocessing on multiple levels). For hyperopt there was a ~4% speedup, but it could have been a measuring error.

Issue #456, opened 2020-12-08
"UK" doesn't seem to be registered as input


mo-fu:
During our 2020 evaluation we discovered that the STW concept for United Kingdom was often not assigned, even though __UK__ was present in the input. Changing it to __U.K.__ helped. This is also true for the web UI on https://ai.finto.fi/ for the __YSO English__. Just try these two short strings. Longer sentences show the same effect.I suppose something happens during preprocessing, that removes the two letter word.

osma:
The minimum token length is [set to 3](https://github.com/NatLibFi/Annif/blob/master/annif/analyzer/analyzer.py#L15) in the Analyzer base class, from which other analyzers such as `simple` and `snowball` inherit the functionality.

Of course it's possible to change this, but it could add a lot of noise and increase the size of models. Any ideas?

mo-fu:
Makes sense. 
My first idea was to preprocess the input to insert dots between letters of allcaps words. Simmillar to how STWFSA detects both variants. This hack asically increasing the size of the token. 
After your response I also thought about allowing only specific two letter tokens but this would probably increase processing time, if possible at all.

osma:
It would also be possible to tweak the [is_valid_token](https://github.com/NatLibFi/Annif/blob/master/annif/analyzer/analyzer.py#L22) method, for example so that all caps words are treated differently (e.g. minimum length 2 for all caps).

Heuristics like this are bound to be imperfect - even very short words may have important meaning. For example in Swedish "ö" means island and "å" means river :)

mo-fu:
I feared that making exceptions would open the flood gates for even more exceptions.

osma:
I think that the best way to approach this is to try to make an adjustment, then benchmark the results before and after the change, for example with an Omikuji model, on a couple of different data sets. In this case important metrics could be model size, training time and RAM, precision, recall, F1 score etc.

mo-fu:
Made some experiments regarding timing and memory used for min token sizes two and three. I did so when no other programs were running and disabled turbo boost to have less impact by thermal throttling.
The project config was
```yaml
[omikuji-yso-en]
name=omikujiy YSO english
language=en
backend=omikuji
vocab=yso
analyzer=snowball(english)
cluster_k=100
max_depth=3 
```
I used the YSO-Finna title data set from the tutorial.

Here are the major stats
Stat | Value for 2 | Value for 3|
----|----|----
NDCG|0.6692629691171502|0.668092981381023
[...]

PR #468, opened 2021-02-18
Set Min Token Size to Two

mo-fu:
The changes in this PR will decrease the min token size for the analyzer to two. As discussed in #456 the impact on run time is not very large.

### DIFF
--- a/annif/analyzer/__init__.py
+++ b/annif/analyzer/__init__.py
@@ -12,6 +12,27 @@ def register_analyzer(analyzer):
     _analyzers[analyzer.name] = analyzer
 
 
+def _extend_posargs(posargs):
+    if not posargs:
+        posargs = [None]
+    return posargs
+
+
+def _parse_analyzer_args(param_string):
+    if not param_string:
+        return [None], {}
+    kwargs = {}
+    posargs = []
+    param_strings = param_string.split(',')
+    for p_string in param_strings:
+        parts = p_string.split('=')
+        if len(parts) == 1:
+            posargs.append(p_string)
+        elif len(parts) == 2:
+            kwargs[parts[0]] = parts[1]
+    return _extend_posargs(posargs), kwargs
+
+
 def get_analyzer(analyzerspec):
     match = re.match(r'(\w+)(\((.*)\))?', analyzerspec)
     if match is None:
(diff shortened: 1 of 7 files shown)


juhoinkinen:
I checked the effect of this PR (changing min token size from 3 to 2), training on Finna data set, evaluating on JYX thesis test set. 

Omikuji Parabel - Finnish:
| Stat | Value for 2 | Value for 3 |
|-|-|-|
| F1@5 | 0.3251 | 0.3386 
| NDCG | 0.4393 | 0.4560 
| NDCG@5 | 0.4250 | 0.4407 
| train time (min) | 174 | 192
| omikuji-train.txt (MB) | 955 | 1012

Omikuji Parabel - English:
| Stat | Value for 2 | Value for 3 |
|-|-|-|
| F1@5 | 0.3157 |  0.3115
| NDCG | 0.4345 |  0.4292
| NDCG@5 | 0.4380 |  0.4319
| train time (min) | 36 | 33
| omikuji-train.txt (MB) | 373 | 422

TFIDF - Finnish (training only on the first dataset file):
| Stat | Value for 2 | Value for 3 |
|-|-|-|
| F1@5 | 0.2277 |  0.2745
| NDCG | 0.3165 | 0.3833 
| NDCG@5 | 0.3169 | 0.3752 
| train time (min) |  11 | 7

The sizes of model files do not really change. I run these on my laptop, so timings are not very accurate. Analyzer was Snowball for all.

So for English slight increase in quality, but unfortunately for Finnish Parabel model I get ~1 %-point decrease and more for TFIDF. I should probably have used Voikko instead of Snowball for Finnish, I'll check if that makes a difference.

osma:
Thanks for the additional testing @juhoinkinen!

I think that getting an improvement for English but worse results for Finnish should be an argument for making this setting configurable instead of using a fixed value of either 2 or 3. I suggested making this value configurable in [this comment](https://github.com/NatLibFi/Annif/issues/456#issuecomment-779039463), perhaps it would be worth revisiting that idea?

juhoinkinen:
Using Voikko instead of Snowball (training on only one file, `yso-finna-fi-01.tsv.gz`):

Omikuji Parabel - Finnish:
| Stat | Value for 2 | Value for 3 |
|-|-|-|
| F1@5 | 0.2950 | 0.2972
| NDCG | 0.3949 |  0.4025
| NDCG@5 | 0.3852 | 0.3929
| train time (min) | 42 | 40
| omikuji-train.txt (MB) | 300 | 283

TFIDF - Finnish:
| Stat | Value for 2 | Value for 3 |
|-|-|-|
| F1@5 | 0.2316 | 0.2828
| NDCG | 0.3255 | 0.3923
| NDCG@5 | 0.3261 | 0.3881
| train time (min) | 8 | 13

juhoinkinen:
I tested this with Voikko & Snowball, no problems :+1: 

Wiki needs to updated for this, at least [the Analyzers page](https://github.com/NatLibFi/Annif/wiki/Analyzers).

juhoinkinen (approved):

juhoinkinen:
Thanks @mo-fu for the complexity fixes, seems ready, so I'll merge this.

PR #471, opened 2021-03-02
Make sure suggestion scores are in the range 0.0-1.0

### LINKED ISSUE #470: NN ensemble may return scores above 1.0
The NN ensemble doesn't always return scores in the range 0.0-1.0.

For example, the `yso-sv` model currently available in Finto AI returns this for the input text `historia`:

```js
{
  "results": [
    {
      "label": "historia",
      "notation": null,
      "score": 1.3140300512313843,
      "uri": "http://www.yso.fi/onto/yso/p1780"
    },
    [...]
}
```

This should be avoided, for example by capping the scores to 1.0.

osma:
Fixes #470

This is a really simple fix to ensure that the NN ensemble can never return a score below 0.0 or above 1.0.

I don't see any obvious downsides except it loses the distinction between, say, 1.0 (very sure this is correct) and 1.1 (extra sure this is correct!). An alternative would be to use some sort of activation function such as [sigmoid](https://en.wikipedia.org/wiki/Sigmoid_function) that squeezes any kind of score values into the range 0.0-1.0 but I think this would be overkill and the resulting typical score range would then probably not make a lot of sense.

Any opinions about this @juhoinkinen ? Would it be fine to merge this as is or should we do something else here?

I couldn't figure out a way to test this since it's pretty rare and it requires a heavily biased model to trigger this.

### DIFF
--- a/annif/suggestion.py
+++ b/annif/suggestion.py
@@ -161,9 +161,7 @@ class ListSuggestionResult(SuggestionResult):
 
     @staticmethod
     def _enforce_score_range(hit):
-        if hit.score < 0.0:
-            return hit._replace(score=0.0)
-        elif hit.score > 1.0:
+        if hit.score > 1.0:
             return hit._replace(score=1.0)
         return hit
 
--- a/annif/suggestion.py
+++ b/annif/suggestion.py
@@ -92,7 +92,9 @@ class VectorSuggestionResult(SuggestionResult):
     """SuggestionResult implementation based primarily on NumPy vectors."""
 
     def __init__(self, vector):
-        self._vector = vector.astype(np.float32)
+        vector_f32 = vector.astype(np.float32)
+        # limit scores to the range 0.0 .. 1.0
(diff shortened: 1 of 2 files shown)


juhoinkinen:
I think it is fine to merge as it is. 

Even if the information that a suggestion is "extra correct" gets lost, those cases seem so rare that the capping should not change the top-10 suggested subjects.

osma:
Thanks @juhoinkinen !

After some thought I will change the implementation slightly - to move the min/max constraints to the `*SuggestionResult` classes. This way the score range will be enforced for all backends, not just nn_ensemble, and it should be easier to write unit tests for this functionality.

osma:
I moved the constraint enforching to the `ListSuggestionResult` and `VectorSuggestionResult` classes and wrote unit tests to check that they are working.

juhoinkinen (approved):
Yes, better to have this on all backends (with Yake there was a problem that sometimes it gave negative scores).

PR #472, opened 2021-03-03
Cleanup suggestion tests

osma:
While working on PR #470, I noticed that some tests in `test_suggestion.py` were inconsistently named and some of them took a `document_corpus` fixture which wasn't actually used. This PR cleans up those.


osma:
Rebased on current master. I had originally based this on the wrong commit

PR #475, opened 2021-03-10
Allow configuration of nn ensemble learning rate

mo-fu:
When training nn-ensemble with optimizers that do not automatically adjust the learning rate it is helpful to configure it manually. Even with automatically adjusting optimizers, it may sometimes be beneficial to change the learning rate.
This pr allow to set the learning rate in the nn-ensemble project definition.
As different optimizers have different default learning rates, it is not possible to just add a default parameter. Instead the presence of a `lr` key in the config map is checked.


osma:
Thanks @mo-fu !

PR #479, opened 2021-03-26
Update stwfsa to 0.2.*

### LINKED ISSUE #478: STWFSA backend returns duplicate results
I set up a stwfsa based project using this configuration adapted from the wiki page of the STWFSA backend:

```
[stw-stwfsa-en]
name=STWFSA STW english
language=en
backend=stwfsa
vocab=stw-en
concept_type_uri=http://zbw.eu/namespaces/zbw-extensions/Descriptor
sub_thesaurus_type_uri=http://zbw.eu/namespaces/zbw-extensions/Thsys
thesaurus_relation_type_uri=http://www.w3.org/2004/02/skos/core#broader
thesaurus_relation_is_specialisation=False
```

Then I trained this with stw-econbiz-small.tsv.gz from the Annif-tutorial stw-zbw data set.

When I use `annif suggest` with text that contains multiple matches for the same term, I get duplicate results:

```
$ echo "economics. economics" | annif suggest stw-stwfsa-en
<http://zbw.eu/stw/descriptor/10032-2>	Economics	0.14664009111617313
<http://zbw.eu/stw/descriptor/10032-2>	Economics	0.14664009111617313
```

This is obviously wrong, since the list of suggestions should not contain duplicates.

Looking at [this PR](https://github.com/zbw/stwfsapy/pull/32) that was quite recently merged to stwfsapy, I think this may already have been fixed upstream. Annif would just need to switch to a newer version of stwfsapy (and such a version must be released first). Ping @mo-fu - do you agree with the analysis? Could this be fixed on the Annif side as well by upgrading stwfsapy?

mo-fu:
A PR for an updated version of stwfsa. The Improvements are:
* No more duplicate concepts in suggestions Fixes #478 
* Better support for long texts by using frequency and positional distribution features of concepts.
* Text Vectorizer is disabled by default to reduce memory requirements for large corpora.
* Improved Escaping of symbols in automaton construction

### DIFF
--- a/annif/backend/stwfsa.py
+++ b/annif/backend/stwfsa.py
@@ -20,6 +20,7 @@ _KEY_EXPAND_ABBREVIATION_WITH_PUNCTUATION = (
     'expand_abbreviation_with_punctuation')
 _KEY_SIMPLE_ENGLISH_PLURAL_RULES = 'simple_english_plural_rules'
 _KEY_INPUT_LIMIT = 'input_limit'
+_KEY_USE_TXT_VEC = 'use_txt_vec'
 
 
 class StwfsaBackend(backend.AnnifBackend):
@@ -40,6 +41,7 @@ class StwfsaBackend(backend.AnnifBackend):
         _KEY_EXPAND_ABBREVIATION_WITH_PUNCTUATION: boolean,
         _KEY_SIMPLE_ENGLISH_PLURAL_RULES: boolean,
         _KEY_INPUT_LIMIT: int,
+        _KEY_USE_TXT_VEC: bool,
     }
 
     DEFAULT_PARAMETERS = {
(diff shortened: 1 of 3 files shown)


osma:
Thanks @mo-fu ! I tested this briefly and it seems to work as advertised.

Am I right that old models (trained with stwfsapy 0.1.x) need to be retrained? At least I got a `KeyError: 'use_txt_vec'` error when I tried to use an old model. No big issue, but should be mentioned in the Annif release notes.

mo-fu:
Yes, that is correct. I have plans to add the version to the serialized files in order to have a better error message.

osma:
Thanks for clarifying @mo-fu ! I will merge this now so it gets into the soon-to-be-released Annif 0.52.

PR #481, opened 2021-04-12
Backwards compatibility fix for the token_min_length setting

osma:
After PR #468 that introduced the `token_min_length` setting, old Omikuji models no longer work. They fail with this error:

```
  File "/home/ozone/git/Annif/annif/analyzer/analyzer.py", line 28, in is_valid_token
    if len(word) < self.token_min_length:
AttributeError: 'SnowballAnalyzer' object has no attribute 'token_min_length'
```

The problem is that the new code expects that the `token_min_length` field is always present. It was set by the new constructor, but old Analyzer objects (serialized alongside the model) don't have it.

This PR changes the implementation so that the field with a default value is defined for the class, but can be overridden in instances. This way old models will keep working, they will just rely on the default value.

Ping @mo-fu

### DIFF
--- a/annif/analyzer/analyzer.py
+++ b/annif/analyzer/analyzer.py
@@ -14,9 +14,11 @@ class Analyzer(metaclass=abc.ABCMeta):
     be overridden when necessary."""
 
     name = None
+    token_min_length = 3  # default value, can be overridden in instances
 
     def __init__(self, **kwargs):
-        self.token_min_length = int(kwargs.get(_KEY_TOKEN_MIN_LENGTH, 3))
+        if _KEY_TOKEN_MIN_LENGTH in kwargs:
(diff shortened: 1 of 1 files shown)


mo-fu:
Thank you for the info and the fix. I wasn't aware of this mechanism in pickle serialization.

osma:
It's not really pickle itself that's the key here, it's that the new code needs to deal with old serialized objects that were created with a different codebase. In this case there was a new constructor introduced in PR #468 that initialized the field, but old objects were created without that constructor so didn't have the field. Luckily (?), Python classes may have fields that are set either on the class level and the instance level, with the instance fields overriding (shadowing) the class level fields of the same name - this works nicely as a mechanism for default values (class level) and optional instance specific values.

This is something I learned many years ago when I did web application development using Zope :) You could get a serious headache when dealing with objects that were serialized several years ago, when the code that was used to create them was entirely different.

I will merge this now.

PR #491, opened 2021-05-27
Raise NotSupportedException for attempt to train YAKE

juhoinkinen:
I noticed that a user could try to train a Yake project (which is unnecessary Yake being unsupervised algorithm), but training would seem to succeed, although actually nothing happens.

A more correct behavior is to raise `NotSupportedException` when training a Yake project is attempted, as in the case of the Ensemble backend.


osma (approved):
LGTM!

Issue #490, opened 2021-05-24
annif --version should show Annif version


osma:
Running `annif --version` currently gives this output:

```
$ annif --version
Python 3.8.5
Flask 1.1.2
Werkzeug 1.0.1
```

It would be nice if this command also showed the version of Annif itself :)

PR #494, opened 2021-05-31
Add --version option to CLI

juhoinkinen:
Currently the version of Annif needs to be obtained by `pip list` or similar not very convenient way, and `annif --version` confusingly outputs the versions of Python, Flask and Werkzeug.

This PR makes the `--version` option to show only the version of Annif itself, utilizing the [Click decorator](https://click.palletsprojects.com/en/7.x/api/?highlight=version_option#click.version_option):
```
$ annif --version
0.53.0.dev0
```

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -23,7 +23,8 @@ from annif.exception import ConfigurationException, NotSupportedException
 logger = annif.logger
 click_log.basic_config(logger)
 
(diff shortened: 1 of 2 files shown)


osma (approved):
Very neat! LGTM

osma:
Should mention that this closes #490 (if you do it in the OP it should work automagically)

PR #495, opened 2021-06-07
Show error when training MLLM on empty corpus

### LINKED ISSUE #484: MLLM train on empty corpus fails with error
When trying to train MLLM on an empty corpus it fails with this:

```
  File "/data/Annif/annif/lexical/mllm.py", line 184, in prepare_train
    for candidates in train_x]), np.array(train_y))
  File "<__array_function__ internals>", line 6, in vstack
  File "/data/Annif/venv/lib/python3.6/site-packages/numpy/core/shape_base.py", line 283, in vstack
    return _nx.concatenate(arrs, 0)
  File "<__array_function__ internals>", line 6, in concatenate
ValueError: need at least one array to concatenate
```

It would be better to show an informative error, similar to the "cannot evaluate empty corpus" message.

osma:
Fixes #484

Creating a draft PR to get feedback from QA tools.

### DIFF
--- a/annif/backend/mllm.py
+++ b/annif/backend/mllm.py
@@ -5,6 +5,7 @@ import joblib
 import numpy as np
 import annif.util
 from annif.exception import NotInitializedException
+from annif.exception import NotSupportedException
 from annif.lexical.mllm import MLLMModel
 from annif.suggestion import VectorSuggestionResult
 from . import backend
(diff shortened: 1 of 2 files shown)


osma:
QA tools are happy except Codecov, which claims that one line in the stwfsa backend is no longer covered. That's a bit puzzling but clearly not related to the changes in this PR.

Issue #464, opened 2021-01-25
Language filtering in input text


juhoinkinen:
A model's attempts to process (in training/suggesting) text in a language different to the project's language are disadvantageous. Generally, any text-parts in a "foreign" language in a document probably just act as noise, but when the input-limit parameter (#446 and #452) is used, the drawback can be more serious: some thesis have abstracts in two or even three languages, and when the text is limited to a fixed number of characters, the retained text can mostly be in "foreign" languages. 

The foreign-language parts could be filtered out from the text that is passed to a backend (before input-limiting). 

One possible library to perform the actual language detection is [pycld3](https://pypi.org/project/pycld3/). It is much faster than [langdetect](https://pypi.org/project/langdetect/), does not require dependencies to be manually installed (installation is using Wheel for Python versions  3.5, 3.6, 3.7, and 3.8), and has Apache 2.0 license. The detected language is given as BCP-47-style language code which is used in Annif too.

The filtering could be optional with a common backend-parameter boolean switch.

juhoinkinen:
Quick runs with a Bonsai model showed that in Finnish test sets line-by-line language filtering (in evaluation only) improved F1@5 scores for [JYX thesis ~1 % point, but worsened for kirjaesittelyt ~2 % points](https://docs.google.com/spreadsheets/d/1sdyXSn_NG2RXfpUHxA4bdNp3ziV9xRb4oDQYlAbUTo0/edit#gid=1956626958&range=A18:O25). Worsening could result from too crude line-by-line filtering (sentence-by-sentence would probably be better), or from the little possibility that Finnish model can learn something from foreign texts (although with lemmatization this seems quite limited).

osma:
> The foreign-language parts could be filtered out from the text that is passed to a backend (before input-limiting).

Sounds good!

> One possible library to perform the actual language detection is pycld3.

+1

> The filtering could be optional with a common backend-parameter boolean switch.

+1

> sentence-by-sentence would probably be better

Agreed - the Analyzer class already has a `tokenize_sentences` method (based on NLTK method `sent_tokenize`) so I think it should be quite easy to rely on that instead of filtering line by line.

PR #496, opened 2021-06-07
Support for adding input-transformation operations

juhoinkinen:
This PR introduces support for adding new, more general text-input transformation capabilities to Annif codebase. Currently transformations include only input-limiting (which was originally implemented as a backend mechanism in #446/#452). 

The transformation specifications are given in the project configuration. For example, `transform=pass,limit(3)` defines a pass-through/identity transformation followed by input-limiting to first three characters. 

In the specification the parts in parentheses are passed as arguments to instantiation of the classes, which internally take care of the actual transformation. Keyword arguments can also be used.

The language-filtering transformation (#464) is left to be implemented in another PR.

At least TODO:
- [x] increase test coverage
- [x] try making specs parsing functions common to analyzer and transformer
- [x] docstrings
- [x] Wiki & `projects.cfg.dist` update


annif/corpus/document.py
```diff
@@ -90,20 +90,21 @@ def documents(self):
...
 
-class TruncatingDocumentCorpus(DocumentCorpus):
-    """A document corpus that wraps another document corpus but truncates the
-    documents to a given length"""
+class TransformingDocumentCorpus(DocumentCorpus):
+    """A document corpus that wraps another document corpus but transforms the
+    documents using a given transform function"""
 
-    def __init__(self, corpus, limit):
+    def __init__(self, corpus, transform_text):
```
osma, line 97: How about calling the parameter `transform_function` or `transform_fn` or something along those lines? It's not obvious that `transform_text` should be a function.

I suggest adding a docstring to the constructor as well, clarifying the meaning of the parameters.

annif/corpus/document.py
```diff
@@ -90,20 +90,21 @@ def documents(self):
...
         self._orig_corpus = corpus
-        self._limit = limit
+        self._transform_text = transform_text
```
osma, line 99: Rename here as well...

annif/project.py
```diff
@@ -90,6 +102,7 @@ def initialize(self):
         logger.debug("Initializing project '%s'", self.project_id)
 
         self._initialize_analyzer()
+        self._initialize_transformer()  # TODO: Is this needed?
```
osma, line 105: I'm trying to decide whether to recommend keeping the initialize function or removing it. It would perhaps be helpful for catching configuration errors early. OTOH they will be encountered anyway in most operations (though probably not in commands like `list-projects`).

Perhaps it would be wise to apply the YAGNI principle. The current transformers (limit and pass) don't need any special initialization, so this can be dropped for now. It can always be added back later if some other transformer (for example machine translation) needs initializing.

annif/project.py
```diff
@@ -117,6 +130,16 @@ def analyzer(self):
...
         return self._analyzer
 
+    @property
+    def transformer(self):
+        if self._transformer is None:
+            if self.transformer_spec:
+                self._transformer = annif.transformer.get_transformer(
+                    self.transformer_spec, project=self)
+            else:
+                self._transformer = None
```
osma, line 140: Assigning None here is basically redundant (`self._transformer` was already None) and it means that the decision not to use an analyzer is never cached. Maybe that's no big deal since the cost of redoing is just an additional if check for the spec. Anyway I suggest dropping the `else` block as it's not doing anything useful.

osma: Another option would be to assign a passthrough transformer here in the `else` block when no spec is set. That could actually be wise, because then `self.transformer` could never return None.

annif/project.py
```diff
@@ -178,6 +201,8 @@ def suggest(self, text, backend_params=None):
                 raise NotInitializedException('Project is not trained.')
         logger.debug('Suggesting subjects for text "%s..." (len=%d)',
                      text[:20], len(text))
+        if self.transformer is not None:
```
osma, line 204: If the property function above falls back to a passthrough transformer (as suggested above), then there is no need for the None check here.

annif/project.py
```diff
@@ -188,6 +213,8 @@ def train(self, corpus, backend_params=None):
             corpus.set_subject_index(self.subjects)
         if backend_params is None:
             backend_params = {}
+        if self.transformer is not None:
```
osma, line 216: Likewise here: no need for a None check if a passthrough transformer fallback is used in the property function

annif/project.py
```diff
@@ -197,6 +224,8 @@ def learn(self, corpus, backend_params=None):
         if backend_params is None:
             backend_params = {}
         beparams = backend_params.get(self.backend.backend_id, {})
+        if self.transformer is not None:
```
osma, line 227: Ditto

annif/transformer/__init__.py
```diff
@@ -0,0 +1,53 @@
+# TODO Add docstring
```
osma, line 1: Yes please

annif/transformer/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+            kwargs[parts[0]] = parts[1]
+    return posargs, kwargs
+
+
+def parse_specs(transformers_spec):
+    """parse a configuration definition such as 'A(x),B(y=1),C' into a tuples
+    of ((A, [x], {}), (B, [None], {y: 1}))..."""  # TODO
+    parsed = []
+    # Split by commas not inside parentheses
+    parts = re.split(r',\s*(?![^()]*\))', transformers_spec)
```
osma, line 28: This regex is a bit hairy...it would be good to have some test cases to verify that it works. Though I tested it on regex101.com and couldn't make it not work :)

annif/transformer/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        if trans not in _transformers:
+            raise ConfigurationException(f"No such transformer {trans}")
+        transformer_classes.append(_transformers[trans])
+        args.append((posargs, kwargs))
+    return transformer.Transformer(transformer_classes, args, project)
+
+
+_transformers = {
+    'pass': transformer.IdentityTransformer,
+    'limit_input': inputlimiter.InputLimiter}
```
osma, line 53: I wonder if just `limit` would be enough as an ID for the InputLimiter transformer

The list of known transformers is now hardcoded as a dict. I think that's OK for now, but in the future we could have a registration mechanism similar to the one used for backends, so that new transformers can be more easily plugged in (perhaps even as optional Python packages / setup.py plugins).

annif/transformer/inputlimiter.py
```diff
@@ -0,0 +1,22 @@
+# TODO Add docstring
```
osma, line 1: Yes please

annif/transformer/inputlimiter.py
```diff
@@ -0,0 +1,22 @@
...
+from annif.exception import ConfigurationException
+from . import transformer
+
+
+class InputLimiter(transformer.IdentityTransformer):
+
+    name = 'limit_input'
```
osma, line 8: The name is now hardcoded in two places: here an in `__init__.py`. It would be better to set it just once here and construct the dict in `init.py` using this information.

annif/transformer/inputlimiter.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __init__(self, project, input_limit):
+        self.project = project
+        self.input_limit = int(input_limit)
```
osma, line 12: Not that Annif generally does much validation of configuration values, but...this could be improved by passing the original `input_limit` string to `_validate_value` which would convert it to int (throwing a ConfigurationException if that fails, for example if the value wasn't a number) and then also testing the value.

Also what happens if the InputLimiter transform is used without a parameter, or with more than one? Maybe that's covered by the "Invalid arguments to..." exception below?

annif/transformer/transformer.py
```diff
@@ -0,0 +1,52 @@
+"""Common functionality for input transforming."""
+
+from annif.corpus import TransformingDocumentCorpus
+from annif.exception import ConfigurationException
+
+
+class IdentityTransformer():
+    """"""""  # TODO
```
osma, line 8: yes

annif/transformer/transformer.py
```diff
@@ -0,0 +1,52 @@
...
+
+    name = 'pass_through'
```
osma, line 10: Here it's `pass_through`, but in the `__init__.py` dict it was just `pass`, this should be consistent I think?

annif/transformer/transformer.py
```diff
@@ -0,0 +1,52 @@
...
+    def transform_text(self, text):
+        """"""  # TODO
+        return text
+
+    def transform_corpus(self, corpus):
+        """"""  # TODO
+        return TransformingDocumentCorpus(corpus, self.transform_text)
+
+
+class Transformer():
```
osma, line 24: I think it's a little confusing that this `Transformer` class is basically unrelated to the actual transformer classes - it's a container for them but not itself a transformer, at least not on the same level. Based on naming I would expect this to be a superclass of all transformers. I suggest renaming it to `TransformerChain` or something along those lines.

annif/transformer/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        if match is None:
+            continue
+        transformer = match.group(1)
+        posargs, kwargs = _parse_transformer_args(match.group(3))
+        parsed.append((transformer, posargs, kwargs))
+    return parsed
+
+
+def get_transformer(transformer_specs, project):
+    transformer_defs = parse_specs(transformer_specs)
```
osma, line 40: The variable naming is a bit inconsistent - here it's `transformer_specs`, but inside the `parse_specs` function it's called `transformers_spec`. Choose one and use it for both :)

annif/transformer/transformer.py
```diff
@@ -0,0 +1,52 @@
...
+    """"""  # TODO
```
osma, line 25: Yes

osma (changes requested):
It's looking good as a first draft and I like how this PR removes complexity from existing backends. I have the feeling that there would be room to simplify this further (there are quite many new classes) but I can't give a concrete suggestion how to do that :) Anyway I gave some suggestions for further improvements.

annif/transformer/inputlimiter.py
juhoinkinen: My idea was that the common argument validation by ConfigurationException with "Invalid arguments to..."  would handle both type validation of arguments (ValueErrors) and also missing (or too many) arguments (TypeErrors)  to transforms. 

For example for missing character number to limit transform:
`Error: Misconfigured project 'yso-yake-fi': Invalid arguments to limit transform: [], {})`

Or for wrong type for character number:
`Error: Misconfigured project 'yso-yake-fi': Invalid arguments to limit transform: ['not-a-number'], {})`

 Each transformer could then also do more specific validation for e.g. value ranges as in here.

juhoinkinen:
I addressed the comments, @osma can you take a look again (or continue with this)?

annif/project.py
```diff
@@ -27,6 +28,7 @@ class AnnifProject(DatadirMixin):
     """Class representing the configuration of a single Annif project."""
 
     # defaults for uninitialized instances
+    _transformer = None
```
osma, line 31: I suggest renaming this field simply `_transform`, and all other entities (packages, classes, variables, functions...) with names like "transformer" similarly to just "transform", because Transformers are an established ML method and it might be confusing to have similarly named classes and variables that are not related to Transformers. In the future, Annif may even get one or more backends based on transformers (for example via pecos)

osma (changes requested):
Looking good. I gave a few more suggestions for improvement for you to consider. I think this can then be merged without another round of review.

PR #498, opened 2021-06-15
Update Dockerfile

juhoinkinen:
This
- updates Docker image to use Python 3.8
- changes dependencies to be installed from `setup.py` (`pip install .[voikko,...]`) instead of installing with separate pip call (thus removing need to have versions pinned in both `setup.py` and `Dockerfile`)
- removes unnecessary installation and linking steps in build-stage (nowdays Vowpal Wabbit has wheel in PyPI)
- reorders layers a bit (Annif source is copied after dependencies are installed, so build should be faster)
- installs Yake in Docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,79 +1,42 @@
-FROM python:3.7-slim-buster AS builder
+FROM python:3.8-slim-buster AS builder
 
 LABEL maintainer="Juho Inkinen <juho.inkinen@helsinki.fi>"
 
-## Install optional dependencies:
+# Install fastText, which needs to be build, and therefore also some system packages:
 RUN apt-get update \
-	## Voikko:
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,79 +1,42 @@
-FROM python:3.7-slim-buster AS builder
+FROM python:3.8-slim-buster AS builder
 
 LABEL maintainer="Juho Inkinen <juho.inkinen@helsinki.fi>"
 
-## Install optional dependencies:
+# Install fastText, which needs to be build, and therefore also some system packages:
```
osma, line 5: Typo: build -> built

osma (approved):
Whoa! All these improvements, and the Dockerfile is now almost 40 lines shorter :) Great work!

Issue #453, opened 2020-11-24
Parallelized eval of nn_ensemble projects hangs


juhoinkinen:
When running evaluations for new models with four parallel jobs I noticed that evaluation of nn_ensemble projects never finished. Without multiprocessing there was no problem. 

Minimal example:
```
# projects.cfg:
[arch-fasttext]
name=FastText
language=fi
backend=fasttext
limit=100
vocab=arch
analyzer=snowball(finnish)

[arch-nn]
name=YSO NN ensemble Finnish
language=fi
backend=nn_ensemble
sources=arch-fasttext
limit=100
vocab=arch
nodes=100
dropout_rate=0.2
epochs=2

# Testing:
annif loadvoc arch-fasttext tests/corpora/archaeology/subjects.tsv
annif train arch-fasttext tests/corpora/archaeology/documents.tsv
annif train arch-nn tests/corpora/archaeology/fulltext/
time annif eval arch-nn tests/corpora/archaeology/documents.tsv --jobs 1  # finishes in 3 min
time annif eval arch-nn tests/corpora/archaeology/documents.tsv --jobs 2  # won't finish
```

This seems to be a common problem, there's many search results for "python multiprocess hangs when calling a tensorflow model". The problem occurs at least with TF versions 2.2.0 and 2.3.1. A quick fix could be to throw a `NotSupportedException` instead of indefinite freeze. Parallelized eval was not promised to work with nn_ensemble when the feature was implemented in #418.

osma:
Thanks for the report and the minimal test case, this is very useful analysis.
I labelled this as a bug because the command just silently hangs forever.

osma:
I took a closer look - the problem is the backend initialization on this line: https://github.com/NatLibFi/Annif/blob/master/annif/cli.py#L365

This is done just before the workers are started in forked processes. This way the initialization (e.g. loading models from disk into memory) has to be done just once and the workers can then make use of them without having to load them anymore. But TensorFlow doesn't like this - maybe it uses shared memory that doesn't behave well after forking.

Commenting out this line makes the NN ensemble work in parallel - since the initialization of projects (not just the ensemble, but also the projects it uses as sources) is delayed and happens within each worker process. The downside is that this takes longer and uses more memory.

I'm now trying to think of an elegant way of delaying the initialization of the TensorFlow/Keras model while still initializing e.g. the source projects.

PR #500, opened 2021-06-16
Implement custom MeanLayer in nn_ensemble

osma:
This PR replaces the Lambda layer with a custom MeanLayer in the nn_ensemble backend. It was intended as a fix to the CustomMaskWarning error encountered in PR #499 (upgrade to TensorFlow 2.5.0) but it turned out that the problem is unrelated.

This might still be a good idea, for example it could improve the compatibility of saved nn_ensemble models between Python versions. But it needs more testing so I'm leaving this as a draft PR.

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -9,7 +9,7 @@ import numpy as np
 from scipy.sparse import csr_matrix, csc_matrix
 import joblib
 import lmdb
-from tensorflow.keras.layers import Input, Dense, Add, Flatten, Lambda, Dropout
+from tensorflow.keras.layers import Input, Dense, Add, Flatten, Dropout, Layer
 from tensorflow.keras.models import Model, load_model
 from tensorflow.keras.utils import Sequence
 import tensorflow.keras.backend as K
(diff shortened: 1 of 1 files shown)


osma:
I tested this in a situation where Python 3.6 is used for training models and Python 3.8 for using them. The PR seems to improve compatibility between Python versions for NN ensemble projects, but both instances need to run the updated code.

First I shamelessly copied the minimal setup from #453:

```
# projects.cfg:
[arch-fasttext]
name=FastText
language=fi
backend=fasttext
limit=100
vocab=arch
analyzer=snowball(finnish)

[arch-nn]
name=YSO NN ensemble Finnish
language=fi
backend=nn_ensemble
sources=arch-fasttext
limit=100
vocab=arch
nodes=100
dropout_rate=0.2
epochs=2

# Training on Python 3.6:
annif loadvoc arch-fasttext tests/corpora/archaeology/subjects.tsv
annif train arch-fasttext tests/corpora/archaeology/documents.tsv
annif train arch-nn tests/corpora/archaeology/fulltext/

# Testing on Python 3.8:
echo arkeologia | annif suggest arch-nn
```

Without this PR, the `annif suggest` command fails with this error:

    ValueError: bad marshal data (unknown type code)

With this PR applied, it works without problems.

I also tested the case of old models trained without this PR. They still keep working with the PR applied. So this PR doesn't break compatibility for old NN ensemble projects, but projects will have to be retrained to be able to benefit from the additional compatibility this PR offers.

juhoinkinen:
Like above I tested training models on Python 3.7 and using them on 3.9: without this PR the `bad marshal data (unknown type code)` error appears, but with this PR applied it does not (suggest call works).

juhoinkinen (approved):

PR #507, opened 2021-07-16
Issue464 language filtering in input text

### LINKED ISSUE #464: Language filtering in input text

juhoinkinen:
This is a draft PR with an initial implementation for the language-filtering feature, closes #464. 

~The branch of this PR is based on the input-transform branch (not yet in master), so PR #496 needs to be merged first to master.~ (Done.)

Needs finalizing and testing.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -38,8 +38,8 @@ jobs:
         # Install the optional fastText dependencies for Python 3.7 only
         if [[ ${{ matrix.python-version }} == '3.7' ]]; then pip install .[fasttext]; fi
         # For Python 3.6
-        # - voikko dependencies
-        if [[ ${{ matrix.python-version }} == '3.6' ]]; then python -m pip install .[voikko]; fi 
+        # - voikko and pycld3 dependencies
+        if [[ ${{ matrix.python-version }} == '3.6' ]]; then python -m pip install .[voikko,pycld3]; fi
         # Install the optional VowpalWabbit dependency for Python 3.6
         if [[ ${{ matrix.python-version }} == '3.6' ]]; then pip install .[vw]; fi
 
--- a/annif/project.py
+++ b/annif/project.py
@@ -120,8 +120,7 @@ class AnnifProject(DatadirMixin):
                     self.analyzer_spec)
             else:
                 raise ConfigurationException(
-                    "analyzer setting is missing (and needed by the backend)",
-                    project_id=self.project_id)
+                    "analyzer setting is missing", project_id=self.project_id)
         return self._analyzer
 
     @property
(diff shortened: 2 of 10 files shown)


juhoinkinen:
Rebased on master (as the PR #496 of input-transform functionality was merged to it) and force-pushed.

juhoinkinen:
Rebased on master and force-pushed.

juhoinkinen:
I added the `text_min_length` parameter, so language filtering of short texts can be bypassed. I think the user should remain responsible on the correctness of short texts, and this prevents somewhat embarrassing cases where a mistake in language detection would filter out all text input. (E.g. "Porovuosi : tutkija pororenkinä Sompiossa" is detected to be Lithuanian with prob. 0.84.). With this parameter set to 500 chars most of the Finna training documents are bypassing the language filtering, but that data set could and maybe should) be cleaned outside Annif.

 The `sentence_min_length` parameter is used to bypass language filtering of short sentences, for which the lang detection is unreliable.

juhoinkinen:
Rebased & force-pushed to include the recent fix #528.

juhoinkinen:
The [evaluation results](https://docs.google.com/spreadsheets/d/1kssU1w9xXbBcp_Wco_El4G827-OergyuWH5pF1vDciQ/edit#gid=810296471) for the current implementation most notably show improvement in the Bonsai model for JYU theses. I think this comes just from the initial motivation for the language filtering (in theses there are abstracts in "foreign" languages and when truncating the text input to 5000 chars the retained parts end up including a large "foreign" language text, which is unuseful for backends).

Bonsai model for Vapaakappaleet worsen by 1 %-point essentially zeroing the average improvement for Bonsai model. I tried to see why is this by checking what does the filtering remove, but in general the filtering works as expected and removes only non-Finnish sentences (and now when training with Finna data essentially nothing is removed). The only idea I have is that those non-Finnish parts include abbreviations, names etc. and removing those is disadvantageous .

The results for other backends on average show very slight improvement.

How does this look @osma?

osma (changes requested):
Code looks good. The only issue I see is that pycld3 is native code (prebuilt wheels are available for macOS and Linux, but not Windows) and in alpha status. It would be better to make it an optional dependency.

PR #509, opened 2021-08-10
Warn instead of error in case of multiple subjects per doc in SVC training

juhoinkinen:
PR #501 fixed training SVC on full-text corpus, and also added raising `NotSupportedException` if any training document had multiple uris (as SVC is a multi-class, not multi-label algorithm). However in real-life even in those data-sets meant to have only one subject per doc there can be some docs with many subjects. To allow training SVC on such data sets, this PR changes the error to warning, and lets a random one of the uris to be used as the target uri.

### DIFF
--- a/annif/backend/svc.py
+++ b/annif/backend/svc.py
@@ -47,16 +47,15 @@ class SVCBackend(mixins.TfidfVectorizerMixin, backend.AnnifBackend):
         self.initialize_vectorizer()
         self._initialize_model()
 
-    @staticmethod
-    def _corpus_to_texts_and_classes(corpus):
+    def _corpus_to_texts_and_classes(self, corpus):
         texts = []
         classes = []
         for doc in corpus.documents:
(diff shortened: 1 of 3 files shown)


annif/backend/svc.py
```diff
@@ -47,16 +47,15 @@ def initialize(self):
...
         for doc in corpus.documents:
             texts.append(doc.text)
             if len(doc.uris) > 1:
-                raise NotSupportedException(
-                    'SVC backend does not support training on documents ' +
-                    'with multiple subjects.')
+                self.warning(
+                    'training on a document with multiple subjects is not ' +
+                    'supported by SVC; selecting one random subject.')
             classes.append(list(doc.uris)[0])
```
osma, line 59: Another option which avoids creating a list:

    classes.append(iter(doc.uris).next())

osma:
LGTM. Suggested an extremely minor change for you to consider before merging.

PR #510, opened 2021-08-12
Use set as container of uris instead of list in DocumentFile

### LINKED ISSUE #502: Discrepancy in types of document subjects
PR #501 was a quick fix for training SVC backend on fulltext corpus, but it did not address the underlying reason:
> ...`DocumentDirectory` [defines uris for documents as a set](https://github.com/NatLibFi/Annif/blob/2b8e3c95e41305532516a9f51be5274788f5579a/annif/corpus/document.py#L46), while in SVC it was assumed that uris are list or other subscriptable (which for `DocumentFile` [is true](https://github.com/NatLibFi/Annif/blob/2b8e3c95e41305532516a9f51be5274788f5579a/annif/corpus/document.py#L71-L72))

Right now I am not sure if there is a reason for subjects to be a set from from DocumentDirectory but a list from DocumentFile, so I'm just reporting this as a possible bug.

juhoinkinen:
Makes the type of the uris container in DocumentFile and DocumentDirectory the same, i.e. set.

Closes #502.

### DIFF
--- a/annif/corpus/document.py
+++ b/annif/corpus/document.py
@@ -68,8 +68,8 @@ class DocumentFile(DocumentCorpus):
     def _parse_tsv_line(self, line):
         if '\t' in line:
             text, uris = line.split('\t', maxsplit=1)
-            subjects = [annif.util.cleanup_uri(uri)
-                        for uri in uris.split()]
+            subjects = {annif.util.cleanup_uri(uri)
+                        for uri in uris.split()}
             yield self._create_document(text=text,
                                         uris=subjects,
                                         labels=[])
--- a/tests/test_backend_svc.py
+++ b/tests/test_backend_svc.py
@@ -86,18 +86,13 @@ def test_svc_suggest(project):
     svc_type = annif.backend.get_backend('svc')
     svc = svc_type(
         backend_id='svc',
-        config_params={'limit': 8},
+        config_params={'limit': 20},
         project=project)
 
-    results = svc.suggest("""Arkeologiaa sanotaan joskus myös
-        muinaistutkimukseksi tai muinaistieteeksi. Se on humanistinen tiede
-        tai oikeammin joukko tieteitä, jotka tutkivat ihmisen menneisyyttä.
-        Tutkimusta tehdään analysoimalla muinaisjäännöksiä eli niitä jälkiä,
-        joita ihmisten toiminta on jättänyt maaperään tai vesistöjen
-        pohjaan.""")
+    results = svc.suggest("""Arkeologiaa sanotaan joskus myös...""")
 
     assert len(results) > 0
-    assert len(results) <= 8
+    assert len(results) <= 20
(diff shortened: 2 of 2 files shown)


juhoinkinen:
By just changing the list to set in DocumentFile [a test for SVC](https://github.com/NatLibFi/Annif/blob/master/tests/test_backend_svc.py#L85) fails occasionally (about half the time), because then in training the "arkeologit" subject is not necessarily taken as the target subject when there are multiple subjects, but a random one is taken. Before the "arkeologit" subject was always the target subject as it is first of the subjects in the training file, e.g. in [here](https://github.com/NatLibFi/Annif/blob/483f047699b09a2029ca3e27401f069385d3d98e/tests/corpora/archaeology/documents.tsv#L193).

Increasing the number of requested subjects to 50 helps to ensure "arkeologit" is always one of the suggested subjects (even 40 is not enough), but I wonder if there would be some better way in this?

One possibility would be to tweak the SVC training to work with many target subjects instead of taking a [random one](https://github.com/NatLibFi/Annif/blob/master/annif/backend/svc.py#L59) by making as many copies of the text as there are uris with something like:
```python
for uri in doc.uris:
     texts.append(doc.text)
     classes.append(uri)
```

osma:
That's a bit unfortunate...in part it stems from using a test corpus which is not really intended for multiclass classification in the SVC unit tests.

The fix you suggested is possible but does it really help? It would change the way SVC works, perhaps not a lot, but it would affect the results in a paper on Libris DDC classification I'm writing :) Not that it matters much since this change would be in a specific release, and the paper can specify the version of Annif that was used.

Can you make a PR implementing this change (perhaps just adding to this one?), then I could check that it won't adversely affect the results I'm getting with SVC on the Libris-DDC data set?

juhoinkinen:
A simple option would be to change the suggest test to use some other input text that always gives a predictable subject.

osma:
That's a great idea - do you have anything specific in mind? Maybe there is some subject in the training set that appears alone in some document.

It wouldn't hurt to have another test corpus for SVC and similar multiclass algorithms, but that's a bit more work...

juhoinkinen:
I thought it would have been trivial to produce an input text that for which SVC always gives the same subject, but it was not... [A document about zikkuratit](https://github.com/NatLibFi/Annif/blob/master/tests/corpora/archaeology/documents.tsv#L5760) is the only document with the "zikkuratit" subject, but to make the unit test pass reliably the input text had contain only "zikkuratit", adding basically anything else gave many other subject suggestions and sometimes the "zikkurarit" subject was missing. However using a short text including "arkeologia" and getting 20 suggestions seems to work.

(There is a similar indeterminacy issue with stwfsa suggest test that I've encountered ~5 times, but I think it's not related to any recent changes, it's been appearing for some time already.)

osma (approved):
So much effort for such a small change! Looks good to me.

PR #511, opened 2021-08-12
Process training docs in parallel in MLLM backend

### LINKED ISSUE #483: Parallel processing of training docs in MLLM
Training the MLLM backend can be a bit slow. Most of the time is spent generating candidates from the training documents. This could probably be done faster by using parallel processing. It's noted as a TODO item in the code: https://github.com/NatLibFi/Annif/blob/master/annif/backend/mllm.py#L23

osma:
Fixes #483 
Early draft, opening to get feedback from QA tools.
Still needs more testing & refinement

### DIFF
--- a/annif/backend/mllm.py
+++ b/annif/backend/mllm.py
@@ -124,7 +124,8 @@ class MLLMBackend(hyperopt.AnnifHyperoptBackend):
             train_data = self._model.prepare_train(corpus,
                                                    self.project.vocab,
                                                    self.project.analyzer,
-                                                   params)
+                                                   params,
+                                                   jobs)
             annif.util.atomic_save(train_data,
                                    self.datadir,
                                    self.TRAIN_FILE,
--- a/annif/lexical/mllm.py
+++ b/annif/lexical/mllm.py
@@ -1,5 +1,6 @@
 """MLLM (Maui-like Lexical Matchin) model for Annif"""
 
+import time
 import collections
 import math
 import joblib
@@ -39,7 +40,7 @@ def conflate_matches(matches, doc_length):
     subj_matches = collections.defaultdict(list)
     for match in matches:
         subj_matches[match.subject_id].append(match)
-    return [
+    ret = [
         Candidate(
             doc_length=doc_length,
             subject_id=subject_id,
@@ -52,6 +53,7 @@ def conflate_matches(matches, doc_length):
             spread=(matches[-1].pos - matches[0].pos) / doc_length
         )
         for subject_id, matches in subj_matches.items()]
+    return ret
 
 
 def generate_candidates(text, analyzer, vectorizer, index):
@@ -68,7 +70,8 @@ def generate_candidates(text, analyzer, vectorizer, index):
                                  pos=sent_idx,
                                  ambiguity=ambiguity))
 
-    return conflate_matches(matches, len(sentences))
+    ret = conflate_matches(matches, len(sentences))
+    return ret
 
 
 class MLLMCandidateGenerator:
@@ -128,6 +131,8 @@ class MLLMModel:
         return matrix
 
     def _prepare_terms(self, graph, vocab, params):
+        print("_prepare_terms starting")
+        starttime = time.time()
         if annif.util.boolean(params['use_hidden_labels']):
             label_props = [SKOS.altLabel, SKOS.hiddenLabel]
         else:
@@ -145,9 +150,12 @@ class MLLMModel:
                                   label=label,
                                   is_pref=False))
 
+        print("_prepare_terms took {:.1f} s".format(time.time()-starttime))
         return (terms, subject_ids)
 
     def _prepare_relations(self, graph, vocab):
+        print("_prepare_relations starting")
+        starttime = time.time()
         self._broader_matrix = make_relation_matrix(
             graph, vocab, SKOS.broader)
         self._narrower_matrix = make_relation_matrix(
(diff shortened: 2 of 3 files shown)


osma:
I did extensive testing on three machines: a HP laptop with 2 CPU cores, an Asus laptop with 4 CPU cores, and a server with 48 CPU cores. The results look pretty promising.

I used two data sets for training:
* kirjastonhoitaja train set (2625 quite short documents) - KK in the table below
* jyu-theses fin-maui-train set (200 very long documents) - JYU in the table below

Configuration:

```
[kk-mllm]
language=fi
backend=mllm
vocab=yso-fi
analyzer=voikko(fi)

[jyu-mllm]
language=fi
backend=mllm
vocab=yso-fi
analyzer=snowball(finnish)
```

# Results

## HP laptop, i5-7200U (2 cores/4 threads)

**-**|**KK train, wall (min:s)**|**KK train, user (s)**|**JYU train, wall (min:s)**|**JYU train, user (s)**
-----|:-----:|:-----:|:-----:|:-----:
Before PR|17:37|1055|5:24|323
jobs=1|18:54|1131|5:51|349
jobs not set|8:35|1678|3:25|480

## Asus laptop, i7-8550U (4 cores/8 threads)

**-**|**KK train, wall (min:s)**|**KK train, user (s)**|**JYU train, wall (min:s)**|**JYU train, user (s)**
-----|:-----:|:-----:|:-----:|:-----:
Before PR|18:52|1130|6:01|355
jobs=1|20:04|1203|6:03|363
jobs not set|9:15|3644|4:19|1002

## Server, EPYC 7401 (48 cores/96 threads)

**-**|**KK train, wall (min:s)**|**KK train, user (s)**|**JYU train, wall (min:s)**|**JYU train, user (s)**
-----|:-----:|:-----:|:-----:|:-----:
Before PR|19:16|1156|6:59|418
jobs=1|20:23|1223|7:07|427
jobs=16|3:34|1252|3:07|465
jobs not set|2:47|1995|2:53|463

# Analysis

Train times (wall clock time) are generally shorter for both data sets and all machines; much shorter for the server that has a lot of CPU cores. But in the jobs=1 case, when only one CPU is used, there is an increase of 5% or so due to overhead. User CPU times increase, so overall the CPUs are doing more work but finishing faster. I think this is a reasonable trade-off.

There is no measurable difference in memory usage.

As a bonus, unit tests for the MLLM backend now run a few seconds faster (30s -> 25s or so).

Still need to clean up the code a bit until it's ready for merging.

osma:
Rebased on current master (after merging PR #512 to add the `--jobs` parameter) and force-pushed.

osma:
I think this is ready for initial review. The QA tools are still complaining a bit but the remaining problems are a bit difficult to fix.

The BaseWorker class and other new code structures (e.g. functions in `annif.lexical.mllm` instead of methods) may need some explanation. The multiprocessing module allows tasks (functions with no side effects) to be executed within a pool of worker processes running in parallel on multiple CPUs. This is used to preprocess the training documents first into "candidates" (possible matching subjects) and further into "features" (numeric values that are used as input to the classifier). Most of MLLM training time is spent doing these transformations, so parallelizing them makes a lot of sense.

But the challenge here is that turning text into candidates, or candidates into features, requires some extra information in the form of pretty large data structures; e.g. the analyzer and vocabulary index in the first operation, and several large matrices in the second operation. The worker processes need to have (read-only) access to these structures. It would be possible to pass these to the worker function as parameters on every invocation, but that would be very inefficient as it would involve serializing them many times (communication between the master process and worker processes is done using serialized objects). It's better to give them to each worker only once, when it starts up. The multiprocessing module provides a facility for this - when creating a pool you can pass an initializer function with arguments, that is called only once when each worker process is created. But the worker process needs to store these somewhere so that they can be accessed by the worker function, which is called separately. The traditional solution is to use global variables for storage, but that's pretty ugly (even though harmless in the sense that each worker process is isolated from everything else so the "global" only applies to a very small world just executing a single function).

Now I found [this blog post](https://thelaziestprogrammer.com/python/multiprocessing-pool-a-global-solution) explaining that it's possible to use a class with two static methods instead - an initializer method and the worker method - and store the data structures as static attributes within the class object. That's basically equivalent to global variables, but at least the data is now in a neatly packaged namespace instead of just dangling in the global namespace.

Abstracting a bit more, I decided to create a BaseWorker class that already has a generic `init` classmethod that will store any data given to it as a dict in a static attribute within the class. Subclasses of this class (two in this PR, for the operations explained above) then only have to implement the actual worker function as a classmethod.

The worker methods need to do some pretty complex operations, e.g. `candidates_to_features`. This used to be a method of `MLLMModel`, but it's difficult to call methods of this class from the worker functions as there is no reference to any MLLMModel available. So instead I moved them to independent functions in the `annif.lexical.mllm` namespace, that can be called both from the worker functions (in the worker processes running in parallel) as well as from within MLLMModel methods. The downside of this is that these functions take a lot of parameters - and QA tools are rightly complaining about this. I couldn't figure out a more elegant solution for now, although it could of course be possible to package up the parameters a bit by putting them e.g. inside a dict instead of passing them individually, which would decrease the number of function parameters.

I think this pattern (implementing subclasses of BaseWorker) could be applied elsewhere in Annif code as well in the future, whenever there is an opportunity to process things in parallel to get a speedup on multicore systems. For example per-document data preprocessing is a typical use case where data parallelism could work. The downside is more complex, and often uglier code.

osma:
I think this is now getting ready for merging. There are still complaints from QA tools but it's getting hard to fix them in a sensible way.

PR #512, opened 2021-08-13
Support jobs parameter in train command

osma:
This PR adds a new `jobs` parameter for the `annif train` command, similar to the one already implemented in `eval` and `hyperopt` commands. The intent is to make it easier to control the amount of threads/CPUs used for training, for those backends that can make use of parallel processing during training.

The PR adds this support to the fasttext and Omikuji backends, for which this was simple to implement.

Adding as draft PR for now, since this needs more testing and QA tool checks.

This is a sideshoot of PR #511 which will benefit from this, but I thought it was cleaner to implement it as a separate PR.

### DIFF
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -54,15 +54,15 @@ class AnnifBackend(metaclass=abc.ABCMeta):
             backend_params.update(params)
         return backend_params
 
-    def _train(self, corpus, params):
+    def _train(self, corpus, params, jobs=0):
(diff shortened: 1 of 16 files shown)


osma:
Fixed test coverage. Tested this with fasttext and omikuji backends; by default both will use all CPUs, but setting the `--jobs` parameter controls the number of CPUs as it should.

Ready for review.

juhoinkinen (approved):

PR #513, opened 2021-08-16
Keep serialized dump of SKOS graph to save parsing time

osma:
SKOS vocabulary parsing can be pretty slow. With the MLLM backend, the SKOS vocabulary needs to be parsed from a Turtle file every time models are trained. With current YSO this takes around a minute.

If we instead serialize the rdflib.Graph using joblib during loadvoc, it will be much faster to load during train runs.

I tested two versions (on i7-8550U laptop), one with uncompressed dumps and another with gzip compression, using the Annif-tutorial yso-nlf data set and the train command from the MLLM exercise. Note that the benchmark wasn't very scientific.

**-**|**loadvoc (s)**|**train mllm (s)**|**dump size**
-----|:-----:|:-----:|:-----:
Before|72|893|-
Uncompressed|106 (+34)|844 (-49)|51MB
Compressed|102 (+30)|856 (-37)|22MB

The gzip version seems better. Basically, this will add ~30s to `loadvoc` but subsequent MLLM training takes ~40s less.

Opening a draft PR to get feedback from QA tools. Unit tests are still needed.

### DIFF
--- a/annif/corpus/skos.py
+++ b/annif/corpus/skos.py
@@ -1,5 +1,6 @@
 """Support for subjects loaded from a SKOS/RDF file"""
 
+import os.path
 import shutil
 import joblib
 import rdflib
@@ -80,7 +81,8 @@ class SubjectFileSKOS(SubjectCorpus):
 
         if self.path.endswith('.ttl'):
             # input is already in Turtle syntax, no need to reserialize
-            if self.path != path:
+            if not os.path.exists(path) or \
+               not os.path.samefile(self.path, path):
                 shutil.copyfile(self.path, path)
         else:
             # need to serialize into Turtle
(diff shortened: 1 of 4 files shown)


osma:
I think this is now ready for merging unless there are any ideas for improvement or objections...

juhoinkinen (approved):
LGTM

niko2342:
Hi,

to update an existing vocabulary with more than 1.3 million of concepts takes hours, because the existings graph is loaded and each triple is checked. I think it would be an good improvement to add an `--overwrite` or `--force` parameter to the `loadvoc` command. After the existing graph is deleted the `loadvoc` command behaves as it would load a new vocabulary.

What do you think?

Kind regards
Nico

osma:
Hi @niko2342 , can you please open a new issue with your suggestion?

PR #517, opened 2021-08-18
Skip unimportant subjects in _vector_to_list_suggestion

osma:
This one-line PR fixes a performance bug in the VectorSuggestionResult._vector_to_list_suggestion method. As the (pre-existing) comment on the line hints, the conversion from VectorSuggestionResult to ListSuggestionResult needs to consider only the subjects with scores above 0, and since the highest scores are processed first, we can stop once we encounter a zero score and skip the remaining subjects.

But instead of `break` there was a `continue` statement, meaning that all subjects would still be checked. This is very inefficient, especially if the vocabulary is large.

### DIFF
--- a/annif/suggestion.py
+++ b/annif/suggestion.py
@@ -103,7 +103,7 @@ class VectorSuggestionResult(SuggestionResult):
         for subject_id in self.subject_order:
             score = self._vector[subject_id]
             if score <= 0.0:
-                continue  # we can skip the remaining ones
+                break  # we can skip the remaining ones
             subject = subject_index[subject_id]
             hits.append(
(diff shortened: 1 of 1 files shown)


osma:
Quick benchmark results with MLLM, yso-fi vocabulary (~40k subjects), `eval -j 2` on the "kirjastonhoitaja" test set (n=312):

- before PR: 165s user, 1m31s wall time
- after PR: 152s user, 1m26s wall time

The difference is much bigger when using a larger vocabulary. With GND (1.3M subjects), this saves nearly 4 seconds per document when using MLLM backend.

PR #530, opened 2021-09-15
Add deprecation warning to Maui & vw_multi train commands

juhoinkinen:
Maui and vw_multi backends will be removed in release 0.56. This PR makes Annif show a deprecation warning when training these backends.


juhoinkinen:
I already added deprecattion notices to relevant Annif wiki pages.

PR #534, opened 2021-10-11
Adjust flask and click versions to avoid dependency mismatches

### LINKED ISSUE #533: Flask, Connexion, Click Dependency Mismatches
I observed a dependency mismatch in `annif==0.54.*` In fact I got this also for older versions.
This leads to an error exit code when using `pipenv`. I think you don't support `pipenv` at the moment. I still wanted to report this as other projects may want to use annif as dependency when using `pipenv`. Additionally the dependency conflict is still there even if it is not reported via exit code when using `pip`.

Steps to reproduce:
```
mkdir /tmp/pipenv_test
cd /tmp/pipenv_test
pipenv --python 3.8
pipenv install "annif==0.54.*"
```
This threw an error with respect to the version of `Werkzeug`: `Could not find a version that matches werkzeug<2.0,>=1.0,>=2.0`
Turns out that the most recent published version of `Connexion` requires `Flask<2.0` But Pipenv installed `Flask 2.0.2` I tried to install `Connexion` first. This would install `flask 1.1.4` which has a dependency `click<8.0,>=5.1`. This conflicts with the annif dependency specification `'click==8.0.*',`
Since the pinning is probably on purpose, I hand this over to you.
**Hint**: The current master of Connexion sets `'flask>=1.0.4,<3'` So this will probably be resolved in the future.

When installing using `pip` without `pipenv`, as done for the Docker container, this will print an error and install `click==8.0.3`. The exit code will indicate a success.

As a workaround with `pipenv` you can do the following:
```
pipenv shell
pip install "annif==0.54.*"
```
But is is not possible to add annif to a Pipfile.

osma:
This PR downgrades Click back to version 7.x and specifies that we don't want to use Flask 2, since it isn't compatible with the current version of Connexion that we're using (2.9.0). Hopefully this will fix the problem of installing Annif via `pipenv`

ping @mo-fu does this work for you?

Fixes #533 (hopefully)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,9 +22,9 @@ setup(
     install_requires=[
         'connexion[swagger-ui]',
         'swagger_ui_bundle',
-        'flask',
+        'flask>=1.0.4,<2',
         'flask-cors',
-        'click==8.0.*',
+        'click==7.1.*',
         'click-log',
         'joblib==1.0.1',
         'nltk',


mo-fu:
lgtm

PR #536, opened 2021-10-18
Allow rdflib version 6.*

mo-fu:
This relaxes the rdflib version dependency as requested by Nicholas Car.
I decided to keep the requirements loose. This allows the use of older versions of rdflib. When using `rdflib==6.*` with old Annif projects, the corresponding vocabularies have to be reloaded. Probably require retraining of projects.

Corresponding stwfsapy PR: https://github.com/zbw/stwfsapy/pull/40


osma:
Needs to be tested. Otherwise this should be trivial, but since Annif nowadays dumps the rdflib Graph object to disk in the loadvoc command, it might break if the original loadvoc is done using an older rdflib version. To test this

1. use Annif with old rdflib (5.x)
2. run `annif loadvoc` on some MLLM project
3. upgrade rdflib (6.x)
4. train the MLLM project

juhoinkinen:
I tested the above steps, and training fails with an error `ModuleNotFoundError: No module named 'rdflib.plugins.memory'`, but as @Osma said this is resolved by removing `subjects.dump.gz` file from the vocab directory (that forces recreation of the dump file). 

However, existing MLLM models keep working straightaway when rdflib is updated 5.0.0 -> 6.0.2.

Two minor points to note:
- rdflib 6 dropped support for Python <3.7
- In unit tests with version 6.0.* there are shown a bit confusing DeprecationWarnings: `graph.load() is deprecated, it will be removed in rdflib 6.0.0.` and `graph.preferredLabel() is deprecated and will be removed in rdflib 6.0.0.`
 
Seems that the deprecations did not end up to the [version 6 release](https://github.com/RDFLib/rdflib/releases/tag/6.0.0) as was intended at some point, but [the warning were neither updated](https://github.com/RDFLib/rdflib/issues/1356).

osma:
Thanks for testing @juhoinkinen !

> I tested the above steps, and training fails with an error ModuleNotFoundError: No module named 'rdflib.plugins.memory', but as @osma said this is resolved by removing subjects.dump.gz file from the vocab directory (that forces recreation of the dump file).

Could this be handled more gracefully in the code, for example by catching the error and then skipping loading of the dump file? Ideally the dump would be recreated with the new rdflib version so that subsequent train operations could again use it.

> rdflib 6 dropped support for Python <3.7

Ah, good catch. Annif still supports 3.6 so we need to maintain support for rdflib 5 as well. I think all should be well, because the new version declaration is `rdflib>=4.2,<7.0` which allows older versions as well.

> In unit tests with version 6.0.* there are shown a bit confusing DeprecationWarnings: graph.load() is deprecated, it will be removed in rdflib 6.0.0. and graph.preferredLabel() is deprecated and will be removed in rdflib 6.0.0.

I suppose the same operations can be done in a different way to avoid the warnings? Would doing that break with older rdflib versions? I think we want to avoid triggering these.

juhoinkinen:
I made a new branch [allow-rdflib-version-6](https://github.com/NatLibFi/Annif/tree/allow-rdflib-version-6) under NatLibFi's repo with commits resolving the discussed issues. Could that be merged to @mo-fu's fork so this PR would also contain the commits, or is making a new PR easier?

osma:
>  Could that be merged to @mo-fu's fork

It seems to me that @mo-fu has allowed pushing to the PR branch. The GitHub UI says (just below the comments here):

> Add more commits by pushing to the master branch on mo-fu/Annif.

So you should be able to push those commits there on your own. Some guidance:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

https://tighten.co/blog/adding-commits-to-a-pull-request/

osma (approved):

PR #527, opened 2021-09-01
Add spaCy analyzer

osma:
Initial draft PR of new spaCy based (optional) analyzer.
Fixes #374

TODO items:

- [x] Option to force lowercasing of lemmas
- [ ] ~~Test with Swedish (which doesn't have a complete pretrained model) and adapt the code as necessary~~ Out of scope for now
- [x] Adapt YAKE backend so it doesn't do word-by-word normalization
- [x] Add English simple model to Docker image
- [x] Add wiki documentation: what the analyzer is, how to install, how to add more languages (native install & Docker case)

Issue #374, opened 2019-12-20
spaCy analyzer


osma:
We could add an analyzer based on spaCy. It would enable support for some new languages and possibly also give better results for eg English and German than the current snowball analyzer.

Getting the full benefit of spaCy may require some internal API changes because it is more object oriented than NLTK and processes whole sentences instead of just individual words, taking some of the context into account.

This would be an optional feature as spaCy is implemented as a native code extension, not just pure Python.

osma:
spaCy also has some support for document (text) categorization, both multiclass and multilabel. This could be supported as an Annif backend: https://spacy.io/api/textcategorizer

osma:
Starting to look at this more seriously...

One immediate issue that comes up is that spacy (3.1.2) depends on [typer](https://github.com/tiangolo/typer) (0.3.2) which at the moment depends on click==7.1.2, while Annif depends on click==8.0.1 since PR #499 was merged just before the 0.53 release.

Now this appears to be fixed with [typer 0.4.0](https://github.com/tiangolo/typer/releases/tag/0.4.0) that was released yesterday(!) and supports Click 8, but it won't help until there's a newer version of spaCy available that upgrades the typer dependency.

For now we may have to downgrade back to Click 7.1.2, which probably isn't a problem for Annif since I don't think we've started using any Click 8 features yet.

PR #540, opened 2021-12-06
Add XTransformer backend

mo-fu:
This PR adds XTransformer as an optional backend to Annif. For now it does not yet use distilbert in the default configuration as this is not yet available on pypi.
The tests for the backend resort to mocking as training would download a pretrained model of size at least 500 mb. 
Also we should discuss cache directories. At the moment xtransformer will download models from the huggingface hub to `~/,cache/huggingface` Is this behavior desired for Annif or should the cache be placed in the `data` folder?
I also haven't modified the docker container yet. When I installed pecos in a venv it required BLAS libraries so this would probably have to be added to the container. Additionally pecos will install the GPU enabled pytorch. Meaning the container size will grow. Therefore I wanted to check with you first before adding it.


osma:
Thanks @mo-fu , great work! I like how you also enhanced `atomic_save` and added `apply_param_parse_config` which you then used also in existing backends.

>  The tests for the backend resort to mocking as training would download a pretrained model of size at least 500 mb.

Sounds reasonable.

> At the moment xtransformer will download models from the huggingface hub to ~/,cache/huggingface Is this behavior desired for Annif or should the cache be placed in the data folder?

The punkt model for NLTK is currently handled in a similar way - it's normally placed under `~/nltk_data/`. That's maybe OK for NLTK (it's only 49M) but I agree that it could make sense to place downloaded models under the data directory, as you suggest. It's also a bit similar for the new spaCy analyzer (#374) which also needs to store downloaded models and those can be quite large.

How about `data/model_cache` or something along those lines? Then that could be further split up with subdirectories named by type/package, for example `huggingface`, `spacy`, `nltk` etc.

> I also haven't modified the docker container yet. When I installed pecos in a venv it required BLAS libraries so this would probably have to be added to the container. Additionally pecos will install the GPU enabled pytorch. Meaning the container size will grow. Therefore I wanted to check with you first before adding it.

So far we've tried to support all features in the container. So following that policy, we would need to add this to the container as well. But I guess it also depends on how much the container would grow? If it's very much (50% or more) then we could consider making a separate flavor of the container that includes XTransformer, or requiring the user to build their own container with XTransformer support if they need it. Does @juhoinkinen have a comment?

Also, can you give an example for how to train and test this backend for example with the tutorial data sets?

annif/util.py
```diff
@@ -13,21 +14,31 @@ def atomic_save(obj, dirname, filename, method=None):
...
     filename, using a temporary file and renaming the temporary file to the
-    final name."""
-
-    prefix, suffix = os.path.splitext(filename)
-    tempfd, tempfilename = tempfile.mkstemp(
-        prefix=prefix, suffix=suffix, dir=dirname)
-    os.close(tempfd)
+    final name. To save a directory explicitly set filename=None."""
+
+    if filename:
```
osma, line 19: There isn't that much shared code between the file vs. directory cases. I wonder if it would be better to leave the `atomic_save` function as it is (only supporting files, not directories) and adding a new function `atomic_save_dir` which takes care of directories only. The implementations could be simpler, without so many if clauses.

juhoinkinen:
Installing Pecos increases the Docker image size quite much: 1.5 GB -> 3.5 GB (download size of PyTorch is nearly 900 MBs).

So it looks like it would be best to have a separate image that includes also Pecos.

First I thought making a `Dockerfile.custom` that a user could customize and build to have an image with only the optional backends they need, possibly with XTransformer. But actually already the default Dockerfile is quite easily customisable (the optional backends are installed in just one line, except for fastText that is built in the builder stage). It would be simpler to just have a commented line with `pip install .[pecos]` in the default Dockerfile. (In Wiki there could be instructions to remove the other optional backends if they are not needed.) 

But again, now I noticed that instead of selecting dependencies to install by commenting/uncommenting lines [build-args](https://docs.docker.com/engine/reference/builder/#arg) could be used. I try how that would work.

However, should we offer an image flavor in quay.io that includes Pecos (with a `pecos` tag or similar)?

osma:
I'm sorry, I just caused a conflict in this PR by merging #544, which optimizes the startup time of Annif and contains a complete rewrite of `annif/backend/__init__.py`. You will need to adjust your code accordingly. The pattern should be quite obvious - just remember to:

* keep alphabetical order (xtransformer should be the penultimate backend, just before yake)
* look at the code for the other optional backends (fasttext, omikuji, nn_ensemble, yake) and use the same approach
* add a unit test for the missing package into `tests/test_backend.py`

osma:
I see that you fixed the conflicts and also addressed some other suggestions @mo-fu - great!
Could you give an example too:
> Also, can you give an example for how to train and test this backend for example with the tutorial data sets?

osma:
> But again, now I noticed that instead of selecting dependencies to install by commenting/uncommenting lines [build-args](https://docs.docker.com/engine/reference/builder/#arg) could be used. I try how that would work.

This was already implemented in PR #548 (and extended for spaCy models in PR #527), both released as part of 0.56 and [documented](https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker#customizing-docker-image) in the wiki. So we already have the mechanism in place for customizing Docker images with build args.

> However, should we offer an image flavor in quay.io that includes Pecos (with a pecos tag or similar)?

Maybe - let's postpone that discussion a little bit.

mo-fu:
A small status update. To figure out good parameter values for the YSO data set from the tutorial repository. Unfortunately the suggest command is really slow. Presumably since I run the parameter optimization on a GPU and copying the representation of every input individually takes a lot of time. I will try whether this changes when using the CPU. But maybe in the future it is possible to explore how parallel evaluation can be improved for backends that support it directly.

osma:
Thanks for the update @mo-fu !
It makes a lot of sense to do parallel evaluation (and other operations as well) directly with backends that support many documents at a time. I've had similar thoughts about other backends that could do this but the limiting factor is the current `suggest` method that only supports one document at a time. I created a new issue #579 for implementing a batched version of `suggest`.

mo-fu:
Rebased the code. If you want to try it you can use the following config:
```
[xtransformer-yso-en]
name=XTransformer YSO english
backend=xtransformer
analyzer=simple
language=en
vocab=yso
batch_size=16
truncate_length=256
learning_rate=0.0001
num_train_epochs=1
max_leaf_size=18000
```
I kept the training epochs low so it is possible to think about training it on a CPU
As this backend does not generalize well from titles to full text I trained it on the `yso-finna.tsv` file from the tuorial repo but removed all publications in the `yso-finna-small.tsv` file. Evaluating it on `yso-finna-small.tsv` gave the following results:
```
Precision (doc avg):            0.324832
Recall (doc avg):               0.7030010309769649
F1 score (doc avg):             0.40509639141725173
Precision (subj avg):           0.16515053567940333
Recall (subj avg):              0.3300418931089765
F1 score (subj avg):            0.20597572080373455
Precision (weighted subj avg):  0.36869694324079433 
Recall (weighted subj avg):     0.6572207238832092
F1 score (weighted subj avg):   0.46264001204230565
Precision (microavg):           0.3253786361086626
Recall (microavg):              0.6572207238832092
F1 score (microavg):            0.43526505606768456                                                                                                       
F1@5:                           0.47701308122837444
NDCG:                           0.6797521954650343
NDCG@5:                         0.673154809582077
NDCG@10:                        0.6904624877618667
Precision@1:                    0.74708
Precision@3:                    0.5939966666666666
Precision@5:                    0.485896
LRAP:                           0.5811457115235024
True positives:                 324832
False positives:                673488
False negatives:                169419
Documents evaluated:            100000
```
On CPU this took two around days and five hours
For comparison 5 epochs gave worse results:
```
Precision (doc avg):            0.280435
Recall (doc avg):               0.6120439409375019
F1 score (doc avg):             0.34979511939886493
Precision (subj avg):           0.1665384503773252
Recall (subj avg):              0.3293110331573767
F1 score (subj avg):            0.18618029292260818
Precision (weighted subj avg):  0.5068295328710835
Recall (weighted subj avg):     0.5673938950047648
F1 score (weighted subj avg):   0.49945040364924664
Precision (microavg):           0.2809069236317013
Recall (microavg):              0.5673938950047648
F1 score (microavg):            0.37577441877136836
F1@5:                           0.4327546591641913
NDCG:                           0.6145451297207569
NDCG@5:                         0.6258527543922358
NDCG@10:                        0.6244413738840844
Precision@1:                    0.72693
Precision@3:                    0.5587466666666666
Precision@5:                    0.442632
LRAP:                           0.5154182741105832
True positives:                 280435
False positives:                717885
False negatives:                213816
Documents evaluated:            100000
```

osma:
This is great, thanks a lot @mo-fu ! We will need to test this on other data sets as well. The F1 score isn't bad at all considering that these are simple short text records. OTOH, Finna contains a lot of duplicates and near-duplicates so it's possible the model has a chance to cheat - we don't normally evaluate on the Finna data set.

Just one question: where does the language model come from? How do you define which one to use? Is it downloaded automatically or do you have to place it somewhere?

mo-fu:
The default language model is currently defined as https://huggingface.co/distilbert-base-multilingual-cased I opted for the multilingual as default so you don't have to worry that much about model selection.
You can try different models by setting the `model_shortcut parameter` to a name from the hugginface hub (Files may also be possible). But it has to be based on either a bert, distilbert, roberta, xlnet or xlmroberta model. The distilbert one  is the best choice for decreasing training time.

osma:
Excellent, thanks!

mo-fu:
The `model_shortcut` parameter only support models from the huggingface hub. Pecos supports local files but this parameter is currently not exposed. If you want I can add it.

osma:
> Pecos supports local files but this parameter is currently not exposed. If you want I can add it.

That would be a great addition! I think we are likely to test some experimental language models in the near future, so not having to go through the hub would be helpful.

juhoinkinen:
Sorry, I did not close this intentionally. I planned to test this (and maybe add batch processing, for which support has been recently implemented), so I merged current master to the PR branch, fixed conflicts, and pushed. GitHub showed very many commits that did not belong to this PR, and they could have been removed by switching the base branch to some other and then back (like in https://github.com/NatLibFi/FintoAI/pull/12#issuecomment-1400568758). 

So I switched the base branch to api-i18n, but then the PR got closed automatically, and I cannot reopen it, apparently because the PR and base branches share no history ("mo-fu:master branch has no history in common with NatLibFi:api-i18n"). And the base branch of closed PR cannot be changed... 

I tried to merge current master to api-i18n, but that did not help. Maybe api-i18n could be completely rewritten to allow the PR to be reopened.

juhoinkinen:
I asked from GitHub support if they can switch the base branch back to master to allow reopening this PR.

mo-fu:
@juhoinkinen I added you as a collaborator to my fork. Maybe this helps. I can also try the `git pull origin api-i18n --allow-unrelated-histories` later.

juhoinkinen:
Thanks @mo-fu, now I was able to do the `git pull origin api-i18n --allow-unrelated-histories` and push, but the commits did not appear in this PR. This seems different to the [behavior that was reported in SO](https://stackoverflow.com/questions/62479813/github-how-to-re-open-a-pull-request-when-the-branch-has-no-more-history-in-com/62498360#comment110513475_62479813).

I also tried to use [GitHub API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#update-a-pull-request) to change the base branch back to master and re-open the PR, but with no luck, the response was: `"Cannot change the base branch of a closed pull request."`

It seems there is no way this PR can be re-opened :( Or maybe rebasing the whole [mo-fu:master](https://github.com/mo-fu/Annif) branch on [NatLibFi:api-i18](https://github.com/NatLibFi/Annif/tree/api-i18n), so the branches would definitely have a common history...? But then it could be that GitHub just would not detect the change for this PR, because I think pulling api-i18n should already have made a common history.

The easiest way to proceed would be to make a new PR from the commit [Merge branch 'master' of github.com:mo-fu/Annif into mo-fu-master](https://github.com/mo-fu/Annif/commit/aa96ebc2ad908dda07d82a6b10e0d264e6e38a1d), which has the (nearly) current master from NatLibFi merged (mo-fu:master could be just resetted to that commit to remove the merge of api-i18n).

juhoinkinen:
> The easiest way to proceed would be to make a new PR from the commit [Merge branch 'master' of github.com:mo-fu/Annif into mo-fu-master](https://github.com/mo-fu/Annif/commit/aa96ebc2ad908dda07d82a6b10e0d264e6e38a1d), which has the (nearly) current master from NatLibFi merged (mo-fu:master could be just resetted to that commit to remove the merge of api-i18n).

Hi @mo-fu! Would you like to open a new PR for XTransformer? So that GitHub history will show credit of it to you. No need to mind about the PR description, we can edit it.

I could also open the PR from your fork, if you dont want to bother with this.

PR #541, opened 2021-12-08
Remove Maui and Vowpal Wabbit multi backends

juhoinkinen:
This PR removes the Maui and vw_multi (Vowpal Wabbit) backends.

The removal makes codebase more compact and easier to maintain. The [MLLM](https://github.com/NatLibFi/Annif/wiki/Backend%3A-MLLM) and [nn_ensemble](https://github.com/NatLibFi/Annif/wiki/Backend%3A-nn_ensemble) backends offer similar functionality as Maui and vw_multi.


juhoinkinen:
No significant test coverage decrease (or really maybe not any, I wonder if CodeCov reporting is right).

But what to do for the Wiki pages of the backends? Just remove them, hide or move somewhere?

juhoinkinen:
I replaced the mentions of Maui to MLLM in Wiki (e.g. in baseprojects of ensembles) or removed them, but the [Maui](https://github.com/NatLibFi/Annif/wiki/Backend:-Maui) and [vw_multi](https://github.com/NatLibFi/Annif/wiki/Backend%3A-vw_multi) pages themselves are still in place. 

The page for [vw_ensemble](https://github.com/NatLibFi/Annif/wiki/Backend:-vw_ensemble) also still exists but is kind of hidden (not in the main backends list, only in the collapsable Pages list).

osma (changes requested):
LGTM! But I noticed that the `responses` package was only used by the Maui tests, so that dependency can be removed as well. I can do that myself and then merge this.

osma:
The wiki pages look good. We should remove the links to the Maui and vw_multi pages from the TOC and Home pages after the 0.56 release. The pages can still be there (like vw_ensemble) but just not linked from anywhere.

juhoinkinen:
> The wiki pages look good. We should remove the links to the Maui and vw_multi pages from the TOC and Home pages after the 0.56 release. The pages can still be there (like vw_ensemble) but just not linked from anywhere.

Removed the links.

PR #543, opened 2021-12-17
Optimize startup time using lazy imports

### LINKED ISSUE #514: Optimize startup time with lazy imports
Annif takes several seconds to start even when it's doing nothing but printing the version number or help text:

```
$ time annif --version
0.54.0.dev0

real	0m4,398s
user	0m4,322s
sys	0m0,470s
```

I investigated this a little bit using the `-X importtime` feature in Python 3.7+ and the [tuna](https://github.com/nschloe/tuna) tool for visualizing profiling information. It seems that the time is mostly spent importing large libraries such as tensorflow, scikit-learn, optuna, connexion and nltk:

kuva

These libraries are all unnecessary in simple operations such as `annif --help` and `--version` so it would be better to avoid importing them altogether. There are some tutorials on lazy importing (e.g. [this one](https://levelup.gitconnected.com/python-trick-lazy-module-loading-df9b9dc111af)) and the `importlib` library contains (since Python 3.5) a [LazyLoader](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) utility class that could be used here.

I experimented a bit with this `lazy_import` function but couldn't get it to work for nltk submodules:

```python
# Adapted from: https://stackoverflow.com/questions/42703908/
def lazy_import(fullname):
    """lazily import a module the first time it is used"""
    try:
        return sys.modules[fullname]
    except KeyError:
        spec = importlib.util.find_spec(fullname)
        module = importlib.util.module_from_spec(spec)
        loader = importlib.util.LazyLoader(spec.loader)
        # Make module with proper locking and get it inserted into sys.modules.
        loader.exec_module(module)
        return module
```

This needs more experimentation but for now I'm just opening the issue...

osma:
Fixes #514 

The goal of this PR is to reduce CLI startup time by avoiding useless work, especially imports that are not necessary for the requested operation.

It makes the following changes to the import statements within the Annif codebase:

- complete rewrite of `annif/backend/__init__.py`; the end result is that backends (and the libraries they require, e.g. fasttext, omikuji and tensorflow) are only imported when they are actually used
- avoid importing NLTK, sklearn, joblib and NumPy unless actually required, by moving import statements inside functions and methods

I tried to craft the changes to have minimal impact on the code so I only chose to make imports local in cases where there were very few uses within the same module. For example `annif.suggestion` uses NumPy a lot, so I didn't try to use local imports within that module, but instead avoided importing the whole `annif.suggestion` module in other parts of the codebase.

Startup time for simple commands such as `annif --help` and `annif --version` has been reduced by more than two thirds.

Before:

```
$ time annif --version
0.56.0.dev0

real	0m4,052s
user	0m4,001s
sys	0m0,568s
```

After:

```
$ time annif --version
0.56.0.dev0

real	0m1,302s
user	0m1,250s
sys	0m0,050s
```

As explained in #514, I also used [tuna](https://github.com/nschloe/tuna) to visualize where the remaining import time is spent after this PR:

image

The main culprits are now `connexion` (with most of the time spent initializing `openapi_spec_validator`!) and `flask`. Those are core libraries and I don't think we can avoid importing them even for the simplest CLI commands.

### DIFF
--- a/annif/analyzer/analyzer.py
+++ b/annif/analyzer/analyzer.py
@@ -3,7 +3,6 @@
 import abc
 import functools
 import unicodedata
-import nltk.tokenize
 
 _KEY_TOKEN_MIN_LENGTH = 'token_min_length'
 
@@ -22,6 +21,7 @@ class Analyzer(metaclass=abc.ABCMeta):
 
     def tokenize_sentences(self, text):
         """Tokenize a piece of text (e.g. a document) into sentences."""
+        import nltk.tokenize
         return nltk.tokenize.sent_tokenize(text)
 
     @functools.lru_cache(maxsize=50000)
@@ -37,6 +37,7 @@ class Analyzer(metaclass=abc.ABCMeta):
 
     def tokenize_words(self, text):
         """Tokenize a piece of text (e.g. a sentence) into words."""
+        import nltk.tokenize
         return [self.normalize_word(word)
                 for word in nltk.tokenize.word_tokenize(text)
                 if self.is_valid_token(word)]
--- a/annif/analyzer/snowball.py
+++ b/annif/analyzer/snowball.py
@@ -1,7 +1,6 @@
 """Snowball analyzer for Annif, based on nltk Snowball stemmer."""
 
 import functools
-import nltk.stem.snowball
 from . import analyzer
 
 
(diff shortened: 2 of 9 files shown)


osma:
I also looked at the option of implementing truly lazy imports, for example as [described](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports) in the importlib standard module. But that didn't seem very appealing - it takes a nontrivial amount of code and can be hard to debug if the imports fail for some reason. So I chose to use local imports instead, which are easier to understand if not always so elegant.

osma:
Looking at CI test results and especially Codecov, I think we need additional tests for the ImportError/ValueError clauses in `annif/backend/__init__.py`

osma:
I will try again with a less invasive approach that doesn't try to avoid importing NumPy. It's not worth complicating the code with many local imports just to speed up the startup by less than 0.1 seconds.

PR #548, opened 2021-12-21
Allow selecting installed optional dependencies in Docker build

juhoinkinen:
Currently we provide a Dockerfile and an image in quay.io with dependencies for all the optional features.

If someone does not need all the features, but e.g. uses only Omikuji backend and would like build their own image (e.g. for smaller size), they could edit the Dockerfile dropping the unwanted (Python) dependencies from the pip install command. However, using `--build-arg` option of `docker build` provides a way to pass the user-defined list of dependencies to install that overrides the default list:
```
docker build -t annif-omikuji-voikko --build-arg optional_dependencies=omikuji,voikko .
```
A user wanting to have a Docker image to use with Xtransfomer could add `pecos` to the default list in build time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,16 @@ FROM python:3.8-slim-bullseye AS builder
 
 LABEL maintainer="Juho Inkinen <juho.inkinen@helsinki.fi>"
 
-# Install fastText, which needs to be built, and therefore also some system packages:
-RUN apt-get update \
-	&& apt-get install -y --no-install-recommends \
-		build-essential \
-	&& pip install --no-cache-dir \
-		fasttext==0.9.2
+SHELL ["/bin/bash", "-c"]
+ARG optional_dependencies=dev,voikko,pycld3,fasttext,nn,omikuji,yake
+# Bulding fastText needs some system packages
+RUN if [[ $optional_dependencies =~ "fasttext" ]]; then \
+		apt-get update && \
+		apt-get install -y --no-install-recommends \
+			build-essential && \
+		pip install --no-cache-dir \
+			fasttext==0.9.2; \
+	fi
 
 
 FROM python:3.8-slim-bullseye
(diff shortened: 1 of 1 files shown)


osma:
What about fasttext? It will be installed regardless of this setting, right?

juhoinkinen:
> What about fasttext? It will be installed regardless of this setting, right?

Yes, that's true, I thought it would not be installed... Or at least not available right away. Hmm, I don't if its copying from builder stage can be controlled this way conveniently enough.

osma:
I don't think always including fasttext is a very big deal, this PR is still a big improvement.

That said, I think it could be possible to avoid installing fasttext by making the `pip install` command on line 9 conditional. Something like:

```bash
if [[ $optional_dependencies =~ "fasttext" ]]; then pip install --no-cache-dir fasttext==0.9.2; fi
```

(perhaps the ` apt-get install build-essential` command could be moved inside the if clause as well, to avoid doing useless extra work in case fasttext isn't needed)

The COPY command would then copy an empty /usr/local/lib/python3.8 directory from the builder image.

Dockerfile
```diff
@@ -26,9 +30,11 @@ RUN apt-get update \
 WORKDIR /Annif
 RUN pip install --upgrade pip --no-cache-dir
 
-# Install all optional dependencies:
 COPY setup.py README.md LICENSE.txt projects.cfg.dist /Annif/
-RUN pip install .[dev,voikko,pycld3,fasttext,nn,omikuji,yake] --no-cache-dir
+# Install dependencies for optional features.
+ARG optional_dependencies=dev,voikko,pycld3,fasttext,nn,omikuji,yake
```
osma, line 35: ARG is now declared twice? Isn't once enough?

juhoinkinen: Once is not enough; each stage needs have its own declaration: https://docs.docker.com/engine/reference/builder/#scope

osma: Ah, OK. Thanks for the clarification!

osma (approved):
LGTM

juhoinkinen:
I added a section of customizing Docker image to Wiki: https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker#customizing-docker-image

PR #549, opened 2022-01-07
Update dependencies v0.56

juhoinkinen:
Dependency updates for Annif v0.56.

Updates to (edited):
- Omikuji 0.4.x
- TensorFlow 2.7.0 & lmbd 1.3.0
- Gensim 4.1.x
- joblib 1.1.0
- Optuna 2.10.x
- SciPy 1.7.x
- Scikit-learn 1.0.2
- NumPy 1.21.x
- Click 8.0.x
- Flask >=1.0.4,<3
- Connexion to use [connexion2 PyPI project](https://pypi.org/project/connexion2/)

Below is a table of Omikuji 0.3.4 and 0.4.1 models comparison. Training was performed on full Finnish Finna dataset (real and user times include preprocessing), and evaluation in parallel.

|                    |                     | v0.3   | v0.4   |
|--------------------|---------------------|--------|--------|
| train time         |                     |        |        |
|                    | real                | 190m   | 174m   |
|                    | user                | 2061m  | 1789m  |
|                    | reported by Omikuji | 7776s  | 6783s  |
| eval time          |                     |        |        |
|                    | real                | 2m59s  | 2m3s   |
|                    | user                | 53m55s | 14m33s |
|                    | sys                 | 3m18s  | 1m47s  |
|                    |                     |        |        |
| model size on disk |                     | 3.8G   | 3.4G   |
|                    |                     |        |        |
| eval results       |                     |        |        |
|                    | JYU-theses          | 0.4716 | 0.4759 |
|                    | kirjaesittelyt2021  | 0.4511 | 0.4496 |
|                    | kirjastonhoitaja    | 0.2759 | 0.2824 |
|                    | satakunnan-kansa    | 0.2808 | 0.2524 |
|                    | vapaakappaleet-orig | 0.4648 | 0.4613 |
|                    | average             | 0.3888 | 0.3843 |

### DIFF
--- a/annif/backend/omikuji.py
+++ b/annif/backend/omikuji.py
@@ -5,7 +5,8 @@ import os.path
 import shutil
 import annif.util
 from annif.suggestion import SubjectSuggestion, ListSuggestionResult
-from annif.exception import NotInitializedException, NotSupportedException
+from annif.exception import NotInitializedException, NotSupportedException, \
+    OperationFailedException
 from . import backend
 from . import mixins
 
@@ -40,7 +41,12 @@ class OmikujiBackend(mixins.TfidfVectorizerMixin, backend.AnnifBackend):
             path = os.path.join(self.datadir, self.MODEL_FILE)
             self.debug('loading model from {}'.format(path))
             if os.path.exists(path):
-                self._model = omikuji.Model.load(path)
+                try:
+                    self._model = omikuji.Model.load(path)
+                except RuntimeError:
+                    raise OperationFailedException(
+                        "Omikuji models trained on Annif versions older than "
+                        "0.56 cannot be loaded. Please retrain your project.")
             else:
                 raise NotInitializedException(
                     'model {} not found'.format(path),
--- a/annif/backend/pav.py
+++ b/annif/backend/pav.py
@@ -95,9 +95,9 @@ class PAVBackend(ensemble.BaseEnsembleBackend):
         scores = coo_matrix((data, (row, col)),
                             shape=(ndocs, len(source_project.subjects)),
                             dtype=np.float32)
-        true = coo_matrix((np.ones(len(trow), dtype=np.bool), (trow, tcol)),
+        true = coo_matrix((np.ones(len(trow), dtype=bool), (trow, tcol)),
                           shape=(ndocs, len(source_project.subjects)),
-                          dtype=np.bool)
+                          dtype=bool)
         return csc_matrix(scores), csc_matrix(true)
 
     def _create_pav_model(self, source_project_id, min_docs, corpus):
--- a/annif/backend/tfidf.py
+++ b/annif/backend/tfidf.py
@@ -4,17 +4,12 @@ TF-IDF normalized bag-of-words vector space"""
 import os.path
 import tempfile
 import annif.util
+import gensim.similarities
+from gensim.matutils import Sparse2Corpus
 from annif.suggestion import VectorSuggestionResult
 from annif.exception import NotInitializedException, NotSupportedException
 from . import backend
 from . import mixins
-
-# Filter UserWarnings due to not-installed python-Levenshtein package
-import warnings
-with warnings.catch_warnings():
-    warnings.simplefilter('ignore')
-    import gensim.similarities
-    from gensim.matutils import Sparse2Corpus
 
 
 class SubjectBuffer:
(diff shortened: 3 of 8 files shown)


juhoinkinen:
Other points:
- Omikuji models trained with previous version do not work on the new Omikuji version. Others models stay working, also NN ensemble and TFIDF models despite TensorFlow and Gensim version updates.
- Python 3.6 & 3.7 tests show now a [`RuntimeWarning`](https://github.com/NatLibFi/Annif/runs/4740435552?check_suite_focus=true) raised by `test_english_analyzer_normalize_word`, but it has been around since #544.
- Python 3.6 tests show also many `DeprecationWarnings`.

juhoinkinen:
Some dependencies were not updated to newer versions because they do not support Python 3.6; some latest versions do not support even 3.7:
- [SciPy 1.7.x](https://github.com/scipy/scipy/releases/tag/v1.7.0) and [1.6.x](https://github.com/scipy/scipy/releases/tag/v1.6.0) require 3.7+ (to-be-released 1.8.0 will require Python 3.8+)
- [NumPy 1.20](https://github.com/numpy/numpy/releases/tag/v1.20.0) requires Python 3.7-3.9; 1.21 requires 3.7-3.10; 1.22 requires 3.8-3.10
- [Scikit-learn 1.0](https://github.com/scikit-learn/scikit-learn/releases/tag/1.0) requires Python 3.7-3.9 (stwfsa has scikit-learn pinned to 0.24.x)
- TensorFlow 2.7 has wheels and [is tested](https://www.tensorflow.org/install/source#tested_build_configurations) for Python 3.7-3.9

Should we keep Python 3.6 support for Annif?

juhoinkinen:
Sounds good, I make a separate PR for dropping Python 3.6  support.

osma:
Omikuji eval time has improved a lot in 0.4! The other numbers look good as well.

What happens when you try to use an old model with omikuji 0.4? Is there some kind of exception? Should we catch that and display a more user friendly error message?

juhoinkinen:
Rebased on master after dropping Python 3.6 support (#550) and force-pushed.

osma:
> [...] about one-screen length of traceback and RuntimeError: Failed to load model from data/projects/yso-bonsai-old-fi/omikuji-model
> This already hints something being wrong with the model, so I'm not sure if it is worth processing it within Annif...?

I think that showing tracebacks to the user is not a recipe for good UX... It would be better to show a more informative message such as "Omikuji models trained on Annif versions older than 0.56 cannot be loaded. Please retrain your project."

osma:
@juhoinkinen Since you've upgraded (again) Click to 8.0.*, we need to take care of also upgrading Connexion and Flask to avoid hitting the version mismatch problem reported in #533 once more (which was fixed by downgrading Click back to 7.1.2 in PR #544).

There's a little problem with Connexion releases though - PyPI currently only has 2.9.0 which doesn't support Flask 2. Version 2.10.0 has been released on GitHub but on the PyPI side it has a different name `connexion2`. See [here](https://github.com/zalando/connexion/releases/tag/2.10.0) for details. So we should probably switch to `connexion2` at least for now.

juhoinkinen:
Yes, I already switched to `connexion2` in Annif's requirements, but now I noticed some dependency is still using `connexion`, so both `connexion 2.7.0` and `connexion2 2.10.0`  end up to be installed. 

As we don't actually need `connexion2` or Click 8, maybe it's best to just go back to Click 8 (again)...?

juhoinkinen:
[`swagger-tester`](https://github.com/Trax-air/swagger-tester/blob/master/requirements.txt) being the dependency requiring `connexion`.

osma:
Ah, right. The dependency on `connexion` seems to come from [swagger-tester](https://pypi.org/project/swagger-tester/) which is a `dev` dependency for Annif. Its last release 0.2.12 was made in May 2018 and the project seems pretty dead. Maybe we should try to get rid of this dependency - it's used in `test_swagger.py` for some basic exercise of the REST API methods.

juhoinkinen:
I checked the compatibility of models trained with Annif 0.55 after upgrading also SciPy, Numpy and Scikit-learn, and TensorFlow again/further.

The usual warnings like `
UserWarning: Trying to unpickle estimator TfidfTransformer from version 0.24.2 when using version 1.0.2. [...]` are shown but models seem to work (suggestions and scores remain the same for a short text input).

juhoinkinen:
Rebased on master after removing swagger-tester dependency (#551) and force-pushed.

osma (approved):

PR #551, opened 2022-01-12
Remove swagger-tester dependency 

juhoinkinen:
The [swagger-tester](https://github.com/Trax-air/swagger-tester) seems unmaintained and depends on [connexion](https://github.com/zalando/connexion). Since Annif is switching to install connexion (#549) from its new PyPI project (named [connexion2](https://pypi.org/project/connexion2/)), the result is that connexion ends up to be installed twice (currently as `connexion v2.7.0` and `connexion2 v2.10.0`).

This PR removes the swagger-tester dev dependency from Annif and adds self-written tests for REST API methods, so only `connexion2` will be installed. 

Also avoids running Flask in a background thread (at least explicitly) by [using Flask's TestClient](https://flask.palletsprojects.com/en/2.0.x/testing/#the-testing-skeleton).

Now the tests cover basically only the status codes related to cases defined in [`swagger/annif.yaml`](https://github.com/NatLibFi/Annif/blob/master/annif/swagger/annif.yaml). Should there be more detailed tests as in the tests in [`test_rest.py`](https://github.com/NatLibFi/Annif/blob/master/tests/test_rest.py)? (Or are these tests already overlapping, like testing the same functionality...?)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,6 @@ setup(
             'pytest-watch',
             'pytest-flask',
             'pytest-flake8',
-            'swagger-tester',
             'bumpversion',
             'autopep8'
         ]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,8 @@ import os.path
 import shutil
 import pytest
 import py.path
+import time
+import threading
 import unittest.mock
 import annif
 import annif.analyzer
@@ -32,6 +34,21 @@ def app_with_initialize():
     app = annif.create_app(
         config_name='annif.default_config.TestingInitializeConfig')
     return app
+
+
+@pytest.fixture(scope='module')
+def app_with_server(app):
+    # run a Flask/Connexion server in a background thread
+    def run_app():
+        # We need to set this env var to 'false' because otherwise Flask
+        # thinks, due to the CLI tests that could have run before, that
+        # it has been started via its  CLI and refuses to run.
+        os.environ['FLASK_RUN_FROM_CLI'] = 'false'
+        app.run(port=8000)
+
+    thread = threading.Thread(target=run_app, daemon=True)
+    thread.start()
+    time.sleep(1)
 
 
 @pytest.fixture(scope='module')
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,8 +4,6 @@ import os.path
 import shutil
 import pytest
 import py.path
-import time
-import threading
 import unittest.mock
 import annif
 import annif.analyzer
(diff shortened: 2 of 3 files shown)


osma:
Great work!

These are indeed already somewhat overlapping with the tests in `test_rest.py`. IIRC originally the idea was to separate the test files as follows:

* `test_swagger.py` does some basic exercise of the methods defined in the OpenAPI spec (as is done automatically by swagger-tester) using actual HTTP requests; as well as checking for CORS headers (which also requires HTTP)
* `test_rest.py` does more detailed tests (checking that the methods work as intended, including things like access control) only on a Python code level, without actual HTTP requests.

I think it would be useful to keep this kind of separation in mind even though we abandon swagger-tester - the tests in `test_swagger.py` should be kept in sync with the OpenAPI spec file and operate on a HTTP level, while `test_rest.py` can go a bit deeper into the application logic using Python only. On a quick reading, this seems to be the case in the tests you've written in this PR.

osma:
Thinking aloud here: What this PR is missing is that it doesn't check that the returned data conforms to the shapes (return data types) specified in the OpenAPI spec file. I think swagger-tester did that, though I'm not 100% sure. There's a greater risk now that the app will not actually honor the contracts in the spec file.

The tests could be extended so that they perform more detailed testing of the returned data, but that still wouldn't guarantee that it follows the spec file, just that the person who wrote the tests thinks that the tests are checking for the right things. It would be better to derive the tests automatically from the spec file, not to craft them manually. But unfortunately we cannot use swagger-tester anymore for this.

osma (approved):
lgtm

juhoinkinen:
For the future, I checked some alternatives for swagger-tester:
- [swagger-unittest](https://github.com/mild-blue/swagger-unittest) is a fork from swagger-tester, but cannot handle 204 responses from learn-method (and has no license).
- [flex](https://github.com/pipermerriam/flex/) seems suitable, but it would require writing [manually a test for each API method](https://flex-swagger.readthedocs.io/en/latest/#api-call-validation) and has not been updated in 2 years. However, flex noticed null to be a wrong type of response for `modification_time`.
- [Schemathesis](https://github.com/schemathesis/schemathesis) is a large testing tool, seems a bit too big gun for Annif API at the moment. Also there was some package conflicts when trying to install.
- [swagger-conformance](https://github.com/olipratt/swagger-conformance) is unmaintained.

Stranger6667:
Hi @juhoinkinen 

Schemathesis author here :)

> Schemathesis is a large testing tool, seems a bit too big gun for Annif API at the moment. Also there was some package conflicts when trying to install.

I'd be happy to review the dependencies on the Schemathesis side so it is easier to install :) Let me know if there is anything I can help with on this matter

juhoinkinen:
Hi @Stranger6667 

I took another look at this, and on current master Schemathesis actually gets installed, but by backtracking Schemathesis to version 0.7.3. 

Trying to install the latest Schemathesis v3.12.3 this conflict message is printed out:

```
ERROR: Cannot install annif and connexion2[swagger-ui]==2.10.0 because these package versions have conflicting dependencies.

The conflict is caused by:
optuna 2.10.0 depends on PyYAML
schemathesis 3.12.3 depends on pyyaml<7.0 and >=6.0
connexion2[swagger-ui] 2.10.0 depends on PyYAML<6 and >=5.1
```

In Connexion the pyyaml version has been limited to <6 [last June](https://github.com/zalando/connexion/pull/1367).

However, because Annif is not using any Click 8 feature, at the moment the easiest way to resolve this for us would be to allow older Click, e.g. >= 7, which would allow Schemathesis older than [3.11.1](https://github.com/schemathesis/schemathesis/releases/tag/v3.11.1). But as we have just updated to Click 8, which required some work to avoid other conflicts (#533), it's best just wait for Connexion to catch up.  

But maybe some other people could be facing the same conflict, so allowing older Click in Schemathesis could be considered, if Click 8 is not needed there.

Stranger6667:
@juhoinkinen 

Thank you so much for providing the context! Indeed, there is no hard dependency on Click, so we can relax it a bit - I'll take a look at it soon :)

PR #553, opened 2022-01-18
Update Scrutinizer config

juhoinkinen:
I noticed Scrutinizer was still running on Python 3.6 and installing code with pipenv, although those are not anymore in use with Annif (since #550 and #405). So in Scrutinizer runs this changes 

- Python version to 3.8.12
- Use directly pip instead of pipenv

(Draft PR first to see if this works.)


juhoinkinen:
Works, so I'll merge.

PR #554, opened 2022-01-18
Make LMDB map size a configurable parameter for `nn_ensemble` backend.

### LINKED ISSUE #552: LMDB can overflow
When training a `nn_ensemble` project with a large number of documents, the LMDB will throw an error message regarding size
```
lmdb.MapFullError: mdb_put: MDB_MAP_FULL: Environment mapsize limit reached
```
I propose three different solutions to get around this issue:

* Make the size configurable using an environment variable.
Probably the easiest solution. Users still may run into the problem when the size was not set correctly.
* When the DB is full close end the transaction and double the size of the DB.
Would need some refactoring regarding the handling of transactions.
Probably leading to more complicated code, as the transaction handling is on a different level of the call hierarchy than the actual writing.
* Use the data set functionality of tensor flow:
[TFRecordWriter](https://www.tensorflow.org/api_docs/python/tf/io/TFRecordWriter?hl=en) and [TFRecordDataset](https://www.tensorflow.org/api_docs/python/tf/data/TFRecordDataset?hl=en)
This would introduce a new technology for handling input data of nn_ensemble. Required effort is probably somewhere between the first to proposals. It also seems like you want to use LMDB more in the future (See #378 ) So this may not align with your roadmap.

What do you think about the options? I could work on this.

mo-fu:
This PR adds the simple solution to close #552, as discussed in the issue.

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -91,7 +91,6 @@ class NNEnsembleBackend(
 
     MODEL_FILE = "nn-model.h5"
     LMDB_FILE = 'nn-train.mdb'
-    LMDB_MAP_SIZE = 1024 * 1024 * 1024
 
     DEFAULT_PARAMETERS = {
         'nodes': 100,
@@ -99,6 +98,7 @@ class NNEnsembleBackend(
         'optimizer': 'adam',
         'epochs': 10,
         'learn-epochs': 1,
+        'lmdb_map_size': 1024 * 1024 * 1024
     }
 
     # defaults for uninitialized instances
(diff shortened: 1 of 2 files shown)


osma:
Great! Can we say this fixes #552 (to auto-close the issue when this is merged)? I think it has to be stated in the opening post to be effective.

osma (approved):
LGTM

osma:
Merged, thanks a lot @mo-fu !

osma:
As we are nearing the 0.56 release, I added some documentation about this setting to the [wiki page of the NN ensemble backend](https://github.com/NatLibFi/Annif/wiki/Backend:-nn_ensemble) - see [this diff](https://github.com/NatLibFi/Annif/wiki/Backend:-nn_ensemble/_compare/d20af08f9ce599d67321fab4739ac9b7853890dc...dcab6f8136953e91a3ea29010acac5e7025607e7) for the changes.

PR #557, opened 2022-01-27
Output eval metrics as a JSON file compatible with DVC

### LINKED ISSUE #546: Eval metrics file output compatible with DVC
To make it possible to use Annif in [DVC](https://dvc.org/) workflows, Annif should be able to produce a metrics file in a format that DVC understands, i.e. either JSON or YAML 1.2 (see documentation for [dvc metrics](https://dvc.org/doc/command-reference/metrics)).

Example command:

    annif eval my-project --metrics-file metrics.json /path/to/my-test-corpus/

and the created `metrics.json` file would look something like this:

```
{
  "Precision (doc avg)": 0.1560,
  "Recall (doc avg)": 0.2333,
  ...
  "F1@5": 0.1789,
  ...
  "Documents evaluated": 300
}
```

The option could be named `-M` or `--metrics-file`.

It's possible that the selection of metrics should be reduced - the default set is quite large and this may be cumbersome to deal with in DVC. It would be helpful to implement also an option to select which metrics to calculate (see #545).

Also, DVC may not like the current naming of the metrics which includes capital letters, spaces and parentheses (e.g. `F1 score (doc avg)`) and perhaps it would be better to use names like `f1_score_doc_avg`.

osma:
This PR adds an option `-M`/`--metrics-file` to the `annif eval` command, like this:

    annif eval my-project --metrics-file metrics.json /path/to/my-test-corpus/

The command writes evaluation metrics into the file in a format like this:

```json
{
  "Precision (doc avg)": 0.12429731417863835,
  "Recall (doc avg)": 0.10921187650575533,
  "F1 score (doc avg)": 0.11149022571570978,
  "Precision (subj avg)": 0.841174914733954,
  "Recall (subj avg)": 0.14326083068180426,
  "F1 score (subj avg)": 0.21448794969174276,
  "Precision (weighted subj avg)": 0.736846816147547,
  "Recall (weighted subj avg)": 0.09503311258278145,
  "F1 score (weighted subj avg)": 0.1529771375758776,
  "Precision (microavg)": 0.6067653276955602,
  "Recall (microavg)": 0.09503311258278145,
  "F1 score (microavg)": 0.16432865731462926,
  "F1@5": 0.11149022571570978,
  "NDCG": 0.1130682392238403,
  "NDCG@5": 0.11315954263925752,
  "NDCG@10": 0.1130682392238403,
  "Precision@1": 0.127420362273579,
  "Precision@3": 0.12429731417863835,
  "Precision@5": 0.12429731417863835,
  "LRAP": 0.11837633968849189,
  "True positives": 861,
  "False positives": 558,
  "False negatives": 8199,
  "Documents evaluated": 6404
}
```

Fixes #546

TODO: Still need to test whether DVC is happy with this format.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -19,6 +19,7 @@ import annif.registry
 from annif.project import Access
 from annif.suggestion import SuggestionFilter, ListSuggestionResult
 from annif.exception import ConfigurationException, NotSupportedException
+from annif.util import metric_code
 
 logger = annif.logger
 click_log.basic_config(logger)
@@ -389,7 +390,9 @@ def run_eval(
     for metric, score in metrics.items():
         click.echo(template.format(metric + ":", score))
     if metrics_file:
-        json.dump(metrics, metrics_file, indent=2)
+        json.dump(
+            {metric_code(metric): val for metric, val in metrics.items()},
+            metrics_file, indent=2)
 
 
 @cli.command('optimize')
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -6,6 +6,7 @@ import collections
 import os.path
 import re
 import sys
+import json
 import click
 import click_log
 from flask import current_app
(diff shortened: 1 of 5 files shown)


osma:
Seems to be working with DVC. I changed the metric names in the JSON to be more computer-friendly, e.g. `Precision_doc_avg` instead of `Precision (doc avg)`, as anticipated in #546, although this probably wouldn't be strictly necessary.

PR #560, opened 2022-01-28
Support TOML as a configuration file format alongside CFG/INI.

### LINKED ISSUE #547: Configuration file format compatible with DVC
To make it possible to use Annif productively in a [DVC](https://dvc.org/) workflow, it would be helpful if DVC tools could read parameters directly from the Annif configuration file. DVC can currently read parameters from YAML 1.2, JSON, TOML and Python files (see documentation for [dvc params](https://dvc.org/doc/command-reference/params)).

Annif uses INI-style syntax (supported by [configparser](https://docs.python.org/3/library/configparser.html) Python standard library module) in the `projects.cfg` configuration file. This is similar to TOML, but not identical.

I can think of at least these options for making the Annif configuration file DVC-compatible:

1. Support YAML configuration files as an alternative to the current format.
2. Support JSON configuration files as an alternative to the current format.
3. Support TOML configuration files as an alternative to the current format.
4. Adjust the current format slightly so that it becomes a valid subset of TOML.

I think we can rule out 2., because JSON is not very nice as a configuration language because of its strict syntax and lack of support for comments. If we want a new configuration format, either YAML (option 1) or TOML (option 3) is better.

For 4., AFAICT the main difference between the current syntax and TOML is that TOML requires string values to be quoted. So instead of this:

```
[tfidf-en]
language=en
backend=tfidf
analyzer=snowball(english)
limit=100
vocab=yso-en
```

the syntax must be

```
[tfidf-en]
language="en"
backend="tfidf"
analyzer="snowball(english)"
limit=100
vocab="yso-en"
```

(note that `limit` can be left as-is, as the value 100 is an integer, not a string)

This syntax doesn't work with ConfigParser currently, because it includes the quotes as part of the value. But it would be simple to change the `optionxform` so that it drops any quotes from the value. The file name `projects.cfg` could still be a problem for DVC, which would probably expect the extension `.toml` so that it recognizes which syntax to use.

osma:
Using Annif efficiently with DVC requires that DVC understands the Annif configuration file; the current INI-style format is not supported by DVC (see #547).

This PR adds support for TOML as a configuration file format alongside the current one (which still works). Unless the path to the configuration file has been set using the `ANNIF_PROJECTS` environment variable or the `-p` command line option, the configuration will be read from `projects.cfg` if it exists, otherwise from `projects.toml`.

The main difference in the configuration syntax is that TOML requires string values to be quoted, e.g.

    language="fi"

instead of

    language=fi

The [tomli](https://github.com/hukkin/tomli) parser is used for parsing project configuration files with a `.toml` extension. It's a very minimal read-only, pure Python TOML parser; in the future, similar functionality (based on tomli) may be included in the Python standard library - see [PEP 680](https://www.python.org/dev/peps/pep-0680/).

Fixes #547

Opening as draft PR to get early feedback from CI tools. Probably some refactoring will be necessary, as well as additional unit tests.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -110,7 +110,7 @@ def set_project_config_file_path(ctx, param, value):
 def common_options(f):
     """Decorator to add common options for all CLI commands"""
     f = click.option(
-        '-p', '--projects', help='Set path to projects.cfg',
+        '-p', '--projects', help='Set path to project configuration file',
         type=click.Path(dir_okay=False, exists=True),
         callback=set_project_config_file_path, expose_value=False,
         is_eager=True)(f)
--- a/annif/config.py
+++ b/annif/config.py
@@ -45,7 +45,8 @@ class AnnifConfigTOML:
                     f"Reading configuration file {filename} in TOML format")
                 self._config = tomli.load(projf)
             except tomli.TOMLDecodeError as err:
-                raise ConfigurationException(err)
+                raise ConfigurationException(
+                    f"Parsing TOML file '{filename}' failed: {err}")
 
     @property
     def project_ids(self):
--- a/annif/config.py
+++ b/annif/config.py
@@ -55,7 +55,7 @@ class AnnifConfigTOML:
         return self._config[key]
 
 
-def parse_config(projects_file):
+def find_config(projects_file):
     if projects_file:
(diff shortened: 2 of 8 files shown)


juhoinkinen (approved):
LGTM

PR #561, opened 2022-01-28
Set a 10 minute timeout for GitHub Actions CI jobs

osma:
Recently, many GitHub Actions CI jobs have been stalling. I don't know the root cause, but they are timing out after 6 hours.
This is clearly too long, since normally the jobs should take less than 5 minutes.

This PR sets a 10 minute timeout for CI jobs to avoid them being stuck for hours and hopefully making it easier to investigate the stalled runs.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -8,6 +8,7 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


osma:
Ironically, the 3.9 pull_request CI job got stuck. But it timed out after 10 minutes instead of 6 hours, so the setting is working!

PR #562, opened 2022-01-31
Add missing import of annif.eval in MLLM backend

osma:
The MLLM backend was missing an import of annif.eval. This didn't always cause issues because usually it was imported by other modules such as annif.cli. It may have caused tests to occasionally hang on GitHub Actions CI.

This PR simply adds the missing import.


osma:
Unfortunately this didn't fix the hanging tests, but it's still a problem worth fixing, so I'm merging this anyway.

PR #563, opened 2022-01-31
Lazy import of spaCy to avoid startup overhead

osma:
This PR moves the import of spaCy inside the SpacyAnalyzer class so that it will be imported only when actually used. It reduces the startup overhead of Annif by around 2 seconds when the optional spacy feature has been installed.


osma:
There's a problem with this approach - the spaCy tests will run even when spacy isn't installed, causing them to fail. It probably has to do with the `importorskip` directive being ineffective now that spacy isn't always imported.

PR #568, opened 2022-02-04
Perform suggest operations in parallel using multiprocessing in nn_ensemble

### LINKED ISSUE #429: Parallelize suggest operations during nn_ensemble training
Both the `eval` and `hyperopt` commands can now perform `suggest` operations in parallel on multiple CPUs. When training the nn_ensemble, running the documents through the source projects is the main bottleneck. We could speed it up by implementing parallel suggest operations there as well. There should probably be a new `jobs` parameter for the `train` CLI command specifying the number of parallel jobs, just like for `eval` and `hyperopt`.

osma:
During training, the NN ensemble has to process the training documents through the source projects, collecting suggestions. Previously this was done sequentially and could take a lot of time, often dominating the time it takes to train a NN ensemble.

With this PR, the suggest operations are done in parallel on multiple CPUs, controlled by the `--jobs` parameter of the `annif train` command. The default is to use all CPUs.

Fixes #429

Opening a draft PR to get feedback from QA tools. More detailed testing and benchmarking still needs to be done.

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -180,23 +180,26 @@ class NNEnsembleBackend(
 
     def _corpus_to_vectors(self, corpus, seq, n_jobs):
         # pass corpus through all source projects
-        project_weights = dict(
+        sources = dict(
             annif.util.parse_sources(self.params['sources']))
 
         # initialize the source projects before forking, to save memory
-        for project_id in project_weights.keys():
+        self.info(
+            f"Initializing source projects: {', '.join(sources.keys())}")
+        for project_id in sources.keys():
             project = self.project.registry.get_project(project_id)
             project.initialize(parallel=True)
 
         psmap = annif.parallel.ProjectSuggestMap(
             self.project.registry,
-            list(project_weights.keys()),
+            list(sources.keys()),
             backend_params=None,
             limit=None,
             threshold=0.0)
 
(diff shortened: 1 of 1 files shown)


osma:
I did some benchmarks using the `yso-nlf` data set from the Annif tutorial. I set up TFIDF, MLLM and Omikuji projects according to the tutorial exercises, then created a NN ensemble project including all three as sources, like this:

    sources=yso-tfidf-en,yso-mllm-en:2,yso-omikuji-parabel-en

I trained the NN ensemble using 500 documents (--docs-limit 500) with different setups on an i7-8550U laptop (4 cores, 8 threads). In the end I also evaluated the ensemble project against the test set, to verify that the ensemble is still working. Results:

|                 | wall time | user time | max RSS | F1@5                |
| --------------- | --------- | --------- | ------- | ------------------- |
| Before PR       | 22:46  | 1448      | 3021780 | 0.3632 |
| This PR, jobs=8 | 10:31  | 4846      | 2643096 | 0.3666 |
| This PR, jobs=1 | 22:31  | 1442      | 3312252 | 0.3727 |

Observations:

- Training time decreases significantly (to less than half) when using parallel processing, although there is significant overhead in the parallelization - the total CPU time (user time) increases a lot.
- With jobs=1, the situation is similar to what it was before.
- Memory usage does not change dramatically.
- Evaluation scores are similar to what they were before.

juhoinkinen (approved):
Great improvement!

PR #571, opened 2022-02-15
Docker builds on GitHub Actions

juhoinkinen:
This switches to use GH Actions instead of Drone to build and publish Docker images.

The functionality should remain as previously: a Docker image is build and published to quay.io
- with the `latest` image tag for every push to master branch
- with [image tags based on the pushed git tag](https://github.com/marketplace/actions/docker-metadata-action#semver) (e.g. `v1.2.3` git tag would give `1.2.3`, `1.2`, and `latest` image tags)

I leave this as a draft, because the necessary secrets for publishing to Quay.io are not in place yet (TODO, I don't have them).

~For the record, I tried to speed the runs by [using pip cache](https://github.com/NatLibFi/Annif/commit/244190ca0499a1d4b55bedbcfbbc05c4480ff0ed), but it actually [always slowed the runs on Python 3.8](https://github.com/NatLibFi/Annif/actions/runs/1841923791).~ Edit: pip cache is now enabled by #576, after more precise timing of runs.

I also did not enable caching of Docker layers, which would speed builds, because I realized that the cache is not invalidated when (unpinned) dependencies are updated in PyPI or apt servers, leading to Docker image possibly having different dependency versions than those that unit tests are using.

Final note: also upgrades `setuptools`, not only `pip`, for the unit tests step.


juhoinkinen:
Rebased and force-pushed to resolve conflicts and drop unrelated commits (pip cache experiments).

juhoinkinen:
Well, now due to the file rename ([.github/workflows/python-package.yml → .github/workflows/cicd.yml](https://github.com/NatLibFi/Annif/pull/571/files#diff-6727e33ccc9195d67f0786e1384f8f1cdaf4090c3e77547943105bd2b28c99d0)) on the Actions tab under All workflows there is twice the "CI/CD" workflow... The old one could be disabled when this is merged, or maybe it's better just to revert the rename...

juhoinkinen:
Rebased and force-pushed to resolve conflicts and drop unnecessary commits made for testing. 

Now the Quay.io secrets are available as GitHub organization secrets, so this is ready.

osma (approved):
LGTM, glad to get rid of Drone which was always a mystery to me :)

PR #572, opened 2022-02-16
Update Dockerfile v0.57

juhoinkinen:
- Install rsync to allow transferring models to OpenShift environment
- Ensure pip & setuptools are updated when building
- Install voikko system packages only if Voikko is included in optional dependencies
- Remove retry loop for nltk data download (hopefully not needed anymore, as the network problems encountered in Drone should not exist in GH Actions)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,10 +27,10 @@ RUN apt-get update && \
 		apt-get install -y --no-install-recommends \
 			libvoikko1 \
 			voikko-fi; \
-	fi \
+	fi && \
 	# curl for Docker healthcheck and rsync for model transfers:
-	&& apt-get install -y --no-install-recommends curl rsync \
-	&& rm -rf /var/lib/apt/lists/* /usr/include/*
+	apt-get install -y --no-install-recommends curl rsync && \
+	rm -rf /var/lib/apt/lists/* /usr/include/*
 
 WORKDIR /Annif
 RUN pip install --upgrade pip setuptools --no-cache-dir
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN if [[ $optional_dependencies =~ "fasttext" ]]; then \
 		apt-get update && \
 		apt-get install -y --no-install-recommends \
 			build-essential && \
+		pip install --upgrade pip setuptools --no-cache-dir && \
```
osma, line 12: Would it make sense to install `wheel` here as well? In my understanding, it might speed up installation of some Python packages - they can be installed from prebuilt wheels instead of rebuilt. Not sure if that is relevant for the Annif dependencies though.

Dockerfile
```diff
@@ -19,26 +20,27 @@ FROM python:3.8-slim-bullseye
...
+			libvoikko1 \
+			voikko-fi; \
+	fi && \
+	# curl for Docker healthcheck and rsync for model transfers:
+	apt-get install -y --no-install-recommends curl rsync && \
+	rm -rf /var/lib/apt/lists/* /usr/include/*
 
 WORKDIR /Annif
-RUN pip install --upgrade pip --no-cache-dir
+RUN pip install --upgrade pip setuptools --no-cache-dir
```
osma, line 36: Or maybe this would be a better place for `wheel`...

Dockerfile
juhoinkinen: Yes it would. When installing deps for unit test runs in GH Action, there are messages like
`Using legacy 'setup.py install' for xxx, since package 'wheel' is not installed.`
I was confused as I did not see these messages in Docker builds, but that was just because `wheel` actually was available as it got installed as a dependency (of TensorFlow?). 

(I'll install wheel also for unit test runs.)

Dockerfile
juhoinkinen: And here too.

PR #578, opened 2022-03-10
Update dependencies v0.57

juhoinkinen:
Dependency updates for Annif v0.57.

Updates to:

- Omikuji 0.5.x to [enable symlinking Omikuji datafiles](https://github.com/tomtung/omikuji/issues/38)
- TensorFlow 2.7.1 to [fix many vulnerabilities](https://github.com/tensorflow/tensorflow/releases/tag/v2.7.1)
- Connexion to again use its [original/main PyPI project](https://pypi.org/project/connexion/) instead of the [Connexion2 project](https://pypi.org/project/connexion2/) (which now seems to be abandoned).

I don't think there are other dependencies that can be upgraded (to keep Python 3.6-3.9 support) or are worth upgrading now, or are there some?

Still needs to check that old Omikuji and NN ensemble models keep working, but I would suppose so.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -45,14 +45,14 @@ jobs:
         # Install the optional fastText dependencies for Python 3.8 only
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then pip install .[fasttext]; fi
         # Install the optional spaCy dependencies for Python 3.8 only
-        # and download the small English pretrained spaCy model needed by spacy analyzer
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then
-        pip install \
-          .[spacy] https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.2.0/en_core_web_sm-3.2.0-py3-none-any.whl
+          pip install .[spacy]
+          # download the small English pretrained spaCy model needed by spacy analyzer
+          python -m spacy download en_core_web_sm
         fi
         # For Python 3.7
         # - voikko and pycld3 dependencies
-        if [[ ${{ matrix.python-version }} == '3.7' ]]; then pip install .[voikko,pycld3]; fi
+        if [[ ${{ matrix.python-version }} == '3.7' ]]; then python -m pip install .[voikko,pycld3]; fi
 
     - name: Lint with flake8
       run: |
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -48,7 +48,7 @@ jobs:
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then
           pip install .[spacy]
           # download the small English pretrained spaCy model needed by spacy analyzer
-          python -m spacy download en_core_web_sm
+          python -m spacy download en_core_web_sm --upgrade-strategy only-if-needed
         fi
         # For Python 3.7
         # - voikko and pycld3 dependencies
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -45,10 +45,10 @@ jobs:
         # Install the optional fastText dependencies for Python 3.8 only
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then pip install .[fasttext]; fi
         # Install the optional spaCy dependencies for Python 3.8 only
+        # and download the small English pretrained spaCy model needed by spacy analyzer
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then
-        pip install .[spacy]
-        # download the small English pretrained spaCy model needed by spacy analyzer
-        pip install https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.2.0/en_core_web_sm-3.2.0-py3-none-any.whl
+        pip install \
+          .[spacy] https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.2.0/en_core_web_sm-3.2.0-py3-none-any.whl
         fi
         # For Python 3.7
         # - voikko and pycld3 dependencies
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -52,7 +52,7 @@ jobs:
         fi
         # For Python 3.7
         # - voikko and pycld3 dependencies
-        if [[ ${{ matrix.python-version }} == '3.7' ]]; then python -m pip install .[voikko,pycld3]; fi
+        if [[ ${{ matrix.python-version }} == '3.7' ]]; then pip install .[voikko,pycld3]; fi
 
     - name: Lint with flake8
       run: |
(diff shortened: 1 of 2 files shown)


juhoinkinen:
I checked that Omikuji and NN ensemble models trained on the previous versions (of Annif 0.56) keep working.

osma:
Now that Connexion development seems to have picked up again (great!), would it make sense to pin the version we use, to avoid a future update of Connexion breaking something?

juhoinkinen:
> Now that Connexion development seems to have picked up again (great!), would it make sense to pin the version we use, to avoid a future update of Connexion breaking something?

Yes, could be so. I'll pin Connexion to 2.12.x.

osma:
There are plans for a new major version (presumably 3.0) of Connexion, which would drop support for Flask 1.x, according to the discussion in this recent PR: https://github.com/spec-first/connexion/pull/1465
That could break things for us, although I'm not sure.
So I suggest we should pin Connexion to 2.12.* or 2.*

juhoinkinen:
Hmh, there are pip errors in the logs of the install step that do not make the step fail.

[On Python 3.8 after spacy install](https://github.com/NatLibFi/Annif/runs/5508418566?check_suite_focus=true#step:5:451):
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
annif 0.57.0.dev0 requires numpy==1.21.*, but you have numpy 1.22.3 which is incompatible.
Successfully installed en-core-web-sm-3.2.0 numpy-1.22.3
```
This is caused due to the [eager upgrade-strategy](https://github.com/NatLibFi/Annif/blob/master/.github/workflows/cicd.yml#L10) (checked on my laptop), which I added to make sure old dependencies are not used from pip cache. Better to remove the eager strategy I think.

[On Python 3.7](https://github.com/NatLibFi/Annif/runs/5508418529?check_suite_focus=true#step:5:374):
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
flake8 4.0.1 requires importlib-metadata<4.3; python_version < "3.8", but you have importlib-metadata 4.11.2 which is incompatible.
```

Probably also because of the eager strategy...

osma:
> This is caused due to the [eager upgrade-strategy](https://github.com/NatLibFi/Annif/blob/master/.github/workflows/cicd.yml#L10) (checked on my laptop), which I added to make sure old dependencies are not used from pip cache. Better to remove the eager strategy I think.

The way I read it, it seems that when spacy is installed, numpy 1.21.x is also for some reason upgraded to 1.22.x and then pip complains about not matching numpy 1.21.x which Annif depends on.

It's not obvious to me why pip decides to upgrade numpy at the time of installing spacy. Spacy only requires numpy>=1.15.0: https://github.com/explosion/spaCy/blob/297dd82c86372c7aa0a181e55dc72512718aafe8/requirements.txt#L16

In particular the lines 428-430 in the Python 3.8 test output you linked to are puzzling:

```
Requirement already satisfied: numpy>=1.15.0 in /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages (from spacy<3.3.0,>=3.2.0->en-core-web-sm==3.2.0) (1.21.5)
Collecting numpy>=1.15.0
  Using cached numpy-1.22.3-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.8 MB)
```

Maybe this has to do with the eager strategy you mentioned? Would it help to adjust the `pip install` command that installs spacy somehow, so it won't install numpy 1.22.x?

juhoinkinen:
These conflicts do not fail the GH Actions install step because the [exit code of `pip install` is zero even for conflicting dependencies](https://github.com/pypa/pip/issues/8104), but (as mentioned in the thread) after installation [`pip check`](https://pip.pypa.io/en/stable/cli/pip_check/) could be used to see if there are conflicts and error out if there are (tested [here](https://github.com/NatLibFi/Annif/runs/5509570264?check_suite_focus=true#step:5:457) for the current case). 

Should have the `pip check`, it would make the conflicts detectable by the CI pipeline, instead of showing them on the logs only? (Could be another PR.)

juhoinkinen:
> It's not obvious to me why pip decides to upgrade numpy at the time of installing spacy. Spacy only requires numpy>=1.15.0: https://github.com/explosion/spaCy/blob/297dd82c86372c7aa0a181e55dc72512718aafe8/requirements.txt#L16
> 
> In particular the lines 428-430 in the Python 3.8 test output you linked to are puzzling:
> 
> ```
> Requirement already satisfied: numpy>=1.15.0 in /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages (from spacy<3.3.0,>=3.2.0->en-core-web-sm==3.2.0) (1.21.5)
> Collecting numpy>=1.15.0
>   Using cached numpy-1.22.3-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.8 MB)
> ```
> 
> Maybe this has to do with the eager strategy you mentioned? Would it help to adjust the `pip install` command that installs spacy somehow, so it won't install numpy 1.22.x?

Actually the conflict arised when installing the spacy en_core_web_sm model. Seems that the model depends on numpy, and with the eager upgrade strategy, when the spacy downloader (or direct pip call) installed only the model, also numpy got upgraded (makes sense actually). But now, when the model and spacy itself as Annif's extra module are both installed in one `pip install` call, pip knows that also Annif's depedencies should be satisfied.

Now, as the spacy model is installed with pip instead of the spacy downloader, the version of the model needs to be fixed. I don't know whether that's good or bad.


> [On Python 3.7](https://github.com/NatLibFi/Annif/runs/5508418529?check_suite_focus=true#step:5:374):
> 
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> flake8 4.0.1 requires importlib-metadata<4.3; python_version < "3.8", but you have importlib-metadata 4.11.2 which is incompatible.

This other conflict has actually been around from January and is not related to eager strategy.

osma:
> Now, as the spacy model is installed with pip instead of the spacy downloader, the version of the model needs to be fixed. I don't know whether that's good or bad.

That's a bit unfortunate. I think it would be better to use the `spacy download` command, as it's the official way of installing spaCy models and also what we recommend in Annif documentation.

Would it be possible to disable the eager strategy just for the `spacy download` command?

juhoinkinen:
> > Now, as the spacy model is installed with pip instead of the spacy downloader, the version of the model needs to be fixed. I don't know whether that's good or bad.
> 
> That's a bit unfortunate. I think it would be better to use the `spacy download` command, as it's the official way of installing spaCy models and also what we recommend in Annif documentation.
> 
> Would it be possible to disable the eager strategy just for the `spacy download` command?

Yes, the `--upgrade-strategy only-if-needed` option worked for `spacy download` as for `pip install`.

PR #580, opened 2022-03-14
Make importlib_metadata regular dependency

juhoinkinen:
When some DeprecationWarnings of unit tests were resolved in #575, I [put `importlib_metadata` installation under dev dependencies](bfb7a1a6a99ab070eef56cad87151301c7b77cf3), but it should be a regular dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,8 @@ setup(
         'optuna==2.10.*',
         'stwfsapy==0.3.*',
         'python-dateutil',
-        'tomli==2.0.*'
+        'tomli==2.0.*',
+        'importlib_metadata'
     ],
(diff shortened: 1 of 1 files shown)


osma:
Can you clarify why it should be a regular dependency?

juhoinkinen:
Uups, now when I checked why it should be a regular dependency, I see no reason. 

I thought that `importlib_metadata` was used in the `annif --version` CLI option, but it is not; it is used only in [testing it](https://github.com/NatLibFi/Annif/blob/3629139f1017399dd7da844374da3c584064b1f9/tests/test_cli.py#L809), so it should remain as a dev dependency. 

I'll close this, thanks for questioning :)

PR #585, opened 2022-03-31
Support for multiple configuration files in a directory

### LINKED ISSUE #584: Support for multiple configuration files in a directory
Currently Annif only supports only a single configuration file, usually named `projects.cfg` (or `projects.toml` after #560 added support for the TOML format). This is sometimes a limitation, because there can be many independent projects (or groups of projects), but they all have to be in the same file. For example, if projects for different vocabularies are developed independently (perhaps in separate DVC repositories), it would ease their deployment if their configurations could be kept in separate files.

The proposal is to add support for a configuration directory. All the configuration files in that directory (filenames matching the patterns `*.cfg` or `*.toml`) would be read and their contents merged. This is similar to e.g. how the `sites-available` directory works in typical Apache HTTPD installations.

Annif already looks for files in the current working directory called `projects.cfg` and `projects.toml`, unless the path to the configuration file has been set using the `-p` CLI parameter or the `ANNIF_PROJECTS` environment variable. The search could be extended to look for a directory with a specific name. Of course, it should also be possible to use `-p` or `ANNIF_PROJECTS` to point at a specific directory.

The default name of the directory could be e.g. `config`, `conf`, `conf.d`, `projects.d` or something along those lines. (opinions?)

juhoinkinen:
Adds support for a project configuration directory. Closes #584.

Checks for duplicated project IDs across all configuration files in the configuration directory, and raises `ConfigurationException` if any are found.

For now I chose `projects.d` as the name for the default configuration directory, because it is nicely in line with the `projects.cfg` and `projects.toml` names for the default configuration files. Downside is that this name can be confusing: one could use "projects directory" to refer to the (projects) configuration directory (`projects.d`) or the projects data directory (`data/projects`). Or even to the parent directory containing `projects.cfg` and `data/` (in Docker instances I have been using `/annif-projects` directory for this).

I open PR as draft now; this probably needs some more tests etc.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -104,14 +104,15 @@ def set_project_config_file_path(ctx, param, value):
     """Override the default path or the path given in env by CLI option"""
     with ctx.ensure_object(ScriptInfo).load_app().app_context():
         if value:
-            current_app.config['PROJECTS_FILE'] = value
+            current_app.config['PROJECTS_CONFIG_PATH'] = value
 
 
 def common_options(f):
     """Decorator to add common options for all CLI commands"""
     f = click.option(
-        '-p', '--projects', help='Set path to project configuration file',
-        type=click.Path(dir_okay=False, exists=True),
+        '-p', '--projects',
+        help='Set path to project configuration file or directory',
+        type=click.Path(dir_okay=True, exists=True),
         callback=set_project_config_file_path, expose_value=False,
         is_eager=True)(f)
     return click_log.simple_verbosity_option(logger)(f)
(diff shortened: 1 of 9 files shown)


juhoinkinen:
One reason to use the `projects.d` name instead of `conf`, `config` etc. is that the latter names make to think about some general configuration not necessarily about projects (like log file location, parallelization settings... although those are not existing in Annif now, maybe some day some like those could be).

I don't see anything to add to this now, @osma can you take a look?

osma (approved):
Code looks good!
I tested this briefly and couldn't make it not work ;)
Great work!

osma:
I edited the opening post slightly ("closing #584" -> "closes #584") so that GitHub recognizes the relationship and will automatically close the issue when this PR is merged.

PR #587, opened 2022-04-14
Update GitHub Actions

juhoinkinen:
Upgrades GH Actions to latest versions, and pins the codecov and pypi-publish Actions commit hashes.

Upgrading codecov Action (from v1 to v3) necessitates generating the coverage report in xml format ([apparently this applies to both v2 and v3](https://github.com/codecov/codecov-action/issues/577)).

~Also makes the coverage report upload conditional: it is uploaded only by Python 3.8 test run. Until now, the report was uploaded three times (by Python 3.7, 3.8 and 3.9 runs). That may be the reason for weird coverage changes I have noticed (the coverage is different for different Python versions, and it is somewhat random which run finishes last, and I think the report by the last run overrides earlier ones).~

~Now I chose to upload the report by Python 3.8 run. This evens out the the running times (very slightly). This means the test coverage of NN ensemble, Omikuji and Yake backends is not tracked, while of fastText and spaCy code it is.~

Edit: Actually [codecov merges the reports that are uploaded by different runs](https://docs.codecov.com/docs/merging-reports), so it is desired to upload all of them as previously.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -64,7 +64,7 @@ jobs:
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: Test with pytest
       run: |
-        pytest --cov=./
+        pytest --cov=./ --cov-report xml
     - name: Upload coverage to Codecov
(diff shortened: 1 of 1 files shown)


osma (approved):
LGTM!

PR #592, opened 2022-05-30
Add Python 3.10 support & update dependencies

### LINKED ISSUE #589: Support Python 3.10
Ubuntu 22.04, which has been just released, comes with Python 3.10 by default. Annif should fully support Python 3.10.

[As reported in annif-users](https://groups.google.com/g/annif-users/c/suuql5d1ufs/m/-MmETFfrAAAJ), installing Omikuji and NN backends fail, and also installing Gensim and pycld3 dependencies have some issues:

## Omikuji
Current Omikuji does not have a wheel for Python 3.10, so installing Omikuji for Python 3.10 requires installing first cargo, but even after that, there is an error (at least on Docker image python:3.10-slim):
```
error: failed to parse manifest at `/tmp/pip-install-5yzjqez2/omikuji_cbdca1e9c9fb435e9d32fc4d65a5d33f/c-api/Cargo.toml`
Caused by:
     failed to parse the `edition` key
Caused by:
     this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
 ```
I'm not sure if this can be handled some way or should it be fixed in Omikuji.

## NN ensemble
[Python 3.10 support requires TensorFlow 2.8 or later.](https://www.tensorflow.org/install/pip) Apparently due to NumPy version requirements.

## Gensim
Current Gensim releases have [wheels only for Python 3.6-3.9](https://pypi.org/project/gensim/#files), but the [next release will have wheel for Python 3.10 too](https://github.com/RaRe-Technologies/gensim/pull/3298). Before next release, for installing Annif on Python 3.10 it is required to have gcc etc. from build-essential package (which is the case for fastText anyway).

## pycld3
Also pycld3 has wheels only for Python 3.6-3.9, and installing pycld3 fails with
```
RuntimeError: The Protobuf compiler, `protoc`, which is required for building this package, could not be found.
```
Next release of pycld3 [should include wheel for Python 3.10](https://github.com/bsolomon1124/pycld3/issues/18).

## Other dependencies
Dropping Python 3.7 support  [would allow upgrading some pinned dependencies](https://github.com/NatLibFi/Annif/pull/549#issuecomment-1008662572).

juhoinkinen:
This changes CI/CD tests to be run on Python 3.8-3.10 instead on 3.7-3.9 and updates mentions of Python version support to 3.8+. Wiki may still need updates for this.

Also 
- upgrades pinned dependencies to newest available (except for [simplemma](https://github.com/adbar/simplemma/releases/tag/v0.7.0) as there is some breaking change), which is now possible since dropping Python 3.7 support
- switches to [use importlib of the standard library](https://github.com/NatLibFi/Annif/pull/592/commits/61df4def91065295709d5b45f8fb2ad4a1c7bc28), previously the 3rd party package was needed for Python 3.7
- [verifies installed packages have compatible dependencies](https://github.com/NatLibFi/Annif/pull/592/commits/3714fcd3d8adeb9fadcd4c050af23332f89ae3a7) using `pip check`

Compatibility of models trained with older dependencies should still be checked and documented.

Closes #589.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -16,7 +16,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, 3.10]
     name: test on Python ${{ matrix.python-version }}
 
     steps:
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -16,7 +16,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, 3.10]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     name: test on Python ${{ matrix.python-version }}
 
     steps:
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -39,22 +39,22 @@ jobs:
         python -m nltk.downloader punkt
         # Selectively install the optional dependencies for some Python versions
         # Install the optional neural network dependencies (TensorFlow and LMDB)
-        # - except for one Python version (3.8) so that we can test also without them
-        if [[ ${{ matrix.python-version }} != '3.8' ]]; then pip install .[nn]; fi
+        # - except for one Python version (3.9) so that we can test also without them
+        if [[ ${{ matrix.python-version }} != '3.9' ]]; then pip install .[nn]; fi
         # Install the optional Omikuji and YAKE dependencies
-        # - except for one Python version (3.8) so that we can test also without them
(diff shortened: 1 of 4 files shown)


juhoinkinen (commented):
Merge commit for resolving conflicts due to #587.

juhoinkinen:
I run evaluation of just released Finto AI models with the upgraded dependencies: Models work with the usual warnings by sk-learn and evaluation results change just slightly for some data-sets.

New releases of connexion and NumPy were made just yesterday, so I upgraded them now again. 

[Simplemma v0.7.0](https://github.com/adbar/simplemma/releases/tag/v0.7.0) includes a breaking change, and I'm not sure how that should be fixed (can the `simplemma.load_data()` be just dropped?), so simplemma is left pinned to v0.6.*.

osma:
I think current master (with the flake8 fix, #598) should be merged to this PR branch and if everything still works, it can be merged to master.

PR #597, opened 2022-08-02
Include labels without language tag and concepts without labels in vocabulary

### LINKED ISSUE #556: vocabulary in SKOS (Turtle serialization) should be loaded even in case of lacking language tags
Hi!
I've noticed some potentially inconsistent behaviour of the annif loadvoc command when loading voacabulary in ttl SKOS without language tags.

We've currently switched in our project from simple tsv vocabulary format to SKOS. I assumed that since we do not provide any information about language in tsv file, we don't necessarily have to add appropriate language tags in SKOS either. But it turned out, that loading vocabulary in SKOS without language tags prevents annif from creating the subject index (though original ttl file is being copied and gzipped file is being dumped). 

It seems, that when annif converts tsv to SKOS, it adds language tags (they're based upon language configuration in projects.cfg), but when it loads vocabulary directly from SKOS format, it checks if language tag for a label is the same as language code in projects.cfg, and when it's not **or if there is no language tag at all** it skips the whole concept:

```
def get_concept_labels(self, concept, label_types, language):
    return [str(label)
            for label_type in label_types
            for label in self.graph.objects(concept, label_type)
            if label.language == language]
```

```
@property
    def subjects(self):
        for concept in self.concepts:
            labels = self.get_concept_labels(
                concept, [SKOS.prefLabel, RDFS.label], self.language)
            notation = self.graph.value(concept, SKOS.notation, None, any=True)
            if not labels:
                continue
            label = labels[0]
            if notation is not None:
                notation = str(notation)
            yield Subject(uri=str(concept), label=label, notation=notation,
                          text=None)
```

I think, that maybe it would be safer to assume, that when there is no explicit information about the language in SKOS file (there is a label without the language tag), its language corresponds with the language defined in projects.cfg and skip a concept (or label from the concept) only when the language tag truly exists and is not equal to the language from projects.cfg.

If this change is not possible for some reasons, it would be nice to provide some information about this behaviour in annif wiki (it took me some time to find this "bug").

Maciej Sagata, National Library of Poland

osma:
Fixes #556 by modifying the way concepts from SKOS vocabularies are loaded. There are two main changes:

1. If a concept doesn't have a prefLabel in the configured language, but has a prefLabel without a language tag, use that instead.
2. If a concept doesn't have any suitable prefLabels (in the configured language, or without a language tag), generate a pseudo label from the qualified name (e.g. `yso:p12345` or `lcsh:sh85061212`)

This should improve the support for multilingual vocabularies and handle cases when SKOS data is missing language tags, which can happen for example when converting MARC21 records to SKOS like @macsag did when reporting #556.

Note that unlike the solution drafted in [this comment](), there is no BCP 47 style matching of language tag variants (e.g. `en` in the SKOS file would match the configured language `en-US`). I considered this out of scope for now (YAGNI principle) although it could easily be added later, but it would require using a library such as [langcodes](https://pypi.org/project/langcodes/) for the actual language tag matching. 

This PR may change the results for some multilingual corpora, for example the YSO based corpora used to train and evaluate models for Finto AI, because the vocabulary will now be larger in some cases. YSO usually lacks Swedish and/or English language labels for some recently added concepts and these used to be dropped when loading the vocabulary, but will now be included after this PR.

### DIFF
--- a/annif/corpus/skos.py
+++ b/annif/corpus/skos.py
@@ -49,10 +49,12 @@ class SubjectFileSKOS(SubjectCorpus):
         for concept in self.concepts:
             labels = self.get_concept_labels(
                 concept, [SKOS.prefLabel, RDFS.label], self.language)
+            if labels:
+                label = labels[0]
+            else:
+                # No labels - use qualified name instead (derived from URI)
+                label = self.graph.namespace_manager.qname(concept)
             notation = self.graph.value(concept, SKOS.notation, None, any=True)
-            if not labels:
-                continue
-            label = labels[0]
             if notation is not None:
                 notation = str(notation)
             yield Subject(uri=str(concept), label=label, notation=notation,
@@ -66,10 +68,19 @@ class SubjectFileSKOS(SubjectCorpus):
             yield concept
 
     def get_concept_labels(self, concept, label_types, language):
-        return [str(label)
-                for label_type in label_types
-                for label in self.graph.objects(concept, label_type)
-                if label.language == language]
+        # 1. Labels with the correct language tag
+        labels = [str(label)
+                  for label_type in label_types
+                  for label in self.graph.objects(concept, label_type)
+                  if label.language == language]
+        if labels:
+            return labels
+        # 2. Labels without a language tag
+        labels = [str(label)
+                  for label_type in label_types
+                  for label in self.graph.objects(concept, label_type)
+                  if label.language is None]
+        return labels
 
     @staticmethod
     def is_rdf_file(path):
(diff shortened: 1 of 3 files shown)


osma:
Hmm, I think at least the MLLM backend, possibly also YAKE and (less likely) STWFSA will need to be changed so that they don't rely on the label stored in the vocabulary. Otherwise they could be confused by the qnames.

osma:
> Hmm, I think at least the MLLM backend, possibly also YAKE and (less likely) STWFSA will need to be changed so that they don't rely on the label stored in the vocabulary. Otherwise they could be confused by the qnames.

Adjusted the MLLM code so it reads prefLabels directly. Checked YAKE and STWFSA, they are OK already as they are not using the labels from the vocabulary either.

osma:
Ready for wider testing. Code Climate still has a couple of complaints but I can't figure out how to address them without making the code harder to understand.

osma:
Rebased on current master (after the 0.59 release) and force-pushed.

juhoinkinen (approved):

Issue #559, opened 2022-01-28
Multilingual vocabularies


osma:
# Introduction

This is a proposed conceptual change to how Annif handles subject vocabularies. Currently, vocabularies are strictly monolingual. The same vocabulary can be used by multiple projects, as long as they share the same language. Typically, vocabularies have identifiers like `yso-fi` or `stw-en`, reflecting that they are specific to a single language, even though in reality both YSO and STW are multilingual.

# Proposed change

Annif vocabularies should be multilingual; that is, projects with different languages can still share the same vocabulary.

Note that #556 would be a small step in this direction and it would probably make sense to implement that first.

# Rationale

Making vocabularies multilingual would have the following benefits:

1. Enable scenarios where the vocabulary of the indexing language is different from the language of documents. For example, Finnish language documents indexed with LCSH (terms in English only), or German language documents indexed with YSO (Finnish, Swedish and English terms). At present this is not possible, or at least requires some dirty tricks.
2. Align with SKOS: concepts are language-agnostic, terms can be given in multiple languages.
3. Simplify configuration: no need to have separate vocabulary id's for different languages.
4. Simplify loading vocabularies: no need to load e.g. YSO three times just because you have projects in Finnish, Swedish and English. Loading once is enough. 

# Changes in Annif usage and implementation

This is somewhat speculative and may change when actually trying to implement this...

## Changes to configuration

The main change would be to start using language-agnostic vocabulary id's. So instead of

    vocab=yso-fi

use

    vocab=yso

Optionally, it should be possible to override the language of labels using a parameter (the default being the language of the project). For example, in a Finnish language project where the vocabulary is LCSH, English LCSH labels would be used like this:

    language=fi
    vocab=lcsh(en)

## Changes to vocabulary data files

Currently, the vocabulary is stored as three files: `subjects` (TSV), `subjects.ttl` (SKOS), `subjects.dump.gz` (dumped rdflib Graph). The last two don't need to change as they are already multilingual. Instead of a single `subjects` file, there should be one file per language, e.g. `subjects.fi.tsv`, `subjects.sv.tsv`, `subjects.en.tsv`.

## Changes to the loadvoc command

When given a SKOS file, the `loadvoc` code should figure out which languages are used in the labels and write as many `subjects.<lang>.tsv` files as there are languages.

[...]

PR #600, opened 2022-08-04
Make vocabularies multilingual

osma:
This PR implements #559 - making vocabularies multilingual, so that there is no need to use separate language-specific vocabulary id's such as `yso-fi`, `yso-sv` and `yso-en`. Instead the vocabulary id `yso` can be used for all projects and the `loadvoc` command needs to be executed just once as it will detect which languages are available in the vocabulary and load the labels in all available languages.

It's also possible to define/override the language of labels, for example to use `vocab=lcsh(en)` in a Finnish language project.

The changes need to be carefully tested as they are quite disruptive. Documentation (including the Annif tutorial) should be updated, mainly by stripping language suffixes from vocabulary id's in examples. However, old examples (vocabulary id's with a language suffix) should still keep working.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -187,7 +187,7 @@ def run_loadvoc(project_id, force, subjectfile):
     proj = get_project(project_id)
     if annif.corpus.SubjectFileSKOS.is_rdf_file(subjectfile):
         # SKOS/RDF file supported by rdflib
-        subjects = annif.corpus.SubjectFileSKOS(subjectfile, proj.language)
+        subjects = annif.corpus.SubjectFileSKOS(subjectfile)
     else:
         # probably a TSV file
         subjects = annif.corpus.SubjectFileTSV(subjectfile)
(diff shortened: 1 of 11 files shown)


osma:
Rebased on current master (which now contains PR #597 that was the starting point of this branch) and force-pushed.

osma:
This is more or less done now, pending review.

I tried to fix the issues reported by QA tools. Code Climate still complains about load_vocabulary, but I can't figure out how to make it better. Codecov says there's one more missed line than before in annif.vocab, but I can't find it in the detailed report.

juhoinkinen (approved):
LGTM

osma:
Thanks for reviewing @juhoinkinen ! I will merge this now, although we still need to do more testing before the next release, however there are other related changes to the vocabulary functionality coming up (possibly e.g. #602) and it makes sense to test them all in one go.

PR #610, opened 2022-08-17
Store vocabs in AnnifRegistry so they are shared between projects

### LINKED ISSUE #603: Share vocabulary objects between projects
Currently, each AnnifProject creates its own vocabulary (AnnifVocabulary object, which wraps a SubjectIndex) even though in reality many projects may use the same vocabulary - which is now even more likely since vocabularies are multilingual (#600). This leads to unnecessary use of RAM and also CPU when the subject index has to be recreated many times.

Instead, the vocabularies could be moved under AnnifRegistry which currently only stores projects. The vocabularies could be loaded lazily as needed using a get_vocab method, similar to the current get_vocab method in annif.vocab.

osma:
Fixes #603

This PR makes the handling of vocabularies more efficient by making it possible to use a single AnnifVocabulary instance shared by multiple projects. The vocabularies are now stored in AnnifRegistry, similar to how projects are stored.

I did a little benchmarking. I set up TFIDF, MLLM and Parabel projects with YSO as the vocabulary, as well as an ensemble using these three as sources. I trained them using the archaeology test corpora from the test suite. Then I measured the user time and maximum RSS of two commands targeting the ensemble project: a simple `annif suggest` and a parallelized `annif eval -j4` command against the fulltext documents in the test suite, before and after this PR:

|           | time before | time after | RSS before | RSS after |
|-----------|-------------|------------|------------|-----------|
| suggest   | 6.81        | 5.51       | 438072     | 305828    |
| eval -j 4 | 24.35       | 23.24      | 418492     | 304464    |

This PR avoids loading the YSO vocabulary four times (once per project) and instead loads it only once. The result is 1.1-1.3 second reduction in CPU time and 110-130MB reduction in RAM usage.

### DIFF
--- a/annif/project.py
+++ b/annif/project.py
@@ -9,7 +9,6 @@ import annif.analyzer
 import annif.corpus
 import annif.suggestion
 import annif.backend
-import annif.vocab
 from annif.datadir import DatadirMixin
 from annif.exception import AnnifException, ConfigurationException, \
     NotSupportedException, NotInitializedException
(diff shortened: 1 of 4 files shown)


juhoinkinen (approved):
LGTM

Issue #602, opened 2022-08-04
loadvoc command should take a vocabulary id, not project id


osma:
The `annif loadvoc` command currently takes a project ID as a parameter, like this:

    annif loadvoc yso-tfidf-fi path/to/yso.ttl

But this can be a bit misleading, because the same vocabulary could be used by many other projects and thus their vocabulary will be loaded/updated as well. The problem will be even more prominent after implementing #559 / #600 , which makes vocabularies multilingual, and thus even projects in different languages may share the same vocabulary (and vocabulary id).

I suggest changing this so that the `loadvoc` command instead takes a vocabulary ID, like this:

    annif loadvoc yso path/to/yso.ttl

This would align better with current reality, but of course it's a potentially disruptive change, since for example scripts that perform loadvoc operations have to be modified and all the relevant documentation updated, including the Annif tutorial. There could perhaps be a transition period where loadvoc with a project id keeps working but prints a deprecation warning...

There's also the question of how to deal with languages, especially when loading a vocabulary from a TSV file. Currently the language of a TSV vocabulary is inferred from the project configuration. But if the vocabulary is loaded directly, there is no project configuration, so the language may need to be specified directly, for example with a `--language` option (shortened to `-L` which is not otherwise used in current CLI commands). This would also be useful for SKOS vocabularies that lack language tags, as discussed in #556.

osma:
Based on a suggestion by @juhoinkinen , I think this would be a better approach:

1. Keep the `loadvoc` command around for a while (one release cycle?) in its current form, but mark it as deprecated and print a warning message when it is used.
2. Add new commands `load-vocab`, `list-vocabs`, `show-vocab` and possibly other similar ones as well (not necessarily all of them implemented at once). These would all work on vocabulary IDs instead of project IDs and would be analoguous to the `list-projects` and `show-project` commands.

PR #613, opened 2022-08-24
Restore ability to use vocab language different from project language

osma:
The ability to use a vocabulary language different from the project language was implemented in PR #600, but subsequently broken by mistake in PR #608. For example, it should be possible to use `vocab=lcsh(en)` in a project with `language=fi` where all documents are in Finnish but English language labels are used for LCSH concepts (which don't even have Finnish labels) both when reading corpora and outputting results.

This PR aims to restore that functionality by making sure that

1. When reading corpora in the directory-based format, labels are compared to vocabulary labels in the vocabulary language;
2. When performing suggest operations (CLI or REST), the labels of suggested subjects are in the vocabulary language;
3. When writing an evaluation results file which contains subject labels, the labels will be in the vocabulary language.

Currently there are unit tests to verify item 2. above, but not 1. or 3.

Also some of the test vocabularies were renamed and repurposed to better match current needs.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -226,7 +226,7 @@ def run_train(project_id, paths, cached, docs_limit, jobs, backend_param):
         documents = 'cached'
     else:
         documents = open_documents(paths, proj.subjects,
-                                   proj.language, docs_limit)
+                                   proj.vocab.language, docs_limit)
     proj.train(documents, backend_params, jobs)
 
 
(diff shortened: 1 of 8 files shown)


osma:
I think this is good enough for now. Need to get moving with the `load-vocabulary` command (#602) which will likely be touching some of the same bits of code anyway.

PR #611, opened 2022-08-22
Autogenerated CLI commands documentation on ReadTheDocs

juhoinkinen:
Switches to document Annif CLI commands on annif.readthedocs.io with automated builds instead of manually maintaining documentation in a GitHub wiki page 

Template for the page is in `docs/source/commands.rst`. 

Due to an [issue with sphinx-rtd-theme](https://github.com/readthedocs/sphinx_rtd_theme/issues/1115) (not showing bullet points in TOC lists) an older version of docutils (0.16) is used.

Edit: Also [upgrades to Click 8.1.* to allow sphinx-click to access full docstring after truncation marker `\f`
](https://github.com/NatLibFi/Annif/pull/611/commits/e830bd0f077334c8040181fd44bba00f53188e8e)

Closes #595.

PR #614, opened 2022-08-25
Implement load-vocab and list-vocab commands

### LINKED ISSUE #602: loadvoc command should take a vocabulary id, not project id

osma:
This PR adds two new CLI commands, `annif load-vocab` and `annif list-vocabs`. It also deprecates the old `annif loadvoc` command. The main difference between `load-vocab` and `loadvoc` is that `load-vocab` operates directly on the vocabulary and takes a vocabulary ID as argument, while `loadvoc` takes a project ID as argument. See #602 for more details on these.

Here are examples of running the new commands:

```
$ annif load-vocab dummy tests/corpora/dummy-subjects.csv 
Loading vocabulary from CSV file tests/corpora/dummy-subjects.csv...
updating existing subject index
saving vocabulary into SKOS file data/vocabs/dummy/subjects.ttl
```

As is apparent, the `load-vocab` command is more verbose than `loadvoc` which produced little or no output.

The new `list-vocabs` command shows some basic information about the vocabularies:

```
$ annif list-vocabs
Vocabulary ID       Languages                 Size  Loaded
----------------------------------------------------------
yso                 en,fi,sv                 37869  True  
stw                 de,en                     6243  True  
gnd                 -                            -  False 
yletest             -                            -  False 
arch                fi                         125  True  
ykl                 -                            -  False 
dummy               en,fi                        2  True  
```

The `loadvoc` command now shows a deprecation message (in red color when possible) and has also become slightly more verbose:

```
$ annif loadvoc dummy-fi tests/corpora/dummy-subjects.csv 
DeprecationWarning: The command 'loadvoc' is deprecated.
updating existing subject index
saving vocabulary into SKOS file data/vocabs/dummy/subjects.ttl
```

There are also changes to the CLI commands documentation that is nowadays published on ReadTheDocs (see #611). The section on `loadvoc` command has been deleted, since we don't want to encourage its use. I will file a separate issue on removing the `loadvoc` command in a future release, probably 0.60.

Fixes #602

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -19,6 +19,7 @@ import annif.registry
 from annif.project import Access
 from annif.suggestion import SuggestionFilter, ListSuggestionResult
 from annif.exception import ConfigurationException, NotSupportedException
+from annif.exception import NotInitializedException
 from annif.util import metric_code
 
 logger = annif.logger
@@ -37,6 +38,20 @@ def get_project(project_id):
     except ValueError:
         click.echo(
             "No projects found with id \'{0}\'.".format(project_id),
+            err=True)
+        sys.exit(1)
+
+
+def get_vocab(vocab_id):
+    """
+    Helper function to get a vocabulary by ID and bail out if it doesn't
+    exist"""
+    try:
+        return annif.registry.get_vocab(vocab_id,
+                                        min_access=Access.private)
+    except ValueError:
+        click.echo(
+            f"No vocabularies found with the id '{vocab_id}'.",
             err=True)
         sys.exit(1)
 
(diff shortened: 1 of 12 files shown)


juhoinkinen (approved):
LGTM

Wiki pages need update to align to these chages too.

osma:
> Wiki pages need update to align to these chages too.

Right, many wiki pages (e.g. all the backend pages) show use of the loadvoc command. Also vocabulary id's in example configurations and commands should be edited to remove the language suffixes that are no longer useful. I think it's best to do all the wiki edits synchronized with the 0.59 release.

PR #618, opened 2022-09-01
upgrade to simplemma 0.8 and disable unnecessary cache

### LINKED ISSUE #617: Upgrade simplemma dependency and remove function cache
Hi, I just noticed that you wrapped a custom class around [simplemma](https://github.com/adbar/simplemma/), notably to use a function cache.

Interestingly, I had the same idea which indeed speeds up the process. The cache is included in the versions that you use (0.7.* upwards), so you could delete the following line without any impact on performance:

https://github.com/NatLibFi/Annif/blob/aa50441e769a4c93866ca93ee07d3b9677c4d201/annif/analyzer/simplemma.py#L15

In addition, you could use the `lang` attribute (new in 0.7.0) directly with `simplemma.lemmatize()`.

Best,
Adrien

osma:
Fixes #617

This PR upgrades Simplemma to version 0.8.* (from 0.7.*) and removes the LRU cache, because Simplemma nowadays already has a similar cache. Thanks to @adbar for the tip!

### DIFF
--- a/annif/analyzer/simplemma.py
+++ b/annif/analyzer/simplemma.py
@@ -1,6 +1,5 @@
 """Simplemma analyzer for Annif, based on simplemma lemmatizer."""
 
-import functools
 import simplemma
 from . import analyzer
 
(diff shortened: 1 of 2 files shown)


adbar:
The last change looks good to me! :heavy_check_mark:

PR #623, opened 2022-09-23
remove loadvoc CLI command and related tests

osma:
This PR removes the `annif loadvoc` command which was deprecated in PR #614 but still available in the Annif 0.59 release but scheduled for removal in 0.59. The `load-vocab` command should be used instead.


osma:
All seems well in CI so I'll merge this.

PR #626, opened 2022-09-28
use Simplemma instead of pycld3 for language detection

### LINKED ISSUE #593: Replace pycld3 dependency?

osma:
This draft PR is similar to PR #615 except it replaces the language detection previously performed using pycld3 with [Simplemma based language detection](https://github.com/adbar/simplemma/#language-detection) (by @adbar)

I intend to benchmark this against PR #615 and the current status quo (with pycld3) in the near future.

Fixes #593

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -59,7 +59,7 @@ jobs:
         # Selectively install the optional dependencies for some Python versions
         # For Python 3.8:
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then
-          poetry install -E "nn omikuji yake voikko pycld3";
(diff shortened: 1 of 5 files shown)


osma:
Rebased on current `master` (with black & isort reformatting), fixed up and force-pushed.

juhoinkinen (approved):
LGTM

PR #630, opened 2022-10-07
Allow overriding subject label language in CLI and REST suggest operations

### LINKED ISSUE #628: Allow selecting label language in suggest operations
The suggest operation (both CLI and REST) always returns the subject labels in the configured vocabulary language, even though the vocabulary could include labels in other languages. For example, YSO is trilingual (fi, sv, en) and thus the subject labels could be given in any of these languages.

We could make it possible to select the subject label language in suggest operations. In the CLI, it could be a parameter `-L` / `--language`:

    annif suggest -L en yso-mllm-fi <document.txt          # return English language subject labels
    annif suggest --language en yso-mllm-fi <document.txt  # same but with long option

In the REST API, it could be a parameter `lang` or `language` that works similarly to the current `limit` and `threshold` parameters.

Currently the Finto AI user interface has similar functionality, but it relies on the Finto REST API to fetch labels in non-default languages. Having this available in the Annif API would simplify the implementation.

osma:
Fixes #628

This PR implements an option `--language` / `-L` for the `annif suggest` command as well as an optional `language` parameter for the REST API `suggest` method, both of which allow overriding the language of returned subject labels.

TODO:

- [x] error handling: don't break if the given language is broken or not available in the vocabulary
- [x] implement the `--language` option for the `annif index` command as well
- [x] address QA tool complaints

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -323,9 +323,10 @@ def run_learn(project_id, paths, docs_limit, backend_param):
 @click.argument('project_id')
 @click.option('--limit', '-l', default=10, help='Maximum number of subjects')
 @click.option('--threshold', '-t', default=0.0, help='Minimum score threshold')
+@click.option('--language', '-L', help='Language of subject labels')
 @backend_param_option
 @common_options
-def run_suggest(project_id, limit, threshold, backend_param):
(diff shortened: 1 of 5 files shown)


juhoinkinen (approved):
LGTM

PR #636, opened 2022-10-28
Better set up for docker-compose

juhoinkinen:
The docker-compose setup has not been really used much. It is mentioned in the [Docker Wiki page](https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker#using-annif-with-gunicorn-and-nginx) mostly as a step towards production environment. To make NGINX proxy requests to Annif, the NGINX configuration has been written in the `annif/nginx/nginx.conf`file, which quite uselessly clutters the annif directory. 

However [there is an alternate solution](https://github.com/docker/compose/issues/3538): creating the config file with Docker cmd from content written inline in the `docker-compose.yaml`.

Changes:
- Remove `annif/nginx` directory with the config file
- Create `nginx.conf` with Docker run command
- Remove unnecessary bash container
- Better name for Annif container


docker-compose.yml
```diff
@@ -2,27 +2,29 @@ version: "3"
...
-      - ${ANNIF_PROJECTS}:/annif-projects
-    user: ${MY_UID}:${MY_GID}
-    stdin_open: true
-    tty: true
-    command: bash
-
-  gunicorn_server:
-    image: quay.io/natlibfi/annif
+  annif_app:
+    image: quay.io/natlibfi/annif:latest
```
osma, line 6: I see that this changed to `:latest`, was this intended?

juhoinkinen: Yes, the tag is `latest` by default; this just makes the behaviour more explicit.

osma (approved):
LGTM, although I didn't test it

juhoinkinen:
On second thought, would it be better to remove the `docker-compose.yaml` from the Annif repository and just show it in Wiki? If one wants to use Annif with Docker, it does not make sense to have to clone the repository for using the compose file.

osma:
My slight preference would be to keep it as a file in the main repo. I see your point, but I think it's better to keep the file in a visible place under version control, and also being a part of Annif releases.

juhoinkinen:
I updated the [Wiki page](https://github.com/NatLibFi/Annif/wiki/Usage-with-Docker) to align to these changes.

osma:
The [NGINX proxy server](https://www.nginx.com/resources/wiki/) link on that page seems outdated, probably it should be switched to https://nginx.org/en/docs/ or some other suitable resource

juhoinkinen:
> The [NGINX proxy server](https://www.nginx.com/resources/wiki/) link on that page seems outdated, probably it should be switched to https://nginx.org/en/docs/ or some other suitable resource

Link switched to the up-to-date one.

PR #638, opened 2022-11-01
add support for spawn multiprocessing mode

### LINKED ISSUE #637: Support spawn multiprocessing mode on Mac OS and Windows
Currently Annif performs parallel processing using the multiprocessing module for some functionality, for example parallel evaluation and preparing the training data for the NN ensemble. This relies (implicitly) on the `fork` model, which is the default on Linux, and has some nice benefits: models only need to be loaded once, and most of the memory can be shared between worker processes.

However, the `fork` mode is not available on Windows. On Mac OS, `spawn` is the default since Python 3.8 because there are problems with the `fork` mode. In practice, the parallel processing functionality will not work on Mac OS or Windows (unless Annif is run in a Docker container or a Linux VM). See for example [this report](https://groups.google.com/g/annif-users/c/dj7ifu3yGhE/m/8B3FsDXyAAAJ) on annif-users.

We could support the `spawn` mode as well (and possibly `forkserver`, although it's not available on Windows). This would be less efficient than `fork`, since each worker needs to load models separately and they will all require more memory. But it could still be useful on machines with many CPU cores and lots of RAM available.

osma:
This (draft) PR attempts to add support for the `spawn` multiprocessing mode. This is the only mode supported on Windows, and the default mode on Mac OS. The `fork` mode, supported on Linux and some other *nix systems, is more efficient because it allows loading models only once and then reusing them in forked child processes, with most of the memory shared between the processes. But `spawn` mode is still better than not being able to perform multiprocessing at all.

This PR needs more testing, ideally on both Mac OS and Windows. Currently there is just a single unit test that exercises parallel evaluation using the `spawn` mode.

Fixes #637

### DIFF
--- a/annif/parallel.py
+++ b/annif/parallel.py
@@ -4,6 +4,11 @@
 import multiprocessing
 import multiprocessing.dummy
 
+# Start method for processes created by the multiprocessing module.
+# A value of None means using the platform-specific default.
+# Intended to be overridden in unit tests.
+MP_START_METHOD = None
+
 
 class BaseWorker:
     """Base class for workers that implement tasks executed via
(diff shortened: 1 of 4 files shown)


osma:
Successfully tested parallel training of NN ensemble with this PR branch on a MacBook with an M1 processor (arm architecture) by @holehan, according to [this report](https://groups.google.com/g/annif-users/c/dj7ifu3yGhE/m/Na1N7ZzWAAAJ). Thanks a lot!

osma:
Rebased on current `master` (with black & isort reformatting), fixed up and force-pushed.

monalehtinen:
Tested parallel processing in this PR branch using Windows 10 (Intel Core i5-8250U processor). Succesfully trained an nn-ensemble model, ran into no problems related to PR.

PR #640, opened 2022-11-04
Use black code style

### LINKED ISSUE #609: Use Black code style?
[Black](https://github.com/psf/black) seems to be in fashion. Adopting it would rid us of code style related worries.

If we decide to use it, we should consider using the following [integrations](https://black.readthedocs.io/en/stable/integrations/index.html):

* integrate Black with IDEs used by developers, when possible
* have a simple command to reformat all files with Black (or just check conformity)
* have a git pre-commit hook that performs Black reformatting (e.g. [this tutorial](https://towardsdatascience.com/keep-your-code-clean-using-black-pylint-git-hooks-pre-commit-baf6991f7376) which shows how to do it without using pre-commit.com, which would be an option too)
* have Black check the code under GitHub Actions CI

Black would replace autopep8 that we're currently using. There's still a need for flake8 checks, because flake8 checks other things than just pep8 code style.

juhoinkinen:
Integrates Black code style tool to development flow and formats codebase to follow [Black default style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html).

The main changes that Black makes when using the default style seem to be:
- changes single quotes to double quotes (where single quotes are not necessary)
- joins some short lines, because max. line length is 88 instead of 79
- adds a trailing comma to the last argument of list of arguments spanning multiple lines
- in list etc. comprehensions a newline is added after open parenthesis
- lines are broken before a binary operator
- adds newlines after import statement
- adds spaces around not-simple slices, e.g. to `[i : i + j]`

It would be possible to skip some of the changes with custom configuration (like in branch [issue609-use-black-code-style-less-changes](https://github.com/NatLibFi/Annif/tree/issue609-use-black-code-style-less-changes), where quotes are not changed and the 79 max. line length is retained). However, I think it could be better just to use the default Black style.

The [`.git-blame-ignore-revs` file](https://github.com/NatLibFi/Annif/pull/640/commits/50f7424f0e057c23b8aab45075f096dbf21fe57f#diff-f247fbfbe928b907db42554d0b3006b28dd11c25a59be031abda73b11a2c934a) hides (some) of the changed lines in git-blame view of the [style migration commit](https://github.com/NatLibFi/Annif/pull/640/commits/3bc18907354a40f1d89dca1833a2719ba7fb0933).

## Manual local usage
The Black tool is added to the development dependencies, and the codebase can be formatted with one command. If nothing is left to be changed, the run is:
```
black .
All done! ✨ 🍰 ✨
92 files left unchanged.
```
When using Black the `autopep8.sh` script is unnecessary and it is removed.

## Automated local usage
The pre-commit hook for linting with black and flake8 can be configured by using the following in the file `.git/hooks/pre-commit`:
```bash
#!/bin/sh

black . --check --diff
flake8
```
There exists the [precommit-framework](https://github.com/pre-commit/pre-commit), which could automate the installation of the hook, but the hook would need to be defined in `.pre-commit-config.yaml` file. I don't know if using the precommit-framework is worth adding yet another dev dependency. [This blog](http://www.sefidian.com/2021/08/03/how-to-use-black-flake8-and-isort-to-format-python-codes/) showed how that could be used.

(There exists also the [pytest-black](https://github.com/shopkeep/pytest-black) plugin, but it seems unmaintained.)

## CI usage
Adds a new "lint with Black" job to GitHub Actions CI workflow, which uses the [Action provided by Black developers](https://github.com/psf/black/blob/main/action.yml) to just check the format (uses options `--check --diff`). It is run in parallel with unit test jobs. 

I noticed that the separate Flake8 step for linting has been redundant, because flake8 linting is run also with the pytest, so the separate Flake8 step is removed.

## IDE integration
In VS Code the Black was already included and [could be selected as the formatter](https://dev.to/adamlombard/how-to-use-the-black-python-code-formatter-in-vscode-3lo0).

Closes #609.


juhoinkinen:
BTW, there exists also the [isort](https://github.com/PyCQA/isort) tool to sort import statements, which would be one automation step more.

juhoinkinen:
[CodeClimate complains about too long lines.](https://codeclimate.com/github/NatLibFi/Annif/pull/640)

The complaints by other code-style tools seem unimportant. The SonarCloud complaints about http protocol could could be just marked safe.

juhoinkinen:
@osma How does this look? One question is whether to switch using max. line length of 88 chars.

osma:
I think it looks very good already.

Re: line length, I think 88 is OK, but is there a way to tell that to Code Climate so it won't complain?

osma (approved):
Very good!

PR #641, opened 2022-11-07
Use isort to order import statements

juhoinkinen:
PR #640 introduced Black style and tool for checking and autoformatting code. However, Black does not care about the order of import statements, but [isort](https://github.com/PyCQA/isort) does. Using isort a developer does not need worry about the order of imports.

This PR contains similar changes as the one for Black.

## Manual local usage
Just run this to sort imports:
```
isort .
```
## Automated local usage
isort can check the import order automatically when committing by adding this line to `.git/hooks/pre-commit`:
```
isort . --check-only --diff
```
## CI usage
isort check is added to the linting job in GH Actions.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,4 +1,4 @@
 # Migrate code style to Black
 3bc18907354a40f1d89dca1833a2719ba7fb0933
 # Reorder import statements with isort
-7027e328451d2482b8072124113bd257ce021af1
+68a72c5a603283f70abce2651dcde9c6f0177c41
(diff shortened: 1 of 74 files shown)


juhoinkinen:
> Looks very good. A couple of quick comments:
>  - isort seems to have a [Black profile](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html), we should probably enable that in `pyproject.toml` for maximum compatibility (whatever it does, I'm not sure)
> 
Done already in [Configure isort to be compatible with Black](https://github.com/NatLibFi/Annif/pull/641/commits/920d948161df3984cc4b60469bfa902b972e9a92).

>   - now that both Black and isort will be expected from commits, it would be good to create a CONTRIBUTING.md file which documents, among other things, what style checks/fixups to perform before committing. [This](https://mozillascience.github.io/working-open-workshop/contributing/) seems to be a good introduction to writing one. (this could of course be another PR, not part of this one)

:+1:

osma (approved):
Ah, right, I somehow missed the changes to pyproject.toml even though I looked for them...

PR #645, opened 2022-11-21
Use tailored user-agent in requests by HTTP-backend

juhoinkinen:
The http-backend uses requests library, and so the headers have now user-agent `python-requests/2.28.1`, which is shown e.g. in logs of the targeted server.

It is better to use a user-agent tailored for Annif. This PR changes the user-agent that http-backend sends in request headers to `Annif/<version>`, e.g. currently to `Annif/0.60.0.dev0`.

### DIFF
--- a/annif/backend/http.py
+++ b/annif/backend/http.py
@@ -2,6 +2,8 @@
 and returns the results"""
 
 
+import importlib
+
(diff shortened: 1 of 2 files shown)


osma (approved):
LGTM!

PR #646, opened 2022-11-21
Use axios in web UI

UnniKohonen:
This PR replaces the jQuery ajax calls in the web UI with methods from the Axios library.


osma:
Now that the Bootstrap 5 upgrade has been merged (#648), I suggest merging current `master` into this branch (or rebasing on `master` if you like) so you get the new Bootstrap, which doesn't depend on jQuery. Then it should be possible to remove jQuery entirely in this PR.

PR #648, opened 2022-11-23
Upgrade Bootstrap to version 5.2.2

UnniKohonen:
This PR upgrades Bootstrap from version 4.5.0 to 5.2.2 in the web UI. Also makes small changes to html and css to fix errors resulting from the upgrade.


osma (approved):
Looks good! I tested the PR. There are minor changes in the layout: the content area is a little wider, and the circular buttons with the different limit values (10, 15, 20) are a bit further apart than before. But these are not important regressions. Good for merging!

PR #649, opened 2022-11-25
Upgrade REST API spec to OpenAPI 3

### LINKED ISSUE #644: Upgrade API spec to OpenAPI 3
Annif uses [Connexion](https://github.com/spec-first/connexion) to provide the REST API based on a spec file. The current spec file [annif.yaml](https://github.com/NatLibFi/Annif/blob/master/annif/swagger/annif.yaml) is written following the Swagger 2.0 specification. But this is quite obsolete, as its successor OpenAPI 3.0 was released already in 2017.

We should convert the spec file to OpenAPI 3.0 and probably also switch the name of the directory it resides in from `swagger` to `openapi`.

I already tried this, using [Swagger Editor](https://editor.swagger.io/) for conversion, but the converted version didn't seem to work - it gives a bunch of validation errors when I try to use `annif run` or run unit tests.

osma:
We currently specify the REST API using Swagger 2.0. This has long ago been superseded by OpenAPI 3.0, which is nowadays also well supported by the Connexion framework.

This PR updates the REST API spec for OpenAPI 3.0 and renames the directory it is stored in from `swagger` to `openapi`. I used the Swagger Editor for an initial conversion, but I had to make several fixes until I got it working. Also there are minor changes to the implementation code in `rest.py` that were necessary to make the new spec work.

Fixes #644


osma:
There are still problems with parameter handling

annif/openapi/annif.yaml
```diff
@@ -0,0 +1,274 @@
...
+                limit:
+                  type: integer
+                  description: maximum number of results to return
+                  format: int32
+                  default: 10
+                threshold:
+                  type: number
+                  description: minimum score threshold, below which results will not
+                    be returned
+                  default: 0.0
```
juhoinkinen, line 70: The numerical values could have min. and max. specified. 

I remember at some point I could produce an error in Annif side when using negative value for some parameter, but cannot do it now anymore, so this is ok as it is.

juhoinkinen (approved):

annif/openapi/annif.yaml
osma: I added sensible minimum and maximum values, thanks for the suggestion!

PR #650, opened 2022-11-25
Avoid using pytest-flake8 plugin

### LINKED ISSUE #599: Avoid using pytest-flake8, run flake8 separately instead
We are currently using pytest-flake8, which runs flake8 code style checks within pytest. But it has turned out to be fragile (see PR #598 and https://github.com/tholo/pytest-flake8/issues/87) and many projects are switching to a setup where flake8 is run separately (see https://github.com/pypa/pep517/pull/147 , https://github.com/pytest-dev/pytest/issues/9217#issuecomment-965144375, https://github.com/miketheman/pytest-socket/issues/82 )

We should also consider doing this. At the moment it is not possible to use the recently released flake8 version 5 because it's not compatible with pytest-flake8 1.1.1, so we have for now pinned flake8 to version 4 in #598, but this is not a good long term solution.

juhoinkinen:
Closes #599.

Unpins flake8. On my laptop version 5.0.4 get installed, while [6.0.0 was released two days ago](https://pypi.org/project/flake8/).

In the CI/CD pipeline I added flake8 step that both installs flake8 from PyPI with pip and runs it. There version 6.0.0 get installed. I did not find a good GH Action for flake8 (an official one or any which seemed otherwise reputable), so it can be better to install it this way.

The mismatch of the versions is unfortunate, but makes no difference at this time. 

Also I'm thinking that it might be a good idea to switch installing also isort and Black from PyPI (instead of using the GH Actions), and using Poetry for that (regular dependencies can be omitted by `poetry install --only dev`).


osma:
Looks very good!

> Also I'm thinking that it might be a good idea to switch installing also isort and Black from PyPI (instead of using the GH Actions), and using Poetry for that (regular dependencies can be omitted by poetry install --only dev).

I think that sounds like a good idea! Would it make sense to do this change already in this PR, or a new one after this has been merged?

juhoinkinen:
> Looks very good!
> 
> > Also I'm thinking that it might be a good idea to switch installing also isort and Black from PyPI (instead of using the GH Actions), and using Poetry for that (regular dependencies can be omitted by poetry install --only dev).
> 
> I think that sounds like a good idea! Would it make sense to do this change already in this PR, or a new one after this has been merged?

Better to make a new PR for that think, it might need some experimenting to make it work...

osma (approved):
I tested this locally on my laptop. Works fine, although running `poetry lock` and `poetry install --all-extras` didn't remove `pytest-flake8` so I removed it manually using `pip`.

But as the pytest option `--flake8` was removed in this PR, it won't be used by `pytest` anyway, so even if the package remains installed it shouldn't do any harm.

PR #651, opened 2022-11-25
Upgrade TensorFlow to version 2.11.*

osma:
We are currently using TensorFlow 2.9.1, which has reported vulnerabilities. Also dependabot is complaining about them.
It would be trivial to upgrade to 2.9.3, where those issues have been fixed.

But instead, I tested upgrading directly to 2.11.0. Everything seems to work fine with this new version. I also tested a small model that was trained on 2.9.1, before the upgrade, and it worked perfectly well.

I pinned this as `2.11.*` so that upgrading to patch releases is allowed in the future.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ simplemma = "0.9.*"
 
 fasttext-wheel = {version = "0.9.2", optional = true}
 voikko = {version = "*", optional = true}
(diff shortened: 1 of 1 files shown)


osma:
On second thought, I think this is fine and basically a minimal upgrade, I'll merge it as is.

PR #652, opened 2022-11-29
Website 2022

UnniKohonen:
This PR:
- Adds KB to the list of current users
- Upgrades jQuery to version 3.6.1 from version 3.3.1
- Moves current users into its own section


osma (approved):
LGTM!

Issue #655, opened 2022-12-20
REST API root resource with version information


osma:
The Annif REST API currently doesn't have a root resource. A URL such as http://127.0.0.1:5000/v1/ will give a 404 response.

The API could provide a root resource with some basic information about the API service, including the version of Annif, as a JSON object. Something like this:

```json
{ "title": "Annif REST API", "version": "0.60.0" }
```

One fancier way to do this would be to implement the [JSON Home proposal](https://datatracker.ietf.org/doc/html/draft-nottingham-json-home-06). See [this](https://docs.openstack.org/keystone/rocky/user/json_home.html) for a practical example and [this](https://apievangelist.com/2017/08/03/api-discovery-using-json-home/) for some easy to understand examples. But this may be overkill for the purpose of Annif and the JSON Home spec doesn't seem to be widely used.

Elasticsearch also provides information about the API in a root object as JSON. You can see how it looks (without installing it for yourself) in [this](https://community.progress.com/s/article/what-version-of-elasticsearch-do-i-have-installed) post.

The Annif version information, once exposed via the REST API, could be shown in the Finto AI user interface - see https://github.com/NatLibFi/FintoAI/issues/8

juhoinkinen:
Closed by #658.

PR #658, opened 2022-12-23
Add REST API root resource

UnniKohonen:
This PR adds a resource to the root of the REST API (i.e. http://<annif_host>/v1/) that gives basic information on the API. The information given is a title for the API and the version of Annif being used. The response is a json object with the format:
```json
{
    "title": "Annif REST API",
    "version": "0.60.0"
}
```

Implements #655

### DIFF
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -5,6 +5,19 @@ info:
 servers:
 - url: /v1
 paths:
+  /:
+    get:
+      tags:
+      - API information
+      summary: get basic information about the API service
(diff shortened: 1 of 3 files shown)


osma:
Looks good so far. A couple of suggestions:

- black formatting should be applied to the dict returned by show_info (see [details](https://github.com/NatLibFi/Annif/actions/runs/3765666027/jobs/6401379642))
- there should be a unit test in tests/test_rest.py checking that the show_info method returns what is expected

osma (approved):
Excellent work!

Issue #631, opened 2022-10-10
Language detection method in REST API & CLI


osma:
Currently language detection is only used for the [language filtering transformation](https://github.com/NatLibFi/Annif/wiki/Transforms#filter_lang-transform), but there have been some requests about providing this functionality also for users of the REST API, e.g. for the selection of an appropriate model.

We could have a method in the REST API for language detection, something like

    POST /detect-language

with the parameters `text` (input text whose language to detect) and `candidates` (list/array of language codes to consider, e.g. `["fi", "sv", "en"]`). Not sure about whether it's better to use form data or a JSON object to wrap the parameters. The return format would be a JSON object containing an array of results, something like this:

```json
{
  "results": [
    {"language": "fi", "score": 0.85},
    {"language": "sv", "score": 0.3},
    {"language": "en", "score": 0.3},
    {"language": null, "score": 0.1}
  ]
}
```

Here, the scores are arbitrary values between 0.0 and 1.0, and the language `null` stands for unknown language. The scores wouldn't necessarily add up to 1.

There could also be a similar CLI command for symmetry.

PR #659, opened 2022-12-29
Add language detection to REST API

UnniKohonen:
This PR adds the ability to detect the language of a text to the REST API. The language detection uses the simplemma python library.

A POST method is added to the end-point `/detect-language`. It expects the request body to include a json object with the text whose language is to be detected and a list of candidate languages as their [IETF BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag) ~[ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)~. For example:
```json
{
  "languages": ["en", "fi"],
  "text": "A quick brown fox jumped over the lazy dog."
}
```

The response is a json object with the format:
```json
{
  "results": [
    {"language": "en", "score": 0.85},
    {"language": "fi", "score": 0.1},
    {"language": null, "score": 0.1} 
  ]
}
```
where the scores range from 0 to 1 and a `null` value is used for an unknown language.

Implements REST API part of #631.

---

@juhoinkinen: I edited this for the latest changes (parameter name change candidates -> languages).

### DIFF
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -190,7 +190,7 @@ paths:
       operationId: annif.rest.detect_language
       requestBody:
         content:
-          application/x-www-form-urlencoded:
+          application/json:
             schema:
               type: object
               required:
@@ -200,6 +200,7 @@ paths:
                 text:
                   type: string
                   description: input text
+                  example: A quick brown fox jumped over the lazy dog.
                 candidates:
                   type: array
                   description: candidate languages as ISO 639-1 codes
@@ -207,6 +208,7 @@ paths:
                     type: string
                     maxLength: 2
                     minLength: 2
+                    example: en
         required: true
       responses:
         200:
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -182,6 +182,39 @@ paths:
         "503":
           $ref: '#/components/responses/ServiceUnavailable'
       x-codegen-request-body-name: documents
+  /detect_language:
+    post:
+      tags:
+      - Languages detection
+      summary: detect the language of a text given a list of candidate languages
+      operationId: annif.rest.detect_language
+      requestBody:
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              required:
+              - text
+              - candidates
+              properties:
+                text:
+                  type: string
+                  description: input text
+                candidates:
+                  type: array
+                  description: candidate languages as ISO 639-1 codes
+                  items:
+                    type: string
+                    maxLength: 2
+                    minLength: 2
+        required: true
+      responses:
+        200:
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/DetectedLanguages'
 components:
   schemas:
     ApiInfo:
@@ -316,6 +349,21 @@ components:
                 type: string
                 example: Vulpes vulpes
       description: A document with attached, known good subjects
+    DetectedLanguages:
+      type: object
+      properties:
+        results:
+          type: array
+          items:
+            type: object
+            properties:
+              language:
+                type: string
+                example: en
+              score:
+                type: number
+                example: 0.85
+      description: Candidate languages with their associated scores
     Problem:
       type: object
       properties:
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -185,7 +185,7 @@ paths:
   /detect-language:
     post:
       tags:
-      - Languages detection
+      - Language detection
       summary: detect the language of a text given a list of candidate languages
       operationId: annif.rest.detect_language
       requestBody:
@@ -203,10 +203,10 @@ paths:
                   example: A quick brown fox jumped over the lazy dog.
                 candidates:
                   type: array
-                  description: candidate languages as ISO 639-1 codes
+                  description: candidate languages as IETF BCP 47 codes
                   items:
                     type: string
-                    maxLength: 2
+                    maxLength: 3
                     minLength: 2
                     example: en
         required: true
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -217,6 +217,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/DetectedLanguages'
+        400:
+          description: Bad Request
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Problem'
 components:
   schemas:
     ApiInfo:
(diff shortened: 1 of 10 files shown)


juhoinkinen:
I see this is only a draft at the moment, but I took a glance and I think it would be better that the end-point name had hyphen instead of underscore (`/detect_language` - > `/detect-language`), [it seems to be the preferred convention](https://specs.openstack.org/openstack/api-wg/guidelines/naming.html#rest-api-resource-names).

annif/openapi/annif.yaml
```diff
@@ -134,6 +134,41 @@ paths:
               schema:
                 $ref: '#/components/schemas/Problem'
       x-codegen-request-body-name: documents
+  /detect-language:
+    post:
+      tags:
+      - Languages detection
```
osma, line 140: I think this should be "Language detection"

annif/openapi/annif.yaml
```diff
@@ -134,6 +134,41 @@ paths:
...
+              - text
+              - candidates
+              properties:
+                text:
+                  type: string
+                  description: input text
+                  example: A quick brown fox jumped over the lazy dog.
+                candidates:
+                  type: array
+                  description: candidate languages as ISO 639-1 codes
```
osma, line 158: Please change "ISO 639-1 codes" to "IETF BCP 47 codes", because not all supported languages are necessarily in ISO 639-1. (I noticed the same problem in Simplemma README and opened an [issue](https://github.com/adbar/simplemma/issues/30) there)

annif/openapi/annif.yaml
```diff
@@ -134,6 +134,41 @@ paths:
...
+                  items:
+                    type: string
+                    maxLength: 2
```
osma, line 161: Should be 3, as some valid BCP 47 language tags (e.g. `enm`, `hbs`) may have 3 characters

annif/rest.py
```diff
@@ -50,6 +51,18 @@ def show_project(project_id):
     return project.dump()
 
 
+def detect_language(body):
+    """return scores for detected languages formatted according to Swagger spec"""
+
+    scores = lang_detector(body.get("text"), tuple(body.get("candidates")))
+    return {
+        "results": [
+            {"language": s[0] if s[0] != "unk" else None, "score": s[1]}
```
osma, line 60: How about this instead:

```python
            {"language": lang if lang != "unk" else None, "score": score} for lang, score in scores
```

I think it would be easier to understand than the numeric indexing.

osma (changes requested):
Looks like a good start. I gave some comments on individual code lines. In addition to those:

- black formatting should be applied (see [details](https://github.com/NatLibFi/Annif/actions/runs/3822094392/jobs/6501892811))
- there should be a unit test in tests/test_rest.py which exercises the detect_language method

osma:
Thanks for adding tests. A few more things:

1. There should be tests for special cases, e.g. empty input, no candidate languages, unknown or malformed candidate languages...
2. What if there are 20 candidate languages? How much memory will it take to handle the query? Will the memory be released afterwards? (I think not, Simplemma keeps models in memory AFAIK)

osma:
> Right now, when making a request with no candidates or unknown candidates, the endpoint returns an empty list

The good news is that it's not crashing! :grin: 

My opinions on these cases:

1. With no candidates given, I think it would be good to return a 400 Bad Request status code, with a descriptive error message.

2. With unknown candidates (language codes that are unrecognized/unsupported by simplemma), I think a 400 Bad Request with a descriptive error message would also be appropriate.

There should be unit tests to check that these are indeed the results.

> and when making a request with no text, it returns { "language": null, "score": 1 }. Does this make sense? Or should it always return the unknown language with score 1 when the input is incorrect?

I think this is OK, but there should also be a unit test for this special case.

> I also tested making a request with all 48 possible language candidates. I had about 4 GB of free memory which was used amost completely after making the request. The memory isn't released automatically afterwards but it is freed if the endpoint is accessed again (simplemma is run again) or annif is restarted. Making other requests also slows down a lot after runnig simplemma with all candidates.

Great, thanks for testing! This is mostly what I suspected, although it's a surprise that accessing the endpoint again will free the memory. (Maybe this has to do with Flask running in development mode?)

This has some potential for DoS situations (intended or not), but I guess it's hard for us to avoid that given how Simplemma works. We could, however, limit the number of candidate languages per request to, say, at most 5. What do others think? @juhoinkinen ?

We could also try to work with the Simplemma maintainer if we want to change the way Simplemma allocates and releases models. For example, it could be possible to ask Simplemma to release the memory immediately or after a set period like 60 seconds after use.

juhoinkinen:
> > I also tested making a request with all 48 possible language candidates. I had about 4 GB of free memory which was used amost completely after making the request. The memory isn't released automatically afterwards but it is freed if the endpoint is accessed again (simplemma is run again) or annif is restarted. Making other requests also slows down a lot after runnig simplemma with all candidates.
> 
> Great, thanks for testing! This is mostly what I suspected, although it's a surprise that accessing the endpoint again will free the memory. (Maybe this has to do with Flask running in development mode?)
> 
> This has some potential for DoS situations (intended or not), but I guess it's hard for us to avoid that given how Simplemma works. We could, however, limit the number of candidate languages per request to, say, at most 5. What do others think? @juhoinkinen ?

Limiting the number of candidate languages seems reasonable. If there is no simple way to make the limit configurable, 5 could be a good number for that.

> We could also try to work with the Simplemma maintainer if we want to change the way Simplemma allocates and releases models. For example, it could be possible to ask Simplemma to release the memory immediately or after a set period like 60 seconds after use.

I noticed there is an [issue in Simplemma repository](https://github.com/adbar/simplemma/issues/33) about allocating and deallocating memory, which was opened just yesterday.

juhoinkinen:
Just started to think, if some some testing could be performed also in [tests/test_swagger.py](https://github.com/NatLibFi/Annif/blob/master/tests/test_swagger.py). I don't remember just what more functionality does the tests in `test_swagger.py`  cover than those in `test_rest.py` (if any).

juhoinkinen:
> Just started to think, if some some testing could be performed also in [tests/test_swagger.py](https://github.com/NatLibFi/Annif/blob/master/tests/test_swagger.py). I don't remember just what more functionality does the tests in `test_swagger.py` cover than those in `test_rest.py` (if any).

Background to the question of `test_swagger.py` vs `test_rest.py`: https://github.com/NatLibFi/Annif/pull/551#issuecomment-1011063687

osma:
Rebased on PR #724, adapted to use annif.simplemma_util and Connexion 3 support, force-pushed.

juhoinkinen (approved):
I tested this with some quick ways, works nicely!

PR #661, opened 2023-01-05
Order of projects read from projects configuration directory

### LINKED ISSUE #647: Order of projects when using project configuration directory
Since #584/#585 it has been possible to use a project configuration directory (`projects.d` by default) to contain multiple project configuration files.

A problem is that it seems not possible to set the order of projects that come from different files in a projects listing (all projects from one file come before the projects from the other file, no matter how the files are named; at the moment I'm not sure what dictates the order). This is a problem concerning web UI, when one would like to order the projects in a specific way.

I think it would be best that the order of "project blocks" (projects set in different configuration files) would obey the alphabetical order of the file names. That way the order could be set as wished by using e.g. `0-projects-yso.toml` and `1-projects-hogwarts.toml` etc.

juhoinkinen:
This makes the projects read from a projects configuration directory follow the alphanumerical order of the names of the projects configuration files in the directory.

Previously projects from `.cfg` files came before projects from `.toml` files.

Closes #647.


osma (approved):
LGTM!

Issue #579, opened 2022-03-14
Batch suggest operation


osma:
Currently all the `suggest` methods (CLI command, REST API method, project and backend methods) always take just one document at a time. This is inefficient for backends that could process many documents in parallel.

We should introduce a batch version of `suggest` (called e.g. `suggest_batch` or `suggest_many` unless someone has a better idea?) for each of these contexts. Individual backends can then choose to implement it when it gives a performance boost; otherwise, the batch is simply passed to the regular `suggest` method one document at a time. I believe that at least NN ensemble, SVC, fastText and MLLM backends could benefit from parallel suggest operations. Also, this would be very useful for the proposed [XTransformer backend](https://github.com/NatLibFi/Annif/pull/540).

A note on scope: This issue is about implementing the scaffolding necessary for batching suggest operations, as well as using them in at least some (not necessarily all) operations that would benefit from it: e.g. eval, hyperopt, optimize, index. Changes to individual backends are out of scope but separate issues for them should be opened after this basic scaffolding is in place.

osma:
> If a new suggest-batch CLI command is added, from where should the documents be loadable when using it? From a directory (of text documents), or any paths, maybe to already indexed TSV files?
> Alternatively, maybe the existing suggest CLI command could be turned to use batched processing if it is given a path to documents instead of stdin feed?

My hunch would be to try to extend the current `suggest` CLI command instead of defining a new `suggest-batch` command. The current `suggest` command expects input from stdin; maybe we could change it so it works more like the  `cat` command and other similar *nix tools, i.e. it could take one or more filenames as a parameter, but fall back to `stdin` if no file names are given. So you could do e.g.

    annif suggest yso-tfidf-en <document.txt              # just like before
    annif suggest yso-tfidf-en document.txt               # the same, but from a named file
    annif suggest yso-tfidf-en doc1.txt doc2.txt doc3.txt # many files
    annif suggest yso-tfidf-en doc*.txt                   # similar to above, but using shell expansion

> Opt 2: To stdout like the current suggest does, separating documents by first showing the document name and then on the following lines the subject suggestions for the document, e.g.:

I think this is the way to go. For easier grepping etc., I would perhaps add some kind of extra tag in addition to the filename, something like:

```
Suggestions for tests/corpora/archaeology/fulltext/440866.txt
<http://www.yso.fi/onto/yso/p6218>	riimukirjoitus	0.3213897943496704
[...]

PR #663, opened 2023-01-20
Support for batch suggest operations for CLI commands

juhoinkinen:
Adds support for passing multiple documents in a batch from `suggest`, `index`,  `eval` and `optimize` CLI commands to backends (like discussed in issue #579).

Makes `annif suggest` accept  path(s) to file(s) to be indexed, in addition to stdin:
```
annif suggest yso-tfidf-en <document.txt              # just like before
annif suggest yso-tfidf-en document.txt               # the same, but from a named file
annif suggest yso-tfidf-en doc1.txt doc2.txt doc3.txt # many files
annif suggest yso-tfidf-en doc*.txt                   # similar to above, but using shell expansion
annif suggest yso-tfidf-en -                          # stdin with dash
annif suggest yso-tfidf-en doc1.txt -                 # mixing file and stdin with dash
```
The output for each file path input begins with line `Suggestions for <file/path>`:
```
Suggestions for tests/corpora/archaeology/fulltext/440866.txt
<http://www.yso.fi/onto/yso/p6218>	riimukirjoitus	0.3213897943496704
<http://www.yso.fi/onto/yso/p6479>	viikingit	0.18659920990467072
<http://www.yso.fi/onto/yso/p12738>	viikinkiaika	0.18625082075595856
<http://www.yso.fi/onto/yso/p22768>	Kiinan muuri	0.15950888395309448
<http://www.yso.fi/onto/yso/p3973>	antiikki	0.13840530812740326
<http://www.yso.fi/onto/yso/p14588>	riimukivet	0.1362432837486267
<http://www.yso.fi/onto/yso/p14173>	kaivaukset	0.1201547235250473
<http://www.yso.fi/onto/yso/p5713>	hautalöydöt	0.11249098181724548
<http://www.yso.fi/onto/yso/p15031>	viikinkiretket	0.11039584875106812
<http://www.yso.fi/onto/yso/p5714>	muinaishaudat	0.10336380451917648
```

The documents are passed as batches, i.e. lists of texts (lists are generated by a generator also in the case of `suggest` command with files) to the `backend.py` module, which defines the default `_suggest_batch` method that uses the regular, single-text `_suggest` method. This allows the actual backends to define their own `_suggest_batch` methods that can operate on the document batch. 

The implementation of document batching for the `hyperopt` CLI command is left from this PR, as the hyperopt functionality is implemented in individual backends.

 Note that there is no support the actual batch-processing in any backends yet.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -561,8 +561,7 @@ def run_eval(
         for hit_sets, subject_sets in pool.imap_unordered(
             psmap.suggest_batch, corpus.doc_batches(project.DOC_BATCH_SIZE)
         ):
-            for hits, subject_set in zip(hit_sets[project_id], subject_sets):
-                eval_batch.evaluate(hits, subject_set)
+            eval_batch.evaluate_many(hit_sets[project_id], subject_sets)
 
     template = "{0:<30}\t{1}"
     metrics = eval_batch.results(
@@ -608,20 +607,21 @@ def run_optimize(project_id, paths, docs_limit, backend_param):
 
     ndocs = 0
     corpus = open_documents(paths, project.subjects, project.vocab_lang, docs_limit)
-
     for docs_batch in corpus.doc_batches(project.DOC_BATCH_SIZE):
         texts, subject_sets = zip(*[(doc.text, doc.subject_set) for doc in docs_batch])
-
         raw_hit_sets = project.suggest_batch(texts, backend_params)
-        for raw_hits, subject_set in zip(raw_hit_sets, subject_sets):
-            hits = raw_hits.filter(project.subjects, limit=FILTER_BATCH_MAX_LIMIT)
-            assert isinstance(hits, ListSuggestionResult), (
-                "Optimize should only be done with ListSuggestionResult "
-                + "as it would be very slow with VectorSuggestionResult."
-            )
-            for hit_filter, filter_batch in filter_batches.values():
-                filter_batch.evaluate(hit_filter(hits), subject_set)
-            ndocs += 1
+        hit_sets = [
+            raw_hits.filter(project.subjects, limit=FILTER_BATCH_MAX_LIMIT)
+            for raw_hits in raw_hit_sets
+        ]
+        assert isinstance(hit_sets[0], ListSuggestionResult), (
+            "Optimize should only be done with ListSuggestionResult "
+            + "as it would be very slow with VectorSuggestionResult."
+        )
+        for hit_filter, filter_batch in filter_batches.values():
+            filtered_hits = [hit_filter(hits) for hits in hit_sets]
+            filter_batch.evaluate_many(filtered_hits, subject_sets)
+        ndocs += len(texts)
 
     click.echo("\t".join(("Limit", "Thresh.", "Prec.", "Rec.", "F1")))
 
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -84,16 +84,16 @@ def open_documents(paths, subject_index, vocab_lang, docs_limit):
 
 
 def open_text_documents(paths, docs_limit):
-    docs = []
-    for path in paths[:docs_limit]:
-        if path == "-":
-            docs.append(annif.corpus.Document(text=sys.stdin.read(), subject_set=None))
-        else:
-            with open(path, errors="replace", encoding="utf-8-sig") as docfile:
-                docs.append(
-                    annif.corpus.Document(text=docfile.read(), subject_set=None)
-                )
-    return annif.corpus.DocumentList(docs)
+    def _docs(paths):
+        for path in paths:
+            if path == "-":
+                doc = annif.corpus.Document(text=sys.stdin.read(), subject_set=None)
+            else:
+                with open(path, errors="replace", encoding="utf-8-sig") as docfile:
+                    doc = annif.corpus.Document(text=docfile.read(), subject_set=None)
+            yield doc
+
+    return annif.corpus.DocumentList(_docs(paths[:docs_limit]))
 
 
 def show_hits(hits, project, lang, file=None):
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -559,7 +559,7 @@ def run_eval(
 
     with pool_class(jobs) as pool:
         for hit_sets, subject_sets in pool.imap_unordered(
-            psmap.suggest_batch, corpus.doc_batches(project.DOC_BATCH_SIZE)
+            psmap.suggest_batch, corpus.doc_batches
         ):
             eval_batch.evaluate_many(hit_sets[project_id], subject_sets)
 
@@ -607,7 +607,7 @@ def run_optimize(project_id, paths, docs_limit, backend_param):
 
     ndocs = 0
     corpus = open_documents(paths, project.subjects, project.vocab_lang, docs_limit)
-    for docs_batch in corpus.doc_batches(project.DOC_BATCH_SIZE):
+    for docs_batch in corpus.doc_batches:
         texts, subject_sets = zip(*[(doc.text, doc.subject_set) for doc in docs_batch])
         raw_hit_sets = project.suggest_batch(texts, backend_params)
         hit_sets = [
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -7,7 +7,6 @@ import json
 import os.path
 import re
 import sys
-from itertools import islice
 
 import click
 import click_log
@@ -82,6 +81,14 @@ def open_documents(paths, subject_index, vocab_lang, docs_limit):
     if docs_limit is not None:
         docs = annif.corpus.LimitingDocumentCorpus(docs, docs_limit)
     return docs
+
+
+def open_text_documents(paths, docs_limit):
+    docs = []
+    for path in paths[:docs_limit]:
+        with open(path, errors="replace", encoding="utf-8-sig") as docfile:
+            docs.append(annif.corpus.Document(text=docfile.read(), subject_set=None))
+    return annif.corpus.DocumentList(docs)
 
 
 def show_hits(hits, project, lang, file=None):
@@ -388,25 +395,19 @@ def run_suggest(
     hit_filter = SuggestionFilter(project.subjects, limit, threshold)
 
     if sys.stdin.isatty():
-        documents, filenames = open_docs(paths, docs_limit)
-        subject_sets = project.suggest_batch(documents, backend_params)
-        for filename, subjects in zip(filenames, subject_sets):
-            click.echo(f"Suggestions for {filename}")
+        docs = open_text_documents(paths, docs_limit)
+        subject_sets = project.suggest_batch(docs, backend_params)
+        for (
+            subjects,
+            path,
+        ) in zip(subject_sets, paths):
+            click.echo(f"Suggestions for {path}")
             hits = hit_filter(subjects)
             show_hits(hits, project, lang)
     else:
         text = sys.stdin.read()
         hits = hit_filter(project.suggest(text, backend_params))
         show_hits(hits, project, lang)
-
-
-def open_docs(paths, docs_limit):
-    docs, filenames = [], []
-    for path in islice(paths, docs_limit):
-        with open(path, errors="replace", encoding="utf-8-sig") as docfile:
-            filenames.append(path)
-            docs.append(docfile.read())
-    return docs, filenames
 
 
 @cli.command("index")
@@ -442,7 +443,7 @@ def run_index(
     hit_filter = SuggestionFilter(project.subjects, limit, threshold)
 
     documents = annif.corpus.DocumentDirectory(
-        directory, project.subjects, project.vocab_lang, require_subjects=False
+        directory, None, None, require_subjects=False
     )
     subject_sets = project.suggest_batch(documents, backend_params)
 
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -7,6 +7,7 @@ import json
 import os.path
 import re
 import sys
+from itertools import islice
 
 import click
 import click_log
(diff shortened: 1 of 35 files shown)


juhoinkinen:
Some thoughts & questions:
- Using `DocumentList` object made the implementation for `ìndex` command a bit easier (less changes) that using plain list of texts. But actually, instead of using `DocumentList` class there could be a `DocumentBatch` class just for this use, but would that give any benefits. 
- In REST, is it ok to just return a list of document suggestions, or should there be some
(optional) id for each doc for more careful linking of documents in request and subject suggestions in response?
  - [MonkeyLearn classifier](https://monkeylearn.com/api/v3/#classify)  uses (optional) `external_id` field and also returns the original text from the request.
  -  [AWS Comprehend BatchDetectKeyPhrases](https://docs.aws.amazon.com/comprehend/latest/APIReference/API_BatchDetectKeyPhrases.html) responses contain `index` which matches the index of documents in the input list.
- Should there be a limit for number of docs that can be send via REST batch-suggest? Or
is it enough to have the possibility to set a payload limit for the method in e.g. NGINX?
- Maybe `batch-suggest` would be a better name for the method than `suggest-batch`.
- `rest.py` needs refactoring to reduce code duplication, maybe implementing a helper `_suggest` function, which contains the common functionality operating on `DocumentList`, and the existing `suggest` could turn the text input to a `DocumentList` containing only one document.

annif/cli.py
```diff
@@ -345,35 +364,51 @@ def run_learn(project_id, paths, docs_limit, backend_param):
...
-        subj = project.subjects[hit.subject_id]
-        click.echo(
-            "<{}>\t{}\t{}".format(
-                subj.uri,
-                "\t".join(filter(None, (subj.labels[lang], subj.notation))),
-                hit.score,
-            )
-        )
+
+    if click.get_text_stream("stdin").isatty():
```
osma, line 397: I don't think this is the correct test. Whether stdin is connected to a tty (e.g. run from a console session) should be irrelevant. Instead I think that `if paths:` or `if len(paths) > 0:` could work - meaning that if paths were given on the command line those should be used, but if not, fall back to reading stdin.

tests/test_project.py
```diff
@@ -229,6 +229,31 @@ def test_project_suggest(registry):
...
+    assert len(result) == 0
+
+
+def test_project_suggest_batch(registry, fulltext_corpus):
+    project = registry.get_project("dummy-en")
+    result = project.suggest_batch(fulltext_corpus)
+    assert len(result) == 28  # Number of documents
+    fist_doc_hits = result[0].as_list()
+    assert len(fist_doc_hits) == 1
+    assert fist_doc_hits[0].subject_id == project.subjects.by_uri(
```
osma, line 244: Typo here: `fist` should probably be `first`..? (Was this copied from somewhere else with the same typo?)

tests/test_swagger.py
```diff
@@ -49,6 +49,29 @@ def test_swagger_suggest_novocab(app_client):
...
+
+
+def test_swagger_suggest_batch_no_docs(app_client):
+    data = {}
+    req = app_client.post(
+        "http://localhost:8000/v1/projects/dummy-fi/suggest-batch", json=data
+    )
+    assert req.status_code == 400
+
+
```
osma, line 74: It would be good to have tests also for the 404 and 503 error code cases, just like for the `suggest` method above.

annif/project.py
```diff
@@ -215,6 +225,21 @@ def suggest(self, text, backend_params=None):
...
+        """Suggest subjects for the given documents using batches of documents in their
+        operations when possible."""
+        if not self.is_trained:
+            if self.is_trained is None:
+                logger.warning("Could not get train state information.")
+            else:
+                raise NotInitializedException("Project is not trained.")
+        corpus = self.transform.transform_corpus(corpus)
+        logger.debug(
+            f"Suggesting subjects for a batch of {sum(1 for _ in corpus.documents)}"
```
osma, line 238: This can be quite inefficient, as it requires looping through a potentially very large corpus, which also has to be transformed. It would be better to avoid counting the documents, unless that information is essential in the debug message. Also, it could potentially change the behaviour of the suggest operation below, depending on how `corpus.documents` has been implemented.

annif/cli.py
```diff
@@ -83,6 +83,25 @@ def open_doc_path(path, subject_index):
     return docs
 
 
+def open_text_documents(paths, docs_limit):
+    docs = []
+    for path in paths[:docs_limit]:
+        with open(path, errors="replace", encoding="utf-8-sig") as docfile:
```
osma, line 89: I think it would be good to support the file name `"-"` as a special case here, meaning to use stdin.

annif/cli.py
```diff
@@ -345,35 +364,51 @@ def run_learn(project_id, paths, docs_limit, backend_param):
 
 @cli.command("suggest")
 @click.argument("project_id")
+@click.argument("paths", type=click.Path(dir_okay=False, exists=True), nargs=-1)
```
osma, line 367: Thought: Should it be possible to pass a directory (containing .txt documents) as well?

osma (changes requested):
This is a very good starting point for batching. Also, it already provides some new functionality (possibility to pass many documents in `suggest` operations via CLI and REST) so I think it would be good to try to merge this first, even before any actual improved implementations in the backends have been developed.

What worries me here a little is the potentially large size of batches. On the REST side they are naturally a bit limited by request size (and maybe we need some other limit as well?), but on the CLI side, it's not uncommon to use `index` and `eval` on thousands of large documents at once. But I don't think it makes sense to process that many documents in a single operation on the backend level - more likely, processing something like 16 or 32 documents at once (let's call it a minibatch) would already give a performance boost, and any larger minibatch size would probably just increase memory overhead with diminishing returns.

It's good to use DocumentList here on the "outer" level, because it's generator based and thus scales naturally to even huge numbers of documents. But I don't think it should be passed directly to backends. Instead, some intermediate layer (probably `project.py`, or the layer above i.e. `cli.py` and `rest.py`) should chop this up into minibatches which are then passed to the backend methods, maybe just as simple lists of text strings. The results (hit_sets) from the backends would have the same size as the minibatch and these would then be assembled back to a single iterable - a generator would probably be a good idea here too, since the number of documents can be huge.

I gave a few minor detailed comments on the code.

I won't comment on the naming in this round, let's see first how the code evolves.

juhoinkinen:
I reverted the commits about REST API (i.e. adding the `/v1/suggest-batch` method) and cherry-picked them to a new branch [issue579-batch-suggest-operation-rest-api](https://github.com/NatLibFi/Annif/tree/issue579-batch-suggest-operation-rest-api). This PR was already quite large, and might grow further if batching in other CLI commands is implemented.

annif/cli.py
```diff
@@ -408,27 +449,21 @@ def run_index(
...
-        directory, project.subjects, project.vocab_lang, require_subjects=False
-    ):
-        with open(docfilename, encoding="utf-8-sig") as docfile:
-            text = docfile.read()
+    documents = annif.corpus.DocumentDirectory(
+        directory, None, None, require_subjects=False
+    )
+    subject_sets = project.suggest_batch(documents, backend_params)
+
+    for (docfilename, dummy_subjectfn), subjects in zip(documents, subject_sets):
```
osma, line 457: `_` could be used here insted of `dummy_subjectfn`, as `_` is commonly used for throwaway variables

annif/project.py
```diff
@@ -117,6 +119,29 @@ def _suggest_with_backend(self, text, backend_params):
...
+            if not batch:
+                return
+            yield batch
+
+    def _suggest_batch_with_backend(self, corpus, backend_params):
+        if backend_params is None:
+            backend_params = {}
+        beparams = backend_params.get(self.backend.backend_id, {})
+
+        hit_sets = []
```
osma, line 136: Here `hit_sets` can potentially become quite large, for example in an `eval` operation on thousands of documents. Would it be possible to restructure this so that `_suggest_batch_with_backend` (or at least the outer `suggest_batch`) would return a generator instead? This way there would be no need to collect lots of hit_sets in a single list, but instead the calling code would start getting back results early, as soon as the first minibatch is finished. Perhaps [itertools.chain](https://docs.python.org/3/library/itertools.html#itertools.chain) or its sibling chain.from_iterable would be useful here?

osma: Easiest would be to use `yield self.backend.suggest_batch(...)` here in `_suggest_batch_with_backend`, and then use `return itertools.chain(self._suggest_batch_with_backend(...))` in the outer method `suggest_batch`.

juhoinkinen:
I tried to utilize the new [`suggest_batch` function](https://github.com/NatLibFi/Annif/blob/issue579-batch-suggest-operation/annif/project.py#L237) of the project module for the `eval` CLI command, but it does not seem very straightforward.

First, the eval command uses the [imap_unordered](https://github.com/NatLibFi/Annif/blob/issue579-batch-suggest-operation/annif/cli.py#L562), which expects an iterable as the second argument), but `suggest_batch` operates on a corpus object (that has `documents` iterable only as its property). Also, I'm not sure if `imap_unordered` could in any way feed multiple documents to the `suggest_batch` function: it has the `chunksize` parameter, but afaik even when setting that to a non-default value (!=1) it does not send multiple elements from the iterable to the function in one pass.

juhoinkinen:
I added `BatchingDocumentCorpus` to help getting batches of documents, but I wonder could the `doc_batches(batch_size)` be added already to the `DocumentCorpus` base class, batching of documents is used quite many times. (TODO: add testing of the batching method.)

([Black v23.1.0](https://github.com/psf/black/releases/tag/23.1.0) was just released, and it introduced some changes for the style, which raise complaints in the current Annif code. :( )

osma:
> I wonder could the doc_batches(batch_size) method be added already to the DocumentCorpus base class, as batching of documents is used quite many times.

Yes, this sounds like a great idea! I think it would simplify the code and enable batching in many places.

annif/cli.py
```diff
@@ -83,6 +83,30 @@ def open_doc_path(path, subject_index):
     return docs
 
 
+def open_text_documents(paths, docs_limit):
```
osma, line 86: This adds yet another utility function to the top of cli.py. Nothing wrong with that, but cli.py has grown very long so I think we could refactor this in a follow-up PR, moving the utility functions to a separate module such as cli_util.py. Then cli.py would just contain the Click-decorated functions that implement the CLI commands themselves.

annif/project.py
```diff
@@ -42,6 +43,7 @@ class AnnifProject(DatadirMixin):
 
     # default values for configuration settings
     DEFAULT_ACCESS = "public"
+    DOC_BATCH_SIZE = 32
```
osma, line 46: This constant is declared here in AnnifProject, but its only uses are to pass it as a parameter to corpus.doc_batches() and 2 out of 3 of these are in cli.py. Would it be better to move this constant to the DocumentCorpus class and simply use the constant value within doc_batches instead of passing it around as a parameter?

The alternative would be to make this configurable per project (just like vocab, language etc.), but I'm not sure if that would add any value.

osma (changes requested):
I already gave a comment about moving DOC_BATCH_SIZE inside DocumentCorpus since it's only needed there.

Apart from that, I think there's an opportunity for further simplification. Basically we don't need the single-text versions of suggest methods anymore, as they can be handled as a special case of suggest_batch. This may seem a bit radical but I think it's not very hard (well, except maybe fixing up all the tests):

1. The old AnnifProject.suggest() method is now needed only in two places, in the cli.py run_suggest function (the stdin-only case) and in the rest.py suggest function. Convert these two to use the AnnifProject.suggest_batch() method instead - passing a list with one text.
2. Remove the now unused AnnifProject.suggest() method as well as AnnifProject._suggest_with_backend() that it relies on (but nothing else needs it)
3. Remove the now unused AnnifBackend.suggest() method.

It may also make sense to rename the remaining suggest_batch() methods to just suggest(), now that the shorter name is available. Though we still need both AnnifBackend._suggest and AnnifBackend._suggest_batch, because it's up to backends which variant they will implement.

osma:
Oh by the way, I tested the `annif eval` command (yso-mllm-fi project, evaluated on kirjastonhoitaja test set) with the `-j 4` option. The evaluation results were the same. It took a few seconds longer, probably because the parallel processing is done on larger batches and the final batches cannot be distributed between CPUs. But I think that's not too bad, and we should see performance gains in the future as we implement support for batching in individual backends.

juhoinkinen:
Now some debug-level logs from project module are removed:
Before:
```
annif suggest tfidf-fi <kissa.txt -v DEBUG
debug: creating app with configuration annif.default_config.Config
debug: Reading configuration file projects.cfg in CFG format
debug: loading subjects from data/vocabs/yso/subjects.csv
debug: Suggesting subjects for text "kissa
debug: ..." (len=6)
debug: Backend tfidf: loading vectorizer from data/projects/tfidf-fi/vectorizer
debug: Backend tfidf: loading similarity index from data/projects/tfidf-fi/tfidf-index
debug: Backend tfidf: Suggesting subjects for text "kissa
debug: ..." (len=6)
debug: Got 100 hits from backend tfidf
debug: 100 hits from backend
<http://www.yso.fi/onto/yso/p19378>	kissa	0.9530429244041443
<http://www.yso.fi/onto/yso/p864>	kissaeläimet	0.5541983842849731
```

Now:
```
annif suggest tfidf-fi <kissa.txt -v DEBUG
debug: creating app with configuration annif.default_config.Config
debug: Reading configuration file projects.cfg in CFG format
debug: loading subjects from data/vocabs/yso/subjects.csv
debug: Backend tfidf: loading vectorizer from data/projects/tfidf-fi/vectorizer
debug: Backend tfidf: loading similarity index from data/projects/tfidf-fi/tfidf-index
debug: Backend tfidf: Suggesting subjects for text "kissa
debug: ..." (len=6)
<http://www.yso.fi/onto/yso/p19378>	kissa	0.9530429244041443
<http://www.yso.fi/onto/yso/p864>	kissaeläimet	0.5541983842849731
```

And in the case giving multiple files to `annif suggest`, the debug log has some duplication:
```
annif suggest tfidf-fi *.txt -v DEBUG
debug: creating app with configuration annif.default_config.Config
debug: Reading configuration file projects.cfg in CFG format
debug: loading subjects from data/vocabs/yso/subjects.csv
debug: Backend tfidf: loading vectorizer from data/projects/tfidf-fi/vectorizer
debug: Backend tfidf: loading similarity index from data/projects/tfidf-fi/tfidf-index
debug: Backend tfidf: Suggesting subjects for text "kissa
debug:  cat
debug: dog
debug: viiki..." (len=24)
debug: Backend tfidf: Suggesting subjects for text "kissa
debug: ..." (len=6)
debug: Backend tfidf: Suggesting subjects for text "koira
debug: ..." (len=6)
debug: Backend tfidf: Suggesting subjects for text "laiva
debug: ..." (len=6)
debug: Backend tfidf: Suggesting subjects for text "viikinki
debug: ..." (len=9)
```

osma:
> Now some debug-level logs from project module are removed

Right, because the debug information was printed only in the single text `suggest` methods. I don't think this is a big loss.

> And in the case giving multiple files to annif suggest, the debug log has some duplication:

I don't see any exact duplicates - the messages are related to different input files, right?

osma (approved):
LGTM!

Maybe the PR title could be amended, as it also covers the `eval` command now (but only the CLI side)

PR #669, opened 2023-02-03
Batch suggest in Omikuji backend

osma:
This PR clarifies that backends only have to implement either one of `_suggest` and `_suggest_batch`, then implements batched suggest in the Omikuji backend. In practice, only the text vectorization is performed on the whole batch at once; the Omikuji implementation only supports a [predict method](https://github.com/tomtung/omikuji/blob/master/python-wrapper/omikuji/__init__.py#L106) for a single document at a time so it has to be done within a `for` loop.

There seems to be a small performance benefit. I tested this using `annif eval` the Finto AI yso-parabel-fi project/model, with the kirjaesittelyt2021/fin/test corpus. The evaluation results were unchanged, only the amount of time spent was slightly different. Memory usage remained pretty much the same.

# With 1 job

|                 | user time | wall time | max rss |
| --------------- | --------- | --------- | ------- |
| before (master) | 86.69     | 1:29.94   | 6322624 |
| after (PR)      | 78.66     | 1:22.79   | 6336584 |

# With 4 jobs

|                 | user time | wall time | max rss |
| --------------- | --------- | --------- | ------- |
| before (master) | 121.33    | 1:22.33   | 6293804 |
| after (PR)      | 96.55     | 1:18.78   | 6293640 |

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,7 +157,12 @@ A backend can define these key fields and methods:
 * `name`: field for a name for the backend (a single word, all lowercase)
 * `initialize` (optional): method setting up the necessary internal data structures
 * `_train` (optional): method for training the model on a given document corpus
-* `_suggest`: method for feeding a single document (text) and getting suggested subjects
+* `_suggest`: method for feeding a single document (text) and getting suggested subjects for it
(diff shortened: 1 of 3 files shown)


juhoinkinen (approved):
LGTM

Issue #451, opened 2020-11-10
Need for universal log message verbosity


AhtiAhde:
There are big differences between log message verbosity of different models.

This problem is probably caused due to the fact that annif doesn't filter the messages in any way, but merely directly passes everything the original process produces (at least this seems to be the case for Omikuji).

In production environment this can cause at least two distinct problems:
-Upon container start, the Maui model is mostly silent, but Omikuji floods all kinds of messages; while this is not dangerous it can look daunting for someone new to the project and sometimes it is easy to forget while debugging other bugs, which can cause doubts targeting to wrong candidates
-When training the models, Omikuji floods warnings that depend on the size of the data set. This can cause the training to fail to exceeding the log size, rather than real error or continuing regardless of warnings; it would be better that these warnings would be aggregated and an explicit "maximum allowed warnings" might generate an error rather than the exceeding of logs.

There are probably more problems, but in general it might be nice that annif would provide some universal log messages / steps of processing, that would be easy and coherent to follow and then give the user ability to add deeper levels of logging, which would allow more streamlined access to the original warnings etc. provided by the underlying models.

juhoinkinen:
> - Upon container start, the Maui model is mostly silent, but Omikuji floods all kinds of messages; while this is not dangerous it can look daunting for someone new to the project and sometimes it is easy to forget while debugging other bugs, which can cause doubts targeting to wrong candidates

Related to this: when Annif is starting, especially in a container environment when one is just looking a log stream, there is no very clear indication that all models are already loaded and Annif is ready to serve requests. It could happen that the Pod or container for Annif is killed for some reason, and I think this is not visible in the logs. 

There could be an info message from Annif like `Model(s) loaded.` when all models are ready, if not a message for every individual model.

osma:
> There could be an info message from Annif like Model(s) loaded. when all models are ready, if not a message for every individual model.

...and now we have an implementation of this in PR #671 :smile:

juhoinkinen:
A step toward this was implemented in #721 for TensorFlow logs.

PR #671, opened 2023-02-07
Add log message indicating finishing projects initialization

juhoinkinen:
By looking at the logs from e.g. OpenShift environment it can be hard to see if Annif pod/container has booted or models are still being loaded. In error situations while Annif is in boot loop (or it seems so) it would be good to easily know from the logs if/when Annif has started.

This PR adds an info level message `finished initializing projects` to log stream for addressing the mentioned concern. For this, it was necessary to also do the [basic configuration](https://docs.python.org/dev/library/logging.html#logging.basicConfig) for the root logger and set log level to "info" (was "unset") for "annif" logger. This makes all log messages from Annif to show also the level and the logger name.

For example:
```
[2023-02-07 16:12:22 +0200] [460823] [INFO] Starting gunicorn 20.1.0
[2023-02-07 16:12:22 +0200] [460823] [INFO] Listening at: http://0.0.0.0:8000 (460823)
[2023-02-07 16:12:22 +0200] [460823] [INFO] Using worker: sync
[2023-02-07 16:12:22 +0200] [460825] [INFO] Booting worker with pid: 460825
2023-02-07T14:12:26.030Z INFO [omikuji::model] Loading model from data/projects/omikuji-parabel-fi/omikuji-model...
2023-02-07T14:12:26.030Z INFO [omikuji::model] Loading model settings from data/projects/omikuji-parabel-fi/omikuji-model/settings.json...
2023-02-07T14:12:26.030Z INFO [omikuji::model] Loaded model settings Settings { n_features: 329277, classifier_loss_type: Hinge }...
2023-02-07T14:12:26.030Z INFO [omikuji::model] Loading tree from data/projects/omikuji-parabel-fi/omikuji-model/tree2.cbor...
2023-02-07T14:12:28.830Z INFO [omikuji::model] Loading tree from data/projects/omikuji-parabel-fi/omikuji-model/tree0.cbor...
2023-02-07T14:12:31.534Z INFO [omikuji::model] Loading tree from data/projects/omikuji-parabel-fi/omikuji-model/tree1.cbor...
2023-02-07T14:12:34.281Z INFO [omikuji::model] Loaded model with 3 trees; it took 8.25s
WARNING:annif:Couldn't initialize backend 'omikuji': vectorizer file 'data/projects/omikuji-parabel-sv/vectorizer' not found
WARNING:annif:Couldn't initialize backend 'omikuji': vectorizer file 'data/projects/omikuji-parabel-en/vectorizer' not found
INFO:annif:finished initializing projects

```
The gunicorn and Omikuji log messages have also timestamps, but they are actually more distracting than helpful when logs are read in Kibana (especially), because there is a dedicated timestamp field for every message.

### DIFF
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -7,7 +7,9 @@ import os.path
 import connexion
 from flask_cors import CORS
 
+logging.basicConfig()
 logger = logging.getLogger("annif")
+logger.setLevel(level=logging.INFO)
 
 import annif.backend  # noqa
 
@@ -36,6 +38,7 @@ def create_app(config_name=None):
 
     if cxapp.app.config["INITIALIZE_PROJECTS"]:
         annif.registry.initialize_projects(cxapp.app)
+        logger.info("finished initializing projects")
 
     # register the views via blueprints
     from annif.views import bp
--- a/tests/test_backend_fasttext.py
+++ b/tests/test_backend_fasttext.py
@@ -104,7 +104,7 @@ def test_train_fasttext_params(document_corpus, project, caplog):
     )
     params = {"dim": 1, "lr": 42.1, "epoch": 0}
 
-    with caplog.at_level(logging.DEBUG):
+    with caplog.at_level(logging.DEBUG, logger="annif"):
         fasttext.train(document_corpus, params)
     parameters_heading = "Backend fasttext: Model parameters:"
     assert parameters_heading in caplog.text


osma (approved):
Seems like a useful feature. :+1: 

As you noted, the different log messages are in different formats and some of them are less useful than others. We also have an old issue #451 which is somewhat related to the problem of handling log messages. IIRC Omikuji does its own logging from within Rust code, and it can be difficult or impossible to control that. Not sure about id and how gunicorn logging could be controlled.

Issue #308, opened 2019-08-08
Encoding UTF-8 error on Windows


hekl:
I get this error when running my training file.

UnicodeDecodeError: 'CP_UTF8' codec can't decode byte 0x88 in position 11: No mapping for the Unicode character exists in the target code page. 

I have produced a set of txt files en tsv files, both encoded with utf-8, like this:
`with open (file,'w+',encoding="utf-8")`. I have produced one archive file from both the txt and tsv files in zip or 7z format.  
I also tried it with no explicit encoding and with utf-16 encoding, running into the same encoding problems.
I am using Windows 10 and python.

osma:
Can you please show the file you've produced that gave you this error? The error appears to be ealy on in the file (position 11) so just the first few lines should be enough.

Also show the command you executed and the full traceback.

0x88 is typically used in UTF-8 escape sequences.

hekl:
`(annif-latest) D:\Programs\annif-latest>annif train tfidf-nl Annif-corpora\training\train-utf8.zip
creating vectorizer
Traceback (most recent call last):
  File "C:\Program Files\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Program Files\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\Programs\annif-latest\Scripts\annif.exe\__main__.py", line 9, in <module>
  File "d:\programs\annif-latest\lib\site-packages\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "d:\programs\annif-latest\lib\site-packages\flask\cli.py", line 586, in main
    return super(FlaskGroup, self).main(*args, **kwargs)
  File "d:\programs\annif-latest\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "d:\programs\annif-latest\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "d:\programs\annif-latest\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "d:\programs\annif-latest\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "d:\programs\annif-latest\lib\site-packages\click\decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
[...]

PR #672, opened 2023-02-09
Fix slow CI/CD runs for Python 3.10

juhoinkinen:
Lately the Python 3.10 test CI/CD jobs have been slow (~7 mins), because the Python dependencies have not been available from cache. First occurrence is in the [CI/CD job #308](https://github.com/NatLibFi/Annif/actions/runs/3875121566/jobs/6607220782), however it seems that the reason for the slow-down is introducing the linting job in PR #656, because this job has shared the cache keys with the Python 3.10 test job. This has prevented (re)creating/updating the cache by the test job (with the necessary optional dependencies); the "Post" steps for setup-python and caching steps in the test jobs have had errors like:
```
Failed to save: Unable to reserve cache with key setup-python-Linux-python-3.10.9-poetry-6025e512699131453ed16a5ae9a4be9139ae85329e34f47f750e6fb7af36d308, another job may be creating this cache. More details: Cache already exists.
```

~~To avoid this in the caching step the cache key is appended with `-${{ github.job }}` so the keys are now clearly different in lint and test jobs (`poetry-installation-and-repos-3.10.8-1.2.0-lint` vs. `poetry-installation-and-repos-3.10-1.2.0-test`). In the setup-python action the key cannot be specified, so the fix (for now) is to to pin the Python version to 3.10.8 in the linting job, but in the test job to more relaxed 3.10, which makes the keys different by the patch version (`setup-python-Linux-python-3.10.8-poetry-<hash>` vs. `setup-python-Linux-python-3.10.9-poetry-<hash>`).~~

Now restoring and saving cache is more controlled by using separate [cache/restore](https://github.com/actions/cache/tree/main/restore) and [cache/save](https://github.com/actions/cache/tree/main/save) actions: restoring is performed before every job, but saving only after the test jobs that have various optional dependencies installed.

Also now the cache should be updated/recreated if some dependency has been updated, as the cache key includes hash of the `poetry.lock`. The cache restoration is based on "closest matching" key by using [`restore-keys`](https://github.com/actions/cache/blob/main/caching-strategies.md#using-restore-keys-to-download-the-closest-matching-cache) (the plain `key` is not used at all).

The cache/save step is conditional: it is skipped if there are no changes in dependencies.
        
To make the workflow setup more manageable, the caching, Poetry install and Python setup steps are gathered in one composite action in `.github/actions/prepare/action.yml` file.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -1,0 +1,28 @@
+name: Prepare
+description:
+inputs:
+  python-version:
+    required: true
+  poetry-version:
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Load cached Poetry installation and its repo info & venvs
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cache/pipx/venvs
+          ~/.local/bin
+          ~/.cache/pypoetry/{cache/repositories,virtualenvs}
+        key: poetry-installation-and-repos-${{ inputs.python-version }}-${{ inputs.poetry-version }}-${{ github.job }}
+    - name: Install Poetry
+      shell: bash
+      run: |
+        pipx install poetry==${{ inputs.poetry-version }}
+    - name: Set up Python ${{ inputs.python-version }}
+      uses: actions/setup-python@5ccb29d8773c3f3f653e1705f474dfaa8a06a912  # v4.4.0
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: 'poetry'
+        cache-dependency-path: 'pyproject.toml'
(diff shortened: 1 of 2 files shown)


juhoinkinen:
Back to draft, there should be a better way to fix this.

juhoinkinen:
Force-pushed to drop very many unsuccessful commits trying to make caching step conditional, with lines like
           
    if: steps.prepare.outputs.cache-matched-key != format('poetry-installation-and-cache-{0}-{1}-{2}', ${{ matrix.python-version }}, ${{ env.POETRY_VERSION }}, ${{ hashFiles('**/poetry.lock') }})

PR #677, opened 2023-03-03
 Batch suggest in ensemble backends

### LINKED ISSUE #674: Memory leak in NN ensemble backend
The Annif pod in the OpenShift environment has been occasionally killed, maybe once in every two weeks. The reason has seemed to be that the memory consumption has reached the limit for the pod (30 GB).

I monitored the memory consumption (RssAnon from `/proc/$PID/status`) of locally run Annif when running suggest requests to NN ensemble project and its base projects with curl (using fulltext documents from JYU test set; the memory consumption was recorded after every 10 documents), but only in the case of NN ensemble there was a strong increase in the memory consumption: see below for a run for the yso-fi model of Finto AI.

image

I confirmed that the issue could be fixed by following the advice from [one relevant discussion](https://github.com/tensorflow/tensorflow/issues/44711#issuecomment-1282054595), i.e. to use `__call__()` of the model: 

https://github.com/NatLibFi/Annif/blob/73d4f2e4ad2d702709af27e92319d4e945a8d019/annif/backend/nn_ensemble.py#L141

The other mentioned fix, that is applying `tf.convert_to_tensor()`, did not fix the memory leak; running (also) `gc.collect()` after each prediction did fix it, but the predictions become very slow (10 requests took ~110 s, when without gc only ~30 s).

However, the NN ensemble could be modified to [allow batch processing of the documents](https://github.com/NatLibFi/Annif/issues/579), and for that use the [Keras documentation seems to suggest using the `predict()` function](https://keras.io/getting_started/faq/#whats-the-difference-between-model-methods-predict-and-call), so I'm not sure if the mentioned fix is the best way to go.

juhoinkinen:
This adds the `_suggest_batch` method to ensemble backends. 

In all ensemble backends (simple, PAV, NN) the suggestions from the source projects are fetched by using batched suggest calls (this is on project level, so whether a backend actually uses batched suggest depends on the backend).

In case of NN ensemble the prediction (now via `_merge_hit_sets_from_sources`) is performed on the whole batch of the base suggestions in one call of the NN model. The prediction is made using the model's `__call__()` method instead of `predict()` as it is the [recommended way for "small numbers of inputs that fit in one batch"](https://www.tensorflow.org/api_docs/python/tf/keras/Model#predict) and it should offer better performance; this should also fix #674.

Also the [EnsembleOptimizer is made to use batched suggest](https://github.com/NatLibFi/Annif/pull/677/commits/5e392564edd179355d61039cf7e86cecfbe4f7b2). Quickly testing the `hyperopt` command on an ensemble project gives very similar weights and best NDCG scores before and after this PR.

This PR improves somewhat the performance of NN ensemble suggest functionality while the results remain (only nearly?) identical (I think there were small differences in suggestion scores on my laptop). I think most of the increase comes from using `__call__` of the model instead of `predict`.

The below results are from runs at kj-kk using the current Finto AI YSO NN ensemble model (having MLLM, fastText and Omikuji base projects).
# suggest
Targeting 6 times `tests/corpora/archaeology/fulltext/*.txt`:

|                 | user time | wall time | max rss |
| --------------- | --------- | --------- | ------- |
| before (master) | 152.59  |2:35.10   | 13755484 |
| after (PR)      |  135.19  |  2:17.27 | 13757912 |

# eval
Targeting 200 documents from `kirjaesittelyt2021/yso/fin/test`:
## With 1 job
|                 | user time | wall time | max rss |F1@5 |
| --------------- | --------- | --------- | ------- |------- |
| before (master) |  174.79   |2:57.70  | 13816688 | 0.4431|
| after (PR)      | 156.12 |  2:36.77 | 13796932 |0.4431 |

## With 4 jobs
|                 | user time | wall time | max rss |F1@5 |
| --------------- | --------- | --------- | ------- |------- |
| before (master) |  188.21  | 1:45.61 | 13639148 |0.4431 |
| after (PR)      |  172.5   | 1:43.78  | 13606508 |  0.4431|

### DIFF
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -103,8 +103,16 @@ class EnsembleOptimizer(hyperopt.HyperparameterOptimizer):
                 self._source_hits.extend(self._hit_sets_to_list(hit_sets))
 
     def _hit_sets_to_list(self, hit_sets):
-        """Convert a dict of lists of hits to a list of dicts of hits"""
-        return [dict(zip(hit_sets.keys(), hit)) for hit in zip(*hit_sets.values())]
+        """Convert a dict of lists of hits to a list of dicts of hits, e.g.
+        {"proj-1": [p-1-doc-1-hits, p-1-doc-2-hits]
+         "proj-2": [p-2-doc-1-hits, p-2-doc-2-hits]}
+        to
+        [{"proj-1": p-1-doc-1-hits, "proj-2": p-2-doc-1-hits},
+         {"proj-1": p-1-doc-2-hits, "proj-2": p-2-doc-2-hits}]
+        """
+        return [
+            dict(zip(hit_sets.keys(), doc_hits)) for doc_hits in zip(*hit_sets.values())
+        ]
 
     def _normalize(self, hps):
         total = sum(hps.values())
--- a/annif/backend/ensemble.py
+++ b/annif/backend/ensemble.py
@@ -33,35 +33,36 @@ class BaseEnsembleBackend(backend.AnnifBackend):
         by subclasses."""
         return hits
 
-    def _suggest_with_sources(self, text, sources):
-        hits_from_sources = []
+    def _suggest_with_sources(self, texts, sources):
+        hit_sets_from_sources = []
         for project_id, weight in sources:
             source_project = self.project.registry.get_project(project_id)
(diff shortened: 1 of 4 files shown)


juhoinkinen:
I rerun the NN ensemble suggest and eval timings and updated the results table after the last change: it seemed to give a few seconds improvement more.

PR #682, opened 2023-03-14
Use Schemathesis to test REST API

juhoinkinen:
For testing the REST API there have been manually written tests since abandoning the unmaintained swagger-tester in #551. Already at that time Schemathesis was considered as a replacement of swagger-tester, but there was a [dependency conflict](https://github.com/NatLibFi/Annif/pull/551#issuecomment-1016506375) that prevented using it.

Now Schemathesis can be installed to restore testing that the API actually conforms to the OpenAPI specification. Also the specification will now be validated.

With the default settings Schemathesis takes 12-15 s to run the tests in `tests/test_swagger.py`, but using `@settings(max_examples=10)`  only 1-2 s.

~Due to a [bug](https://github.com/schemathesis/schemathesis/issues/1718) the pyproject.yaml includes pinned starlette installation (`starlette = ">=0.13,<0.21"`).~ Edit: Pinning starlette is not necessary when pinning to recent version Schemathesis `3.18.*`.

When fixing the errors noted by Schemathesis I marked some properties in the specification as nullable. `is_trained` and `modification_time` should be nullable, but `backend_id` probably should not be:
https://github.com/NatLibFi/Annif/blob/6fc315763a395dff86420f3d1f07e3b3fe3ca869/annif/openapi/annif.yaml#L173
 I marked this only to make the test pass. One project for the testing config is deliberately missing the backend specification:
https://github.com/NatLibFi/Annif/blob/6fc315763a395dff86420f3d1f07e3b3fe3ca869/tests/projects.cfg#L70-L74

I'm not really sure what to do for that.


pyproject.toml
```diff
@@ -70,6 +70,8 @@ flake8 = "*"
 bumpversion = "*"
 black = "23.*"
 isort = "*"
+schemathesis = "*"
+starlette = ">=0.13,<0.21"
```
osma, line 74: It would be nice to refer here to the schemathesis bug that necessitated this

tests/test_swagger.py
```diff
@@ -1,74 +1,12 @@
 """Unit tests for Annif REST API / Swagger spec"""
+import schemathesis
+from hypothesis import settings
 
+schema = schemathesis.from_path("annif/openapi/annif.yaml")
 
-def test_swagger_cors(app_client):
```
osma, line 4: Does schemathesis replace this test, and all other tests below? Or should we also keep these tests?

Also, I think we should rename `test_swagger.py` to something else, maybe `test_openapi.py`, because we're not actually using Swagger (2.0) anymore. I just forgot to rename this in the OpenAPI 3.0 PR #649

osma:
> is_trained and modification_time should be nullable, but backend_id probably should not be:

Agree, backend_id should not be nullable. We should ensure that the API never returns data like that.

pyproject.toml
juhoinkinen: Installing (pinned) Starlette is not required when pinning Schemathesis version to `3.18.*`, which is probably a good idea because Schemathesis is being developed quite rapidly.

tests/test_swagger.py
juhoinkinen: I added a custom check that is run in addition to the default Schemathesis tests to cover checking CORS support:
https://github.com/NatLibFi/Annif/blob/08bf4ce3afd074c24dfcb924fde08684b77f536e/tests/test_openapi.py#L8-L10


[Schemathesis validates the following on the responses](https://schemathesis.readthedocs.io/en/stable/python.html#how-are-responses-checked):


> - not_a_server_error. The response has 5xx HTTP status;
> - status_code_conformance. The response status is not defined in the API schema;
> - content_type_conformance. The response content type is not defined in the API schema;
> - response_schema_conformance. The response content does not conform to the schema defined for this specific response;
> - response_headers_conformance. The response headers does not contain all defined headers.

If I make methods in `rest.py` to return nonsense strings, Schemathesis tests fail:
```
tests/test_openapi.py::test_api[GET /v1/] FAILED                                                                                                                [ 20%]
tests/test_openapi.py::test_api[GET /v1/projects] FAILED                                                                                                        [ 40%]
tests/test_openapi.py::test_api[GET /v1/projects/{project_id}] FAILED                                                                                           [ 60%]
tests/test_openapi.py::test_api[POST /v1/projects/{project_id}/suggest] FAILED                                                                                  [ 80%]
tests/test_openapi.py::test_api[POST /v1/projects/{project_id}/learn] FAILED                                                                                    [100%]
```


If I make error cases to return code `204` instead of `404/503/400` some tests fail:
```
tests/test_openapi.py::test_api[GET /v1/] PASSED                                                                                                                [ 20%]
tests/test_openapi.py::test_api[GET /v1/projects] PASSED                                                                                                        [ 40%]
tests/test_openapi.py::test_api[GET /v1/projects/{project_id}] FAILED                                                                                           [ 60%]
tests/test_openapi.py::test_api[POST /v1/projects/{project_id}/suggest] FAILED                                                                                  [ 80%]
tests/test_openapi.py::test_api[POST /v1/projects/{project_id}/learn] PASSED                                                                                    [100%]
```
(`/learn` method has 204 in the schema, so testing that is passed.)

Both cases are covered also by `test_rest.py`, so I would not keep the old tests here.

juhoinkinen:
> > is_trained and modification_time should be nullable, but backend_id probably should not be:
> 
> Agree, backend_id should not be nullable. We should ensure that the API never returns data like that.

I marked `nobackend` project as private, which I think is enough for this PR.

PR #683, opened 2023-03-16
Update dependencies v0.61

juhoinkinen:
This updates some outdated dependencies:
- [Scikit-learn 1.2.2](https://github.com/NatLibFi/Annif/pull/683/commits/5486db72b0261429113e2a3c82abe1d9cb081e2a)
- [Spacy 3.5.*](https://github.com/NatLibFi/Annif/pull/683/commits/9586d73bb579499b21af821580a8194cea3fb56d)
- [Schemathesis 3.19.*](https://github.com/NatLibFi/Annif/pull/683/commits/1c856327e6fabeb1131f55583e5e9a888917056d)
- [lmdb 1.4.1](https://github.com/NatLibFi/Annif/pull/683/commits/7d2e98d0f2838544750c8d3ea11bbed918d009b7)

Also [pins (at least) minor versions of all non-dev dependencies](https://github.com/NatLibFi/Annif/pull/683/commits/04d1461112017b68611c908eac2449a5b8f8dd39) to the versions currently being installed and removes some unnecessary dependencies ([swagger_ui_bundle](https://github.com/NatLibFi/Annif/pull/683/commits/dba4ff4c51e35d137f327e058d0c4061f66169c8), [codecov](https://github.com/NatLibFi/Annif/pull/683/commits/9244967f1ea30b66ccd7124894eb7eea5acb7e38)).


juhoinkinen:
I think [`swagger_ui_bundle = "*"`](https://github.com/NatLibFi/Annif/blob/9586d73bb579499b21af821580a8194cea3fb56d/pyproject.toml#L34) is nowdays not needed as a direct dependency and could be removed.

The dependency was added in #215 due to a problem discussed in https://github.com/spec-first/connexion/issues/406#issuecomment-440633705, which seems to be fixed now.

osma:
Would it make sense to pin major/minor versions, at least for all non-dev dependencies? At least `nltk` and `gunicorn` are not pinned currently and this could cause sudden API breakage.

juhoinkinen:
[TensorFlow 2.12.0](https://github.com/tensorflow/tensorflow/releases/tag/v2.12.0) has some breaking changes:

> tf.keras:
>
>    Moved all saving-related utilities to a new namespace, keras.saving, for example: keras.saving.load_model, keras.saving.save_model, keras.saving.custom_object_scope, keras.saving.get_custom_objects, keras.saving.register_keras_serializable,keras.saving.get_registered_name and keras.saving.get_registered_object. The previous API locations (in keras.utils and keras.models) will be available indefinitely, but we recommend you update your code to point to the new API locations.

Annif imports  `load_model`, which could or should be updated:
https://github.com/NatLibFi/Annif/blob/33bc561187a453cf1653bda8e041e516fbb34903/annif/backend/nn_ensemble.py#L15

And also `atomic_save` apparently uses the old location.

But now I see there is a conflict in installing the new TF version and numpy:
```
Because tensorflow-cpu (2.12.0) depends on numpy (>=1.22,<1.24)
 and no versions of tensorflow-cpu match >2.12.0,<2.13.0, tensorflow-cpu (>=2.12.0,<2.13.0) requires numpy (>=1.22,<1.24).
So, because annif depends on both numpy (1.24.*) and tensorflow-cpu (2.12.*), version solving failed.
```

osma (approved):
LGTM

Issue #684, opened 2023-03-20
CLI command completions


juhoinkinen:
[Click provides support for tab completion in different shells.](https://click.palletsprojects.com/en/8.1.x/shell-completion/) In the case of Annif, Click documentation instructs to use this to register the autocompletion function (and to add it to `~/.bashrc`):

    eval "$(_ANNIF_COMPLETE=bash_source annif)"

We could add this to installation instructions. Or even add an Annif command to generate the function, like in [`helm` CLI](https://helm.sh/docs/helm/helm_completion_bash/).

The completion is a bit laggy but usable. Apparently the `cli.py` module is run for each tab press, and the imports there take most of the time.

The default completions work for "command names, option names, and values for choice, file, and path parameter types". In the case of Annif it would be nice to have completions also for project-ids, which could be accomplished using a [class with `shell_complete()` method](https://click.palletsprojects.com/en/8.1.x/shell-completion/#custom-type-completion) or just a [helper function](https://click.palletsprojects.com/en/8.1.x/shell-completion/#overriding-value-completion): in either case the method/function would need to read the project ids from the configuration file, but (preferably) not via `registry.get_projects()` because then imports of backend dependencies would make the completion unnecessarily slow.

osma:
Sounds good, and this is another reason why optimizing startup performance is important (which generally means using lazy imports of some kind). Some of this has already been done (#514, #544, #563) but I think another round of optimization should be done after the refactoring in #681.

juhoinkinen:
Closed by the linked issue.

PR #693, opened 2023-04-19
Support for CLI command completions

juhoinkinen:
Adds support for tab-key completions for commands, options and project id and vocabulary id parameters via [Click mechanism](https://click.palletsprojects.com/en/8.1.x/shell-completion/#enabling-completion).

A user needs to enable the support as instructed in README.md, i.e. by sourcing the completion script, either manually for every terminal session or automatically for all new sessions by the sourcing the script in a shell startup file (instructed to use `.bashrc`). 

To produce the completion script the `annif completion` CLI command is added. This seems to be the standard way that many CLI tools use:  [pip](https://pip.pypa.io/en/stable/user_guide/#command-completion), [dvc](https://dvc.org/doc/install/completion?tab=Bash-Debian-Ubuntu), [helm](https://helm.sh/docs/helm/helm_completion_bash/).

Implements #684.


annif/cli.py
```diff
@@ -551,5 +551,21 @@ def run_hyperopt(project_id, paths, docs_limit, trials, jobs, metric, results_fi
...
+@click.option("--fish", "shell", flag_value="fish")
+def completion(shell):
+    """Generate autocompletion script for the given shell."""
+
+    if shell is None:
+        click.echo("Shell not given, try --bash, --zsh or --fish")
+        return
+
+    script = os.popen(f"_ANNIF_COMPLETE={shell}_source annif").read()
+    click.echo("# Generated by Annif")
```
osma, line 566: How about adding the Annif version number to the comment? Later versions may have more commands etc.

annif/cli.py
```diff
@@ -56,7 +56,7 @@ def run_list_projects():
 
 
 @cli.command("show-project")
-@click.argument("project_id")
+@click.argument("project_id", shell_complete=cli_util.complete_project_id)
```
osma, line 59: How about defining a cli_util.project_id decorator instead of repeating this many times?

juhoinkinen:
The completion response is not very instantaneous but quite usable. On my laptop `annif --help` and `annif --version` take 0.6-0.7 s, but I think the completion response is a bit faster.

I tried to optimize startup speed by moving some import statements into functions in `cli.py` and elsewhere. This cut `annif --version` time in half but did not affect `annif --help` and seemed not to affect completion response time. :(

juhoinkinen:
> The completion response is not very instantaneous but quite usable. On my laptop `annif --help` and `annif --version` take 0.6-0.7 s, but I think the completion response is a bit faster.
> 
> I tried to optimize startup speed by moving some import statements into functions in `cli.py` and elsewhere. This cut `annif --version` time in half but did not affect `annif --help` and seemed not to affect completion response time. :(

The response time of completions will be improved by #696 to ~0.3 seconds or slightly less.

PR #696, opened 2023-04-24
Optimize CLI startup time

juhoinkinen:
Optimizes CLI startup time by avoiding unnecessary imports for basic commands, mainly to improve the CLI command completion response times in #693.

Changes: 

- Imports of joblib and numpy are simply moved inside the functions/methods where they are used.

- Import of connexion is avoided for CLI commands (all but `annif run`) by using separate creation scripts for plain-Flask and Connexion-Flask applications. When started via `gunicorn "annif:create_app()` the Connexion-Flask app is created as previously.

This is a continuation of optimizations in #544.

I checked the import time of the [pre-release of  connexion 3](https://github.com/spec-first/connexion/releases/tag/3.0.0a5), and it was even slower (~0.35 s) than of the current v2.14.* (0.12 s), so avoiding connexion import is even more valuable in the future.

**Improvement compared to current main**
The improvement seems to be some 0.3 seconds (coming quite evenly from joblib, numpy, and connexion). 

Before (main):
```
$ time annif --version
1.0.0.dev0

real	0m0.610s
user	0m0.948s
sys	0m0.666s
```
```
$ time annif --help
Usage: annif [OPTIONS] COMMAND [ARGS]...
...
real	0m0.689s
user	0m1.079s
sys	0m0.626s
```
```
$ time annif list-vocabs 
Vocabulary ID       Languages                 Size  Loaded
----------------------------------------------------------
yso                 en,fi,se,sv              38179  True  
ykl                 -                            -  False 
thema-fi            -                            -  False 

real	0m1.214s
user	0m1.500s
sys	0m0.665s
```
After (PR):
```
$ time annif --version
1.0.0.dev0

real	0m0.293s
user	0m0.265s
sys	0m0.029s
```
```
$ time annif --help
Usage: annif [OPTIONS] COMMAND [ARGS]...
...
real	0m0.305s
user	0m0.266s
sys	0m0.038s
```
```
$ time annif list-vocabs 
Vocabulary ID       Languages                 Size  Loaded
----------------------------------------------------------
yso                 en,fi,se,sv              38179  True  
ykl                 -                            -  False 
thema-fi            -                            -  False 

real	0m0.945s
user	0m0.880s
sys	0m0.065s
```

The CLI command completion response times seemed to follow times of `annif --help`, so they are improved similarly.

Timing imports for completion responses does not show any easy targets for further optimization. The rdflib is used via multiple modules and in nearly all methods of `corpus.subject.SubjectFileSKOS`, and I don't know how rdflib import could be done only when needed in there in a nice way. 

Tuna output for the timing of `list-*` completion response time:

image

### DIFF
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -4,11 +4,6 @@ import logging
 import os
 import os.path
 
-import connexion
-from flask_cors import CORS
-
-from annif.openapi.validation import CustomRequestBodyValidator
-
 logging.basicConfig()
 logger = logging.getLogger("annif")
 logger.setLevel(level=logging.INFO)
@@ -16,20 +11,31 @@ logger.setLevel(level=logging.INFO)
 import annif.backend  # noqa
 
 
+def create_flask_app(config_name=None):
+    """Create a Flask app to be used by the CLI."""
+    from flask import Flask
+
+    app = Flask(__name__)
+    config_name = _get_config_name(config_name)
+    logger.debug(f"creating flask app with configuration {config_name}")
+    app.config.from_object(config_name)
+    app.config.from_envvar("ANNIF_SETTINGS", silent=True)
+    return app
+
+
 def create_app(config_name=None):
+    """ "Create a Connexion app to be used for the API."""
     # 'cxapp' here is the Connexion application that has a normal Flask app
     # as a property (cxapp.app)
+    import connexion
+    from flask_cors import CORS
+
+    from annif.openapi.validation import CustomRequestBodyValidator
 
     specdir = os.path.join(os.path.dirname(__file__), "openapi")
     cxapp = connexion.App(__name__, specification_dir=specdir)
-    if config_name is None:
-        config_name = os.environ.get("ANNIF_CONFIG")
-    if config_name is None:
-        if os.environ.get("FLASK_RUN_FROM_CLI") == "true":
-            config_name = "annif.default_config.Config"
-        else:
-            config_name = "annif.default_config.ProductionConfig"
-    logger.debug("creating app with configuration %s", config_name)
+    config_name = _get_config_name(config_name)
+    logger.debug(f"creating connexion app with configuration {config_name}")
     cxapp.app.config.from_object(config_name)
     cxapp.app.config.from_envvar("ANNIF_SETTINGS", silent=True)
 
@@ -52,3 +58,14 @@ def create_app(config_name=None):
 
     # return the Flask app
     return cxapp.app
+
+
+def _get_config_name(config_name):
+    if config_name is None:
+        config_name = os.environ.get("ANNIF_CONFIG")
+    if config_name is None:
+        if os.environ.get("FLASK_RUN_FROM_CLI") == "true":
+            config_name = "annif.default_config.Config"
+        else:
+            config_name = "annif.default_config.ProductionConfig"
+    return config_name
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -25,7 +25,14 @@ from annif.util import metric_code
 logger = annif.logger
 click_log.basic_config(logger)
 
-cli = FlaskGroup(create_app=annif.create_app, add_version_option=False)
+
+if len(sys.argv) > 1 and sys.argv[1] == "run":
+    create_app = annif.create_app  # Use Flask with Connexion
+else:
+    # Connexion is not needed for most CLI commands, use plain Flask
+    create_app = annif.create_flask_app
+
+cli = FlaskGroup(create_app=create_app, add_version_option=False)
 cli = click.version_option(message="%(version)s")(cli)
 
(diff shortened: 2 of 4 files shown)


juhoinkinen:
The rdflib import could be [turned to lazy](https://github.com/NatLibFi/Annif/commit/9983b4bb16cbef8acf9ca60ad3de84582c0a58dc), but the code is not super clean and the time saving is only ~0.05 s. 

However I would still add it. For the command completions even such a tiny improvement in the response time is noticeable I think.

osma (changes requested):
The code looks OK, although making more and more local imports makes the code messier IMHO. But it seems worth it in this case, as the startup speed improves noticeably (I also tested this on my laptop and the speedup was around 0.2 seconds). Maybe someday we could switch to real [PEP 690 lazy imports](https://peps.python.org/pep-0690/), but those are scheduled for Python 3.12, which should be released in a few days, but it will take years until we can drop support for 3.11.

I see that Codecov complains about cli.py line 31 not being covered by tests, which I found puzzling, since you added a test that runs `annif run --help` which should execute this line. But then I added some debugging code around that line and found out that the test doesn't actually exercise that line at all. I think that's because when using CliRunner (e.g. `runner.invoke(annif.cli.cli, ["run", "--help"])`) that doesn't change `sys.argv` at all, so it also won't execute line 31 of cli.py. Is there anything we could do here to make sure that the test also exercises this path?

juhoinkinen:
> I see that Codecov complains about cli.py line 31 not being covered by tests, which I found puzzling, since you added a test that runs `annif run --help` which should execute this line. But then I added some debugging code around that line and found out that the test doesn't actually exercise that line at all. I think that's because when using CliRunner (e.g. `runner.invoke(annif.cli.cli, ["run", "--help"])`) that doesn't change `sys.argv` at all, so it also won't execute line 31 of cli.py. Is there anything we could do here to make sure that the test also exercises this path?

The line could definitely be run in a test such as
```python
def test_run_separate_process():
    result = os.popen("python annif/cli.py run --help").read()
    assert "Run a local development server." in result
```
But I don't know how to ensure this this really starts Flask with Connexion, and not just Flask... And codecov does not take this kind of test into account in coverage counting (at least locally run).

osma:
The `os.popen` solution doesn't seem very useful, for the reasons you mention.

How about using some other information than just `sys.argv[1]` in `cli.py` when deciding whether to choose Connexion or plain Flask?

For testing `parallel.py` using different spawn methods, I added a constant definition like this: https://github.com/NatLibFi/Annif/blob/3451bd351273e3425a2bea683f8ac2bf3618278b/annif/parallel.py#L7-L10

The unit test that exercises this code overrides the value using monkeypatching: https://github.com/NatLibFi/Annif/blob/3451bd351273e3425a2bea683f8ac2bf3618278b/tests/test_cli.py#L887-L900

Something similar could perhaps be done here - define a constant in cli.py which uses sys.argv and then use it in the `if` clause:

```python
USE_CONNEXION = len(sys.argv) > 1 and sys.argv[1] == "run"

if USE_CONNEXION:
    create_app = annif.create_app  # Use Flask with Connexion
else:
    # Connexion is not needed for most CLI commands, use plain Flask
    create_app = annif.create_flask_app
```

Then you can monkeypatch USE_CONNEXION in the unit test to force it to be `True` in any case.

osma:
...except now that I think of it, I'm not sure if that will work, because monkeypatching is probably applied too late for this to have an effect. Anyway, something similar could work here.

juhoinkinen:
> > I see that Codecov complains about cli.py line 31 not being covered by tests, which I found puzzling, since you added a test that runs `annif run --help` which should execute this line. But then I added some debugging code around that line and found out that the test doesn't actually exercise that line at all. I think that's because when using CliRunner (e.g. `runner.invoke(annif.cli.cli, ["run", "--help"])`) that doesn't change `sys.argv` at all, so it also won't execute line 31 of cli.py. Is there anything we could do here to make sure that the test also exercises this path?
> 
> The line could definitely be run in a test such as
> 
> ```python
> def test_run_separate_process():
>     result = os.popen("python annif/cli.py run --help").read()
>     assert "Run a local development server." in result
> ```
> 
> But I don't know how to ensure this this really starts Flask with Connexion, and not just Flask... And codecov does not take this kind of test into account in coverage counting (at least locally run).

Actually using the `os.popen` solution does get the line 31 covered in codecov (when run in CI). When I locally ran `coverage run -m pytest tests/test_cli.py; coverage report -m | grep annif/cli.py` that line was reported as missed. 

And I made also `annif routes` command to start the Connexion app, which registers all REST API endpoints. The list of the registered endpoints from the output of `routes` command is now tested to confirm that Connexion is started.

I think this can be merged. When upgrading to Connexion 3 (#702) this functionality can or may need to be changed again.

PR #700, opened 2023-05-08
Make stwfsapy an optional dependency

### LINKED ISSUE #699: Make stwfsapy an optional dependency
We would like to release the STWFSA backend into the community. To ensure that Annif is not fundamentally dependent on the compatibility of stwfsapy with Annif, we suggest making stwfsapy an optional dependency in Annif. This way, stwfsapy can still be used within Annif but cannot break the entire system in case of temporary incompatibilities, because it can be removed more easily than a core dependency.

 This can be achieved by declaring stwfsapy as an optional dependency in ``pyproject.toml``. The import statement inside ``annif/backend/init.py`` should be wrapped by a ``try...except`` block to handle any import errors.

cbartz:
Closes #699

### DIFF
--- a/annif/backend/__init__.py
+++ b/annif/backend/__init__.py
@@ -62,9 +62,12 @@ def _pav():
 
 
 def _stwfsa():
-    from . import stwfsa
+    try:
+        from . import stwfsa
 
-    return stwfsa.StwfsaBackend
+        return stwfsa.StwfsaBackend
+    except ImportError:
+        raise ValueError("STWFSA not available, cannot use stwfsa backend")
 
 
 def _svc():
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -61,7 +61,7 @@ jobs:
         # Selectively install the optional dependencies for some Python versions
         # For Python 3.8:
         if [[ ${{ matrix.python-version }} == '3.8' ]]; then
-          poetry install -E "nn omikuji yake voikko";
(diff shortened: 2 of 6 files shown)


annif/backend/__init__.py
```diff
@@ -62,9 +62,12 @@ def _pav():
...
 
 def _stwfsa():
-    from . import stwfsa
+    try:
+        from . import stwfsa
 
-    return stwfsa.StwfsaBackend
+        return stwfsa.StwfsaBackend
+    except ImportError:
+        raise ValueError("STWFSA not available, cannot use stwfsa backend")
```
juhoinkinen, line 70: Test coverage could be extended to these lines by adding a test like https://github.com/NatLibFi/Annif/blob/3451bd351273e3425a2bea683f8ac2bf3618278b/tests/test_backend.py#L86-L93

juhoinkinen (changes requested):
STWFSA should be added to be installed in the Docker image by adding it to the `optional_dependencies` variable in https://github.com/NatLibFi/Annif/blob/3451bd351273e3425a2bea683f8ac2bf3618278b/Dockerfile#L5

Otherwise looks good and ready to be merged. The missing test coverage could also be addressed (see the inline comment), but that's a minor thing.

(Annif Wiki needs to be modified to account for the changes, but that should be done when/after next Annif release is made, we should remember that.)

juhoinkinen:
Thanks @cbartz! Merging this now.

Issue #698, opened 2023-04-27
Connexion error on posting wrong content type to suggest method


juhoinkinen:
A connexion error occurs if application/json content-type  is used in requests to `/v1/projects/<project>/suggest` method.

For example this curl request produces a 500 response:

    curl -X POST -H "Content-Type: application/json" -d '{"documents": [{"text": "content"}]}' http://localhost:5000/v1/projects/tfidf-fi/suggest

And the server logs show:
```
Exception on /v1/projects/tfidf-fi/suggest [POST]
Traceback (most recent call last):
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/flask/app.py", line 2529, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/connexion/decorators/decorator.py", line 68, in wrapper
    response = function(request)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/connexion/decorators/uri_parsing.py", line 149, in wrapper
    response = function(request)
[...]

PR #702, opened 2023-05-10
Upgrade to Connexion3

### LINKED ISSUE #689: Lots of pytest warnings / upgrade to Connexion 3
Running pytest nowadays generates lots of warnings:

```
==== 384 passed, 4 skipped, 4 deselected, 807 warnings in 71.66s (0:01:11) ====
```

Many of these are Flask deprecation warnings triggered by connexion, e.g.:

```
 .../lib/python3.10/site-packages/sklearn/feature_extraction/text.py:528: UserWarning: The parameter 'token_pattern' will not be used since 'tokenizer' is not None'
    warnings.warn(

tests/test_openapi.py: 57 warnings
  .../lib/python3.10/site-packages/connexion/apis/flask_api.py:236: DeprecationWarning: '_request_ctx_stack' is deprecated and will be removed in Flask 2.3.
    setattr(flask._request_ctx_stack.top, 'connexion_context', context_dict)

tests/test_openapi.py: 57 warnings
  .../lib/python3.10/site-packages/connexion/apis/flask_api.py:236: DeprecationWarning: '_request_ctx_stack' is deprecated and will be removed in Flask 2.3. Use 'g' to store data, or 'request_ctx' to access the current context.
    setattr(flask._request_ctx_stack.top, 'connexion_context', context_dict)

tests/test_openapi.py: 60 warnings
  .../lib/python3.10/site-packages/flask/json/provider.py:188: DeprecationWarning: Setting 'json_encoder' on the app or a blueprint is deprecated and will be removed in Flask 2.3. Customize 'app.json' instead.
    warnings.warn(

tests/test_openapi.py: 60 warnings
  .../lib/python3.10/site-packages/flask/json/provider.py:230: DeprecationWarning: 'JSONEncoder' is deprecated and will be removed in Flask 2.3. Use 'Flask.json' to provide an alternate JSON implementation instead.
    return json.dumps(obj, **kwargs)
```

It seems to me that these have been [fixed](https://github.com/spec-first/connexion/pull/1582) (and [more](https://github.com/spec-first/connexion/pull/1583)) in the Connexion `main` branch, but the fixes are not in version 2.14.2 that we are currently using. Instead, Connexion 3.0 is coming up soon, with the most recent [release](https://github.com/spec-first/connexion/releases) being 3.0.0a4 currently.

I think we should try to upgrade to Connexion 3 because I don't think version 2 will have a long future. That should take care of most of the pytest warnings.

juhoinkinen:
Upgrades to Connexion 3, closes #689 and #698.

Connexion 3 introduces many changes compared to Connexion 2, see the documentation [here](https://connexion.readthedocs.io/en/stable/v3.html).

Most importantly, after the upgrade to Connexion 3, running Annif with Gunicorn requires the use Uvicorn workers; the workers can be set using the `--worker-class/-k` option of Gunicorn:

    gunicorn --worker-class uvicorn.workers.UvicornWorker "annif:create_app()"

In Dockerfile there is the [enviroment variable](https://docs.gunicorn.org/en/stable/settings.html) `GUNICORN_CMD_ARGS="--worker-class uvicorn.workers.UvicornWorker"`, which gives Gunicorn the option by default, so Docker image users do not have to worry about this change.

---
The original PR discussion below beginning from first draft.

This seems to be surprisingly difficult. Connexion 3 documentation states that there are two modes of running/usage of Connexion: 
  1.  [Using stand-alone Connexion](https://connexion.readthedocs.io/en/latest/v3.html#using-stand-alone-connexion) - use the provided Flask app via `connexion.FlaskApp`
  2. [Using Connexion with existing ASGI or WSGI frameworks](https://connexion.readthedocs.io/en/latest/v3.html#using-connexion-with-asgi-or-wsgi-frameworks) - use the "middleware" `connexion.ConnexionMiddleware` to wrap an existing Flask app 

Previously Annif has been using the mode 1, and tried to continue with that. But now
> [Connexion 3.0 needs to be run using an ASGI server instead of a WSGI server. While any ASGI server should work, connexion comes with uvicorn as an extra](https://connexion.readthedocs.io/en/latest/v3.html#asgi-server)

And the app needs to be started with

     uvicorn --factory annif:create_app

The command

     annif run

does not know about uvicorn but apparently tries to use plain Flask, and the app won't start.

There are also [smaller breaking changes](https://connexion.readthedocs.io/en/latest/v3.html#smaller-breaking-changes), that needed to be addressed:
 - [Set content-types in response headers in rest methods](https://github.com/NatLibFi/Annif/commit/6d0a68fc4582d14393fa7abb97361befe5e8f53c), because

   - "We no longer guess a content type for response serialization if multiple are defined in the spec. We do take into account returned headers." (The Annif OpenAPI spec included both `application/json` and `application/problem+json`, which seemed to be the cause.)
    - "Content type is now validated for requests and responses if defined in the spec"

-  Changing `req.get_json()` -> `req.json()` in tests, apparently because of "connexion.request is now a Starlette Request instead of a Flask Request".

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,4 +54,6 @@ RUN groupadd -g 998 annif_user && \
     chown -R annif_user:annif_user /annif-projects /Annif/tests/data
 USER annif_user
 
+ENV GUNICORN_CMD_ARGS="--worker-class uvicorn.workers.UvicornWorker"
+
 CMD annif
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -7,6 +7,8 @@ import os
 import os.path
 from typing import TYPE_CHECKING
 
+from flask import Flask
+
 logging.basicConfig()
 logger = logging.getLogger("annif")
 logger.setLevel(level=logging.INFO)
@@ -14,12 +16,11 @@ logger.setLevel(level=logging.INFO)
 import annif.backend  # noqa
 
 if TYPE_CHECKING:
-    from flask.app import Flask
+    from connexion.apps.flask import FlaskApp
 
 
 def create_flask_app(config_name: str | None = None) -> Flask:
     """Create a Flask app to be used by the CLI."""
-    from flask import Flask
 
     _set_tensorflow_loglevel()
 
@@ -31,7 +32,7 @@ def create_flask_app(config_name: str | None = None) -> Flask:
     return app
 
 
-def create_app(config_name: str | None = None) -> Flask:
+def create_cx_app(config_name: str | None = None) -> FlaskApp:
     """Create a Connexion app to be used for the API."""
     import connexion
 
@@ -66,6 +67,9 @@ def create_app(config_name: str | None = None) -> Flask:
 
     # return the Connexion app
     return cxapp
+
+
+create_app = create_cx_app  # Alias to allow starting directly with uvicorn run
 
 
 def _get_config_name(config_name: str | None) -> str:
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -30,19 +30,21 @@ def create_app(config_name=None):
     import connexion
     from flask_cors import CORS
 
-    from annif.openapi.validation import CustomRequestBodyValidator
+    import annif.registry
+
+    # from annif.openapi.validation import CustomRequestBodyValidator  # TODO Re-enable
 
     specdir = os.path.join(os.path.dirname(__file__), "openapi")
-    cxapp = connexion.App(__name__, specification_dir=specdir)
+    cxapp = connexion.FlaskApp(__name__, specification_dir=specdir)
     config_name = _get_config_name(config_name)
     logger.debug(f"creating connexion app with configuration {config_name}")
     cxapp.app.config.from_object(config_name)
     cxapp.app.config.from_envvar("ANNIF_SETTINGS", silent=True)
 
-    validator_map = {
-        "body": CustomRequestBodyValidator,
-    }
-    cxapp.add_api("annif.yaml", validator_map=validator_map)
+    # validator_map = {
+    #     "body": CustomRequestBodyValidator,
+    # }
+    cxapp.add_api("annif.yaml")  # validator_map=validator_map)
 
     # add CORS support
     CORS(cxapp.app)
@@ -56,8 +58,8 @@ def create_app(config_name=None):
 
     cxapp.app.register_blueprint(bp)
 
-    # return the Flask app
-    return cxapp.app
+    # return the Connexion app
+    return cxapp
 
 
 def _get_config_name(config_name):
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -33,11 +33,9 @@ def create_flask_app(config_name: str | None = None) -> Flask:
 
 def create_app(config_name: str | None = None) -> Flask:
     """Create a Connexion app to be used for the API."""
-    # 'cxapp' here is the Connexion application that has a normal Flask app
-    # as a property (cxapp.app)
     import connexion
-    from flask_cors import CORS
 
+    # from flask_cors import CORS  # TODO Use CORSMiddleware
     import annif.registry
 
     # from annif.openapi.validation import CustomRequestBodyValidator  # TODO Re-enable
@@ -55,7 +53,7 @@ def create_app(config_name: str | None = None) -> Flask:
     cxapp.add_api("annif.yaml")  # validator_map=validator_map)
 
     # add CORS support
-    CORS(cxapp.app)
+    # CORS(cxapp.app)
 
     if cxapp.app.config["INITIALIZE_PROJECTS"]:
         annif.registry.initialize_projects(cxapp.app)
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -35,13 +35,13 @@ def create_flask_app(config_name: str | None = None) -> Flask:
 def create_cx_app(config_name: str | None = None) -> FlaskApp:
     """Create a Connexion app to be used for the API."""
     import connexion
+    from connexion.datastructures import MediaTypeDict
     from connexion.middleware import MiddlewarePosition
+    from connexion.validators import FormDataValidator, MultiPartFormDataValidator
     from starlette.middleware.cors import CORSMiddleware
 
-    # from flask_cors import CORS  # TODO Use CORSMiddleware
     import annif.registry
-
-    # from annif.openapi.validation import CustomRequestBodyValidator  # TODO Re-enable
+    from annif.openapi.validation import CustomRequestBodyValidator
 
     specdir = os.path.join(os.path.dirname(__file__), "openapi")
     cxapp = connexion.FlaskApp(__name__, specification_dir=specdir)
@@ -50,10 +50,16 @@ def create_cx_app(config_name: str | None = None) -> FlaskApp:
     cxapp.app.config.from_object(config_name)
     cxapp.app.config.from_envvar("ANNIF_SETTINGS", silent=True)
 
-    # validator_map = {
-    #     "body": CustomRequestBodyValidator,
-    # }
-    cxapp.add_api("annif.yaml")  # validator_map=validator_map)
+    validator_map = {
+        "body": MediaTypeDict(
+            {
+                "*/*json": CustomRequestBodyValidator,
+                "application/x-www-form-urlencoded": FormDataValidator,
+                "multipart/form-data": MultiPartFormDataValidator,
+            }
+        ),
+    }
+    cxapp.add_api("annif.yaml", validator_map=validator_map)
 
     # add CORS support
     cxapp.add_middleware(
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -35,6 +35,8 @@ def create_flask_app(config_name: str | None = None) -> Flask:
 def create_cx_app(config_name: str | None = None) -> FlaskApp:
     """Create a Connexion app to be used for the API."""
     import connexion
+    from connexion.middleware import MiddlewarePosition
+    from starlette.middleware.cors import CORSMiddleware
 
     # from flask_cors import CORS  # TODO Use CORSMiddleware
     import annif.registry
@@ -54,7 +56,11 @@ def create_cx_app(config_name: str | None = None) -> FlaskApp:
     cxapp.add_api("annif.yaml")  # validator_map=validator_map)
 
     # add CORS support
-    # CORS(cxapp.app)
+    cxapp.add_middleware(
+        CORSMiddleware,
+        position=MiddlewarePosition.BEFORE_EXCEPTION,
+        allow_origins=["*"],
+    )
 
     if cxapp.app.config["INITIALIZE_PROJECTS"]:
         annif.registry.initialize_projects(cxapp.app)
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -27,7 +27,7 @@ logger = annif.logger
 click_log.basic_config(logger)
 
 
-if len(sys.argv) > 1 and sys.argv[1] == "run":
+if len(sys.argv) > 1 and sys.argv[1] in ("run", "routes"):
     create_app = annif.create_app  # Use Flask with Connexion
 else:
     # Connexion is not needed for most CLI commands, use plain Flask
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -439,7 +439,7 @@ def run_eval(
 
 
 @cli.command("run")
-@click.option("--port", type=int)
+@click.option("--port", type=int, default=5000)
 @click.option("--log-level")
 @click_log.simple_verbosity_option(logger, default="ERROR")
 def run_app(**kwargs):
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -30,6 +30,7 @@ cli = FlaskGroup(
     create_app=create_app, add_default_commands=False, add_version_option=False
 )
 cli = click.version_option(message="%(version)s")(cli)
+cli.params = [opt for opt in cli.params if opt.name not in ("env_file", "app")]
 
 
 @cli.command("list-projects")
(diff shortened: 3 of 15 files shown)


juhoinkinen:
I added [added `annif run` CLI command for starting uvicorn dev server](https://github.com/NatLibFi/Annif/pull/702/commits/39027488af3c4b285746249e3e9e2ff2f87a9b5c) as a custom Click command and removed the default `run` ,`shell` and `routes` Flask commands. The `cli` object used by `annif` command is still an instance of `FlaskGroup`, but maybe that could be turned into a custom [Click command group](https://click.palletsprojects.com/en/8.1.x/commands/). I don't know if this would then uncouple Flask fully from Annif.

Note that the default port for uvicorn is 8000, whereas for Flask it is 5000.

Also the `tests/test_openapi.py` do not pass due to an error:

    AttributeError: module 'anyio' has no attribute 'start_blocking_portal'

but this seems to be some version mismatch of starlette.

And some `tests/test_cli.py` need to be adjusted.

All in all, maybe there could be an alternative to Connexion, which could allow still using Flask.

juhoinkinen:
> Also the `tests/test_openapi.py` do not pass due to an error:
> 
> ```
> AttributeError: module 'anyio' has no attribute 'start_blocking_portal'
> ```
> 
> but this seems to be some version mismatch of starlette.

Starlette-testclient has not pinned (or declared at all?) its anyio dependency, and anyio 4.* do not work with it: https://github.com/Kludex/starlette-testclient/issues/5

[Schemathesis v3.2.3](https://github.com/schemathesis/schemathesis/releases/tag/v3.23.0) takes this into account and requires anyio <4, but for some reason anyio v4.2.0 gets installed.

A workaround is to pin `anyio<4.*` as a direct dependency.

juhoinkinen:
> > Also the `tests/test_openapi.py` do not pass due to an error:
> > ```
> > AttributeError: module 'anyio' has no attribute 'start_blocking_portal'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but this seems to be some version mismatch of starlette.
> 
> Starlette-testclient has not pinned (or declared at all?) its anyio dependency, and anyio 4.* do not work with it: [Kludex/starlette-testclient#5](https://github.com/Kludex/starlette-testclient/issues/5)
> 
> [Schemathesis v3.2.3](https://github.com/schemathesis/schemathesis/releases/tag/v3.23.0) takes this into account and requires anyio <4, but for some reason anyio v4.2.0 gets installed.
> 
> A workaround is to pin `anyio<4.*` as a direct dependency.

Anyio pinning is not needed anymore, when doing these updates
- jsonschema 4.17.* -> 4.21.*
- pytest 7.* -> 8.*

juhoinkinen:
> When setting CORS in `annif/__init__.py` with
> 
> ```python
> cxapp.add_middleware(
>      CORSMiddleware,
>     position=MiddlewarePosition.BEFORE_EXCEPTION,
>     allow_origins=["*"],
>     allow_methods=["*"],
> )
> ```
> 
> CORS is not working for unit tests:
> 
> ```
> tests/test_openapi.py:43: AssertionError
> ...
> FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/projects] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/projects/{project_id}] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/suggest] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/suggest-batch] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/learn] - KeyError: 'access-control-allow-origin'
> FAILED tests/test_openapi.py::test_openapi_list_projects - assert 0
> ```
> 
> However, when looking at the response headers in the Firefox network tab, the `'access-control-allow-origin'` header is there (at least for /suggest request), so there might be an issue due to starlette-testclient.

Also curl output shows the CORS header to be sent (when using `H "Origin: http://example.com"`):

    curl -H "Origin: http://example.com" -X 'POST'   'http://127.0.0.1:8000/v1/projects/yake-fi/suggest'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'text=kissa&limit=10&threshold=0&language=' -v

juhoinkinen:
TODO: Having jsonschema as a [direct dependency](https://github.com/NatLibFi/Annif/blob/22d9599bef86e97e544d886f1cdc9c4025218118/pyproject.toml#L50) may not be required after this PR, if the `annif/openapi/validation.py` will not [be using it anymore](https://github.com/NatLibFi/Annif/blob/22d9599bef86e97e544d886f1cdc9c4025218118/annif/openapi/validation.py#L6).

juhoinkinen:
> > When setting CORS in `annif/__init__.py` with
> > ```python
> > cxapp.add_middleware(
> >      CORSMiddleware,
> >     position=MiddlewarePosition.BEFORE_EXCEPTION,
> >     allow_origins=["*"],
> >     allow_methods=["*"],
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > CORS is not working for unit tests:
> > ```
> > tests/test_openapi.py:43: AssertionError
> > ...
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/projects] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[GET /v1/projects/{project_id}] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/suggest] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/suggest-batch] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_fuzzy[POST /v1/projects/{project_id}/learn] - KeyError: 'access-control-allow-origin'
> > FAILED tests/test_openapi.py::test_openapi_list_projects - assert 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, when looking at the response headers in the Firefox network tab, the `'access-control-allow-origin'` header is there (at least for /suggest request), so there might be an issue due to starlette-testclient.
> 
> Also curl output shows the CORS header to be sent (when using `H "Origin: http://example.com"`):
> 
> ```
> curl -H "Origin: http://example.com" -X 'POST'   'http://127.0.0.1:8000/v1/projects/yake-fi/suggest'   -H 'accept: application/json'   -H 'Content-Type: application/x-www-form-urlencoded'   -d 'text=kissa&limit=10&threshold=0&language=' -v
> ```

The CORS testing problem was resolved by removing CORS check from the fuzzying test and (re)creating a dedicated test for it, which adds Origin header to the request:
```python
app_client.headers = {"Origin": "http://somedomain.com"}
```

--- 


But now there is a (new?) problem: the test_openapi_fuzzy discovered that `%0A`, i.e. encoded newline, in the URL parameters crashes the application. This does not occur with Connexion 2.

osma (changes requested):
I tested this PR and here are some findings:

# CLI operation

* The Flask-related commands `routes` and `shell` have been dropped. This is good because they're irrelevant for Annif anyway. But the `-e, --env-file` and `-A, --app` options that seem to be Flask-specific are still there - could we easily get rid of those too?
* I timed a few commands (`help`, `list-projects`, `suggest`) and they don't seem to take longer to run than they used to - some are maybe even a little faster than before. Excellent!

# `annif run` command

* The console output now looks a bit different than it used to, due to the switch to uvicorn. But I think the new output looks good.
* Annif now listens on port 8000, when it used to listen on port 5000. I think it would be better to keep the old port, because it's documented for example in the Annif tutorial [Web UI](https://github.com/NatLibFi/Annif-tutorial/blob/master/exercises/03_web_ui.md) and [REST API](https://github.com/NatLibFi/Annif-tutorial/blob/master/exercises/OPT_rest_api.md) exercises. Simply defining a default value of 5000 for `--port` should work, right?

# Swagger-UI

* The category view in `/v1/ui` now has a different order - "Learning from feedback" comes last. It's better this way!
* Method parameters are also in a different order - not worse than before, maybe even a bit better. The layout for the "Send empty value" checkbox seems a bit broken, though I'm unsure if we can do anything about it. Here is an example that shows old (main) on the left, new (PR branch) on the right:

image

# REST API

* CORS now works differently than before. I'm not sure if this is bad, just something to note (maybe put in the release notes?).
  * Before: If there's and `Origin` header in request, send `Access-Control-Allow-Origin: <Origin header value>`, otherwise send `Access-Control-Allow-Origin: *`
  * After: If there's and `Origin` header in request, send `Access-Control-Allow-Origin: *`, otherwise send no such header.
* The URL `/projects/` used to give a 404 response, now it redirects to the correct URL `/projects`. Again, not necessarily a problem, just different, and perhaps worth putting into the release notes.
* Successful JSON responses are pretty-printed, but error responses in Problem JSON are not. Previously both kinds of JSON responses were pretty-printed for easier readability. I think readable JSON would be good especially for the error messages, since they are likely to be seen by humans at some point.

# Other

* SonarCloud has a few [complaints](https://sonarcloud.io/project/issues?id=NatLibFi_Annif&pullRequest=702&resolved=false&inNewCodePeriod=true) - are these real problems?
* CodeClimate also [complains](https://codeclimate.com/github/NatLibFi/Annif/pull/702) of complexity, but it seems like the affected code was already complex, so not really a regression
* Codecov [complains](https://github.com/NatLibFi/Annif/pull/702/files#annotation_20180234651) about L25 in validation.py not being covered by tests, would it be easy to test this? Or add a notation that excludes it from test coverage?
* See comment about Connexion version specifier

juhoinkinen:
> I tested this PR and here are some findings:
> # CLI operation
> 
>   * The Flask-related commands `routes` and `shell` have been dropped. This is good because they're irrelevant for Annif anyway. But the `-e, --env-file` and `-A, --app` options that seem to be Flask-specific are still there - could we easily get rid of those too?

Yes, done.
 
>   * I timed a few commands (`help`, `list-projects`, `suggest`) and they don't seem to take longer to run than they used to - some are maybe even a little faster than before. Excellent!
> 
> 
> # `annif run` command
> 
>   * The console output now looks a bit different than it used to, due to the switch to uvicorn. But I think the new output looks good.
> 
>   * Annif now listens on port 8000, when it used to listen on port 5000. I think it would be better to keep the old port, because it's documented for example in the Annif tutorial [Web UI](https://github.com/NatLibFi/Annif-tutorial/blob/master/exercises/03_web_ui.md) and [REST API](https://github.com/NatLibFi/Annif-tutorial/blob/master/exercises/OPT_rest_api.md) exercises. Simply defining a default value of 5000 for `--port` should work, right?

Yes, done.

> # Swagger-UI
> 
>   * The category view in `/v1/ui` now has a different order - "Learning from feedback" comes last. It's better this way!
> 
>   * Method parameters are also in a different order - not worse than before, maybe even a bit better. The layout for the "Send empty value" checkbox seems a bit broken, though I'm unsure if we can do anything about it. Here is an example that shows old (main) on the left, new (PR branch) on the right:
> 
> 
> image
> # REST API
> 
>  * CORS now works differently than before. I'm not sure if this is bad, just something to note (maybe put in the release notes?). 
>    * Before: If there's an `Origin` header in the request, send `Access-Control-Allow-Origin: <Origin header value>`, otherwise send `Access-Control-Allow-Origin: *`
>    * After: If there's an `Origin` header in the request, send `Access-Control-Allow-Origin: *`, otherwise send no such header.
>  * The URL `/projects/` used to give a 404 response, now it redirects to the correct URL `/projects`. Again, not necessarily a problem, just different, and perhaps worth putting into the release notes.

TODO in Relase notes.
 
>  * Successful JSON responses are pretty-printed, but error responses in Problem JSON are not. Previously both kinds of JSON responses were pretty-printed for easier readability. I think readable JSON would be good especially for the error messages, since they are likely to be seen by humans at some point.
 
Yes, but seems that this cannot be configured, see an old issue: https://github.com/spec-first/connexion/pull/313

> # Other
> 
>  * SonarCloud has a few [complaints](https://sonarcloud.io/project/issues?id=NatLibFi_Annif&pullRequest=702&resolved=false&inNewCodePeriod=true) - are these real problems?

I would say no, I silenced them.

>  * CodeClimate also [complains](https://codeclimate.com/github/NatLibFi/Annif/pull/702) of complexity, but it seems like the affected code was already complex, so not really a regression

Yes, and the code in question is to override a [function of Connexion](https://github.com/spec-first/connexion/blob/e6ceb1fc0a13fbe504163592c15e8cac4b52fc6f/connexion/validators/json.py#L64-L75), so it is not meaningful to simplify it. I silenced the complain as "wontfix".
 
>  * Codecov [complains](https://github.com/NatLibFi/Annif/pull/702/files#annotation_20180234651) about L25 in validation.py not being covered by tests, would it be easy to test this? Or add a notation that excludes it from test coverage?

I tried to add test for this line, but it is not detected by CodeCov. I added `# pragma: no cover` annotation for the line to exclude it.

juhoinkinen:
From now on when running Annif with Gunicorn one must set the workers to uvicorn with the `-k` option, that is to use command

    gunicorn -k uvicorn.workers.UvicornWorker "annif:create_app()"

This needs to be clearly communicated and updated to [docker-compose.yml](https://github.com/NatLibFi/Annif/blob/main/docker-compose.yml).

PR #705, opened 2023-05-15
Fix crashing `index` command when targeted directory contains subject files

juhoinkinen:
The `annif index` command crashes when the targeted directory contains subject files (files with either `*.tsv` or `*.key` extensions):
```
annif index dummy-fi tests/corpora/archaeology/fulltext/
...
  File "/home/local/jmminkin/git/Annif/annif/corpus/subject.py", line 248, in from_string
    subject_ids.add(subject_index.by_uri(uri))
AttributeError: 'NoneType' object has no attribute 'by_uri'
```

The bug is due to [this change](https://github.com/NatLibFi/Annif/pull/663/commits/47423e51819045c432761d77b77a681ee5a76abb#diff-30e62ce44e760ec9d177f1be5e0515810b9813c6ff99cc2ed9f4cc5fd48d7fb8R446) done in PR #663.

I made the change because it seemed unnecessary and confusing to pass the subject index and language to `DocumentDirectory` when it is used only for obtaining documents to give suggestions to. 

Reverting the change, i.e. passing subject index and language to `DocumentDirectory` is the simplest way to fix the bug, and on second thought maybe it also is better conceptually: corpora should have subject index and language defined, even if they are not used.

Also [renames vars to better correspond their origin/usage: keyfile(name) -> subjfile(name)](https://github.com/NatLibFi/Annif/commit/b56b31e0c49e5b9270f5067af135a44536a2aa5e).

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -318,9 +318,7 @@ def run_index(
         raise click.BadParameter(f'language "{lang}" not supported by vocabulary')
     backend_params = cli_util.parse_backend_params(backend_param, project)
 
-    documents = annif.corpus.DocumentDirectory(
-        directory, project.subjects, lang, require_subjects=False
-    )
+    documents = annif.corpus.DocumentDirectory(directory, require_subjects=False)
     results = project.suggest_corpus(documents, backend_params).filter(limit, threshold)
 
     for (docfilename, _), suggestions in zip(documents, results):
--- a/annif/corpus/document.py
+++ b/annif/corpus/document.py
@@ -17,7 +17,7 @@ logger = annif.logger
 class DocumentDirectory(DocumentCorpus):
     """A directory of files as a full text document corpus"""
 
-    def __init__(self, path, subject_index, language, require_subjects=False):
+    def __init__(self, path, subject_index=None, language=None, require_subjects=False):
         self.path = path
         self.subject_index = subject_index
         self.language = language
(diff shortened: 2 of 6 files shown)


osma (approved):
This fixes the problem for sure and the variable renamings are sensible.

However, I think another approach would also be possible. DocumentDirectory already takes a `require_subjects` flag, and when set to False, the subjects are not required, but they are still parsed from tsv/key files and returned if such files exist. This is wasted effort in the case of the `annif index` operation, because the subjects aren't going to be used anyway. So another solution would be to modify DocumentDirectory so it doesn't even check for tsv/key files if `require_subjects==False`. Then it would also be unnecessary to pass the subject index and language to the DocumentDirectory. If the default value of these parameters were set to None, it could be constructed like this: `DocumentDirectory(directory, require_subjects=False)`. I think that would be even more elegant than the solution in this PR. However, I'm happy to accept either solution.

osma (approved):
LGTM

san-uh:
Dear Annif colleagues, thank you for already working on this problem. We in the DNB have recently started using Annif 0.61.0 and have now noticed that we cannot currently use an "index" command with this version, so we have to switch back to 0.60.0.  Just one question: Is a bugfix release 0.61.1 planned or will the solution come with 0.62.0? Thanks and greetings, Sandro

juhoinkinen:
Hi @san-uh,

This bug realized only when the targeted directory contains files with either *.tsv or *.key extension (as far as I'm aware). Does `index` command work for you if you remove these files from the targeted directory (if there are them)?

san-uh:
Hello @juhoinkinen, 
thank you very much for the explanation. I deleted the *.tsv files that were previously in the directory and the index command now works. So with this workaround annif 0.61.0 including `index` is usable. Thanks for the feedback! Sandro

Issue #690, opened 2023-04-14
Use Python type hints


osma:
Since Python 3.5 it has been possible to add type hints for variables and function/method signatures ([PEP 484](https://peps.python.org/pep-0484/)).

We should do this within the Annif codebase as well. [Here](https://towardsdatascience.com/12-beginner-concepts-about-type-hints-to-improve-your-python-code-90f1ba0ac49) is a good basic tutorial about the available types and how to use them. Quoting from the post:

> Type hints bring an additional layer of abstraction on top of your code: they help to document it, clarify the assumptions about inputs/outputs and prevent insidious and silent mistakes when static code analysis (👋 mypy) is performed on top.

Eventually we should also enforce the type hints with [mypy](https://mypy-lang.org/), but let's leave that out of scope for this issue so that it can be closed once we have the hints in place at least for a reasonable proportion of the codebase.

juhoinkinen:
[A few tools have been mentioned](https://stackoverflow.com/a/50416746) that can help by (semi)automatically generating the type annotations:

> Sure, there are some projects out there that can help out.
> 
>>   Instagram developed the MonkeyType tool to pick up runtime type information. These are output as stubs files, but you can use the tools included to apply those to your source files as inline annotations, or use the retype project to do the same.
>>   Dropbox has a similar tool called pyannotate; this tool can generate either Python 2 type hint comments or annotations.
> 
> Neither project will produce perfect type hints; always take their output as a starting point, not the authoritative final annotations.

juhoinkinen:
[As instructed](https://github.com/mariusvniekerk/pytest-monkeytype#readme), I run `py.test --monkeytype-output=./monkeytype.sqlite3` to trace and record the variable types to/from functions based on pytest run. This  data covers also the unit tests. 

Then I run `monkeytype apply <module> --pep_563` to all modules (not unit tests) to insert the type annotations to function signatures. The `--pep_563` option makes monkeytype to insert also `from __future__ import annotations` to altered files (needed in Python <3.11) and to wrap imports needed for signatures in `if TYPE_CHECKING:` conditional blocks. 

[PEP563](https://peps.python.org/pep-0563/#rationale-and-goals) is for solving two problems:
 
>    - forward references: when a type hint contains names that have not been defined yet, that definition needs to be expressed as a string literal;
>    - type hints are executed at module import time, which is not computationally free.

[...]

PR #706, opened 2023-05-19
Test CLI startup time with CI/CD job

juhoinkinen:
Adds a GitHub Actions job with a shell script for measuring the CLI startup time to prevent its unnoticed deterioration.

The startup time was optimized in #696, and now it is good make sure the startup remains fast. For example, I noticed that adding type hints (planned in #690) [can slow the startup, but this can be avoided by postponing the evaluation of annotations](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations).

The timing script executes `annif --help` four times and measures user + sys time. If the average of the user + sys time is over a threshold, the job fails. The threshold is set to 0.300 s.

In the [five previous runs](https://github.com/NatLibFi/Annif/actions/runs/5022302154) the measured time was 0.245 s, 0.240 s, 0.252 s, 0.242 s and 0.245 s, but in some earlier runs the measured time was over 0.300 s. So failures can be expected, however I think it does not matter if they are reasonably rare (say ~10% of runs).

Failure of this job does not make the other jobs fail, so unit testing, linting and publishing are run in any case.

### DIFF
--- a/tests/time-startup.sh
+++ b/tests/time-startup.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Function to measure startup time
+measure_startup_time() {
+    startup_time=$( { time -p annif --help >/dev/null; } 2>&1 | awk '/^user/{u=$2}/^sys/{s=$2} END{print u+s}' )
+    echo "$startup_time"
+}
+
+startup_time1=$(measure_startup_time)
+startup_time2=$(measure_startup_time)
+startup_time3=$(measure_startup_time)
+startup_time4=$(measure_startup_time)
+
+# Calculate the average startup time
+average_startup_time=$(echo "scale=3; ($startup_time1 + $startup_time2 + $startup_time3 + $startup_time4) / 4" | bc)
+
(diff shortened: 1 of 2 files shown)


tests/time-startup.sh
```diff
@@ -0,0 +1,27 @@
...
+startup_time1=$(measure_startup_time)
+startup_time2=$(measure_startup_time)
+startup_time3=$(measure_startup_time)
+startup_time4=$(measure_startup_time)
+
+# Calculate the average startup time
+average_startup_time=$(echo "scale=3; ($startup_time1 + $startup_time2 + $startup_time3 + $startup_time4) / 4" | bc)
+
+# Print the average startup time
+echo "Average Startup time: $average_startup_time seconds"
```
osma, line 18: Maybe print the threshold here as well?

osma (approved):
LGTM, very nice guard against slow imports creeping in!
I gave a minor suggestion as a comment.

PR #713, opened 2023-06-09
Scheduled weekly rebuild of Docker image of latest release

juhoinkinen:
It might be a good idea to rebuild Docker images of "maintained" releases regularly for package updates. We could consider only the most recent release to be "maintained" in this sense (for now at least).

This PR adds a weekly run GH Actions workflow, which

1. checkouts the git tag of the latest release
2. builds image and tests it with pytest
3. pushes the image

Currently the image is pushed to my [DockerHub account](https://hub.docker.com/repository/docker/jinkinen/annif/tags?page=1&ordering=last_updated), and the schedule does not work because the (cron) workflow is not in the main branch.

### DIFF
--- a/.github/workflows/cron-rebuild.yml
+++ b/.github/workflows/cron-rebuild.yml
@@ -1,0 +1,53 @@
+name: "Docker rebuild"
+on:
+  push:  # TODO Remove
+    branches:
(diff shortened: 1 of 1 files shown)


juhoinkinen:
Closed in favor of #715.

PR #715, opened 2023-06-20
Add GH Actions workflow for rebuilding the Docker image of the latest release

juhoinkinen:
This is a simplified version of PR #713: instead of doing rebuilds automatically with a schedule, this just intends to use the "workflow dispatch" trigger to manually start the build.

The Docker images of "maintained" releases should be rebuild regularly for system package updates, but this can be done manually (and for Annif deployments it is probably a good idea to have manual control on when updates happen).

For now is probably enough to consider only the most recent release to be "maintained" in this sense, but [any tag (since merging this PR) can selected for rebuild](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#running-a-workflow).

This PR adds a GH Actions workflow, which

1. checkouts the git tag of the selected release
2. builds the image and tests it with pytest
3. pushes the image to quay.io with these tags of the release:
    -  `<major>.<minor>`
    -  `<major>.<minor>.<patch>`
    -  `<major>.<minor>.<patch>-<YYYYMMDD>` (date of build)

The date suffix allows pinning to a particular build of Annif's Docker image (assuming only one build per day), which is necessary also to retain a tag on an image, which would otherwise be purged from the quay.io repository, making pinning with manifest digests impossible.

The date suffix is also added to the image tags when doing the (initial) image builds for Annif releases.

### DIFF
--- a/.github/workflows/docker-rebuild.yml
+++ b/.github/workflows/docker-rebuild.yml
@@ -41,6 +41,7 @@ jobs:
           latest=false
         tags: |
           type=semver,pattern={{version}}
+          type=semver,pattern={{version}},suffix=-{{date 'YYYYMMDD'}}
           type=semver,pattern={{major}}.{{minor}}
     - name: Build and push to Quay.io
       uses: docker/build-push-action@44ea916f6c540f9302d50c2b1e5a8dc071f15cdf  # v4.1.0
--- a/.github/workflows/docker-rebuild.yml
+++ b/.github/workflows/docker-rebuild.yml
@@ -1,6 +1,5 @@
 name: "Docker rebuild"
-on:
-  workflow_dispatch:
+on: workflow_dispatch
 jobs:
   rebuild-docker-images:
     name: "Docker rebuild"
--- a/.github/workflows/docker-rebuild.yml
+++ b/.github/workflows/docker-rebuild.yml
@@ -20,6 +20,7 @@ jobs:
     - name: "Build for testing"
       uses: docker/build-push-action@c56af957549030174b10d6867f20e78cfd7debc5  # v3.2.0
       with:
+        context: .
         push: false
         tags: test-image
     - name: "Test with pytest"
@@ -46,6 +47,7 @@ jobs:
     - name: Build and push to Quay.io
       uses: docker/build-push-action@44ea916f6c540f9302d50c2b1e5a8dc071f15cdf  # v4.1.0
       with:
+        context: .
         push: true
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-rebuild.yml
```diff
@@ -0,0 +1,50 @@
+name: "Docker rebuild"
+on:
+  workflow_dispatch:
```
osma, line 3: Is the colon here needed?

.github/workflows/docker-rebuild.yml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: "Checkout most recent tag"
+      run: |
+        git fetch --tags origin
+        git describe --abbrev=0 | xargs git checkout
```
osma, line 16: Are we sure this always corresponds to the most recent release? I suppose this will always be triggered from the main branch, so it should give the last tag on the main branch. But there could be more recent tags on other branches, for example patch releases like `v0.61.1`.

At some point we had tags like `v0.55-dev` but those are no longer used AFAICT.

osma (approved):
LGTM. I suppose you intend to close #713 in favour of this PR.

A couple of questions in the comments.

.github/workflows/docker-rebuild.yml
juhoinkinen: Only now I noticed that it is possible to select the tag to be used as an input to the `workflow_dispatch`: 
image

 I'll check how that could be used for the build step, so this a flaky way to select the tag for latest release can be removed, and the Annif version for which Docker image to be rebuilt could be freely selected.

Unfortunately the  `workflow_dispatch` trigger can be used only when the workflow in question is already in the default branch, so testing this while in PR is not possible.

.github/workflows/docker-rebuild.yml
juhoinkinen: Was not needed

juhoinkinen:
As noted in https://github.com/NatLibFi/Annif/pull/715#discussion_r1238128712, it is now made possible to select the wanted git tag when triggering the `workflow_dispatch` (I tested this with [my own fork](https://github.com/juhoinkinen/Annif/actions/workflows/docker-rebuild.yml)).

However this is possible only for the tags from the point since the rebuild workflow is added to the main branch, i.e. for the next release on. Luckily the Docker image for Annif v0.60 got rebuild when working on this PR.

juhoinkinen:
> EDIT: but again the current development version of Annif ends up in the image. No good to merge.

So stupid: ofc the current development version ends up in the image, because the tag which I created for testing this was on the current development version (1.0.0-dev).

PR #719, opened 2023-07-06
Check models compatibility in Docker rebuild GH Actions workflow

juhoinkinen:
The recently added GH Actions workflow for rebuilding Docker images (#715) could also verify that the models trained on the previous image build work (results-wise) identically in the new image. It is quite undesirable that the models would work even slightly differently in different Docker image builds of the same Annif version.

These are the steps in the workflow that aim to verify the compatibility and results identicality of models:

1. Train models with all (trainable) algorithms using old image (the one in quay.io with the tag being rebuild)
2. Evaluate the models and store results in `eval.prev.out` file using old image
3. Evaluate the models and store results in `eval.out` file using new image
4. Compare `eval.prev.out` and `eval.out` with `diff`, and fail the workflow in the case of difference, unless the box for allowing this is checked when triggering the workflow

For both training and evaluation the `tests/corpora/archaeology/fulltext/` corpus is used, which is fine for all algorithms I think. Although there could be some dedicated corpora for this. 

Also, there could be a similar workflow for checking models compatibility when preparing an Annif release, instead of doing compatibility checks manually, so the compatibility-check steps could be moved to a separate action for reusability, like the [prepare action](https://github.com/NatLibFi/Annif/blob/main/.github/actions/prepare/action.yml) of CI/CD workflow.

Note: I've been working on this [in my own fork](https://github.com/juhoinkinen/Annif/tree/docker-rebuild-check-models-compatibility), to avoid accidental image pushes to quay.io. 

TODO before merge: 
- [ ] Switch the image used to compare current build to `quay.io/natlibfi/annif` from `jinkinen/annif`

### DIFF
--- a/projects.cfg.dist
+++ b/projects.cfg.dist
@@ -106,24 +106,6 @@ backend=mllm
 analyzer=snowball(swedish)
 vocab=yso
 
-[stwfsa-fi]
-name=STWFSA YSO Finnish
-language=fi
-backend=stwfsa
-vocab=yso
-
-[stwfsa-sv]
-name=STWFSA YSO Swedish
-language=sv
-backend=stwfsa
-vocab=yso
-
-[stwfsa-en]
-name=STWFSA YSO English
-language=en
-backend=stwfsa
-vocab=yso
-
 [yake-fi]
 name=YAKE Finnish
 language=fi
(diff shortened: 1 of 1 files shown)


juhoinkinen:
When #762 is merged, the upload/download functionality could be utilized for the models compatibility check. By downloading models (maybe from GitHub Actions cache?) to check this first step could be omitted:

> 1. Train models with all (trainable) algorithms using old image (the one in quay.io with the tag being rebuild)

juhoinkinen:
#907 is a better implementation for the compatibility checking, I close this PR.

PR #721, opened 2023-07-31
Suppress TensorFlow info messages to debug level

### LINKED ISSUE #720: Annoying TensorFlow messages
I've seen these for a long time but just ignored them. However, I think these messages should not be shown to the user:

```
2023-07-25 11:31:43.949391: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2023-07-25 11:31:44.032688: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
```

The messages are shown for most Annif commands (e.g. `annif suggest`, `annif eval`) whenever TensorFlow is installed, even without having configured any projects using the `nn_ensemble` backend.

juhoinkinen:
Makes TF log level to be set by Annif log level using `TF_CPP_MIN_LOG_LEVEL` env variable.

Mainly used for hiding distracting TF INFO messages. Closes #720. Based on [this SO answer](https://stackoverflow.com/a/42121886).

The env setting is in the `annif/__init__.py` to make the log level change work also when TF is used by spaCy analyzer (which was not  apparent at first), in addition to TF usage by NN ensemble backend.  

I am not seeing the second message mentioned in the issue: 

    2023-07-25 11:31:44.032688: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.

I hope also this message arises during TF import, so this PR should fix that part of the issue too (Edit: no, it did not, but the current approach should take care it). Could @osma verify this?

### DIFF
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING
 logging.basicConfig()
 logger = logging.getLogger("annif")
 logger.setLevel(level=logging.INFO)
+os.environ.setdefault("TF_CPP_MIN_LOG_LEVEL", "1")  # Hide TF info messages by default
 
 import annif.backend  # noqa
 
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -85,8 +85,8 @@ def _set_tensorflow_loglevel():
         0: "0",  # NOTSET
         10: "0",  # DEBUG
         20: "1",  # INFO
-        30: "2",  # WARNING
-        40: "3",  # ERROR
+        30: "1",  # WARNING
+        40: "2",  # ERROR
         50: "3",  # CRITICAL
     }
     tf_loglevel = tf_loglevel_mapping[annif_loglevel]
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -80,6 +80,9 @@ def _get_config_name(config_name: str | None) -> str:
 
 
 def _set_tensorflow_loglevel():
+    """Set TensorFlow log level based on Annif log level (--verbosity/-v
+    option) using an environment variable. INFO messages by TF are shown only on
+    DEBUG (or NOTSET) level of Annif."""
     annif_loglevel = logger.getEffectiveLevel()
     tf_loglevel_mapping = {
         0: "0",  # NOTSET
--- a/annif/__init__.py
+++ b/annif/__init__.py
@@ -10,7 +10,6 @@ from typing import TYPE_CHECKING
 logging.basicConfig()
 logger = logging.getLogger("annif")
 logger.setLevel(level=logging.INFO)
-os.environ.setdefault("TF_CPP_MIN_LOG_LEVEL", "1")  # Hide TF info messages by default
 
 import annif.backend  # noqa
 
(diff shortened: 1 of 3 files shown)


osma:
I tested this, but it didn't seem to do anything. After merging this PR branch to my local main branch, I'm seeing this:

```
(annif-py3.10) ~/git/Annif$ annif eval fasttext-spacy tests/corpora/archaeology/fulltext/
2023-07-31 14:53:46.253944: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F AVX512_VNNI FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2023-07-31 14:53:46.327325: I tensorflow/core/util/port.cc:104] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
```

juhoinkinen:
Well this should work even for TensorFlow usage by spaCy.

osma (approved):
Now it works even with spaCy!

osma:
One question though: would it make sense to set the TF logging level based on the Annif (`logging` module) log level?

juhoinkinen:
> One question though: would it make sense to set the TF logging level based on the Annif (`logging` module) log level?

It would be nice, but then the question is how the levels should be aligned: the INFO messages of TF are the ones we want to hide, while at the same time show Annif INFO messages(?).

One way for this would be to make TF log level "one more" than Annif log level (while Annif shows INFO+WARN+ERROR messages, TF shows WARN+ERROR messages; while Annif shows DEBUG+INFO+WARN+ERROR messages, TF shows INFO+WARN+ERROR).

osma:
The "one more" idea sounds very good! I think it's reasonable to "tone down" log messages from TF by one level. They may sometimes be relevant for debugging problems, so making it possible (but not the default) to see messages from TF would be a nice feature.

juhoinkinen:
Now the feature works for regular commands: TensorFlow INFO messages are shown only when running Annif at DEBUG log level.
```
annif list-projects -v DEBUG
debug: creating flask app with configuration annif.default_config.Config
debug: Reading configuration file projects.cfg in CFG format
2023-08-02 10:50:42.339259: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
```

But I could not make it work in API server mode, I mean with `annif run` or run via gunicorn. I think it should have worked, according to a [SO answer](https://stackoverflow.com/a/20894863). So there might be something out of tune in the overall logging functionality.

juhoinkinen:
The test for this needs to fixed... I'll try something with `@mock.patch.dict(os.environ, ...)` decorator.

Issue #583, opened 2022-03-21
Train command option --prepare-only


osma:
For many backends, training consists of two distinctive steps: preparing the training data, then training a ML model. It is possible to reuse the prepared training data using the `--cached` option. But there is no support for the inverse operation: just preparing the data without training the model. This could be useful for DVC workflows (allowing more granular pipeline stages) and in situations where you intend to perform hyperparameter optimization.

The proposal is to add a `--prepare-only` option to the `annif train` command, which skips the model training part in those backends where there is a separate preparation step (i.e. those backends that currently support the `--cached` option).

PR #722, opened 2023-08-04
Fix train state and modification time for unfinished project training

juhoinkinen:
If initial training of a project is not finished after any files have been created in the project's data directory, the train state and modification time information turn out incorrect (the project shows to be (fully) trained when it is not, and with a modification time). And when retraining a project is interrupted, the modification time is falsely updated.

This problem can realize more commonly when/if implementing [the `--prepare-only` option to the train command](https://github.com/NatLibFi/Annif/issues/583).

This PR makes the methods inquiring the train state and modification time to ignore files in the project's datadir with pattern `*-train*`, `tmp-*` and  `vectorizer`. The `tmp-` prefix is added to all temporary files, because some backends are using a tempfile for the model file during training, which can remain after unfinished training, e.g.  `stwfsa_predictor1mz8z4im.zip`.

The train and temp files should definitely be ignored, but the vectorizer file case is not so clear:
- a project is not trained when only the vectorizer file exists (PR implements this)
- when the vectorizer file (of a fully trained project) changes, the results of the model change, so modification time should be based also on vectorizer file mtime (PR **does not** implements this)

Instead of using global ignore patterns, this functionality could use the actual model file names/patterns per backend, but the field storing them varies a bit (`MODEL_FILE`, `INDEX_FILE`, `MODEL_FILE_PREFIX`).

### DIFF
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -52,14 +52,25 @@ class AnnifBackend(metaclass=abc.ABCMeta):
         return params
 
     @property
+    def _model_file_paths(self) -> list:
+        all_paths = glob(os.path.join(self.datadir, "*"))
+        ignore_patterns = ("*-train*", "vectorizer")
+        ignore_paths = [
+            path
+            for igp in ignore_patterns
+            for path in glob(os.path.join(self.datadir, igp))
+        ]
+        return list(set(all_paths) - set(ignore_paths))
+
+    @property
     def is_trained(self) -> bool:
-        return bool(glob(os.path.join(self.datadir, "*")))
+        return bool(self._model_file_paths)
 
     @property
     def modification_time(self) -> datetime | None:
         mtimes = [
             datetime.utcfromtimestamp(os.path.getmtime(p))
-            for p in glob(os.path.join(self.datadir, "*"))
+            for p in self._model_file_paths
         ]
         most_recent = max(mtimes, default=None)
         if most_recent is None:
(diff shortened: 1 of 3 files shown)


juhoinkinen:
To make the state of projects more coherent, the first step of the training could be to remove old data files. Then there would not be the problem about the modification time of the `vectorizer` file (whose update without updating the rest of the model could also make the project not working either silently or at all, because of wrong index): it would be clear that the `vectorizer` file would not affect the modification time or train state of the project.

This would also keep the project's datadir clean of any temp files hanging around.

However, this would also mean a less gracious behavior, in the sense that if retraining a working project fails, then the project is gone, whereas now the old project could remain working (depending on the step which fails).

osma (approved):
I think the implemented approach looks sensible.

Regarding this idea:

> To make the state of projects more coherent, the first step of the training could be to remove old data files. Then there would not be the problem about the modification time of the vectorizer file (whose update without updating the rest of the model could also make the project not working either silently or at all, because of wrong index): it would be clear that the vectorizer file would not affect the modification time or train state of the project.
>
>This would also keep the project's datadir clean of any temp files hanging around.
>
>However, this would also mean a less gracious behavior, in the sense that if retraining a working project fails, then the project is gone, whereas now the old project could remain working (depending on the step which fails).

I think that this would be a step in the wrong direction. It would be better to try to make train operations more atomic, so that vectorizers, train files, models etc. would first be written into temporary files and only right at the end they would be renamed so that they replace existing files (if any). This is how annif.util.atomic_save is intended to work, but it only applies on individual files and not sets of files that need to be kept in sync (e.g. vectorizer + model).

PR #724, opened 2023-08-10
Upgrade Simplemma & limit its memory usage

osma:
This PR upgrades Simplemma to the latest version 1.1.1 and adds some wrapper code to make it possible to use Simplemma language detection with a limit on the number of languages simultaneously loaded into memory. The number of languages is currently hardcoded at 5.

This will enable adding language detection functionality to the REST API in PR #659 , which has not been merged yet due to concerns about memory usage.

### DIFF
--- a/annif/analyzer/simplemma.py
+++ b/annif/analyzer/simplemma.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-import simplemma
-
-import annif.langsupport
+import annif.simplemma_util
 
 from . import analyzer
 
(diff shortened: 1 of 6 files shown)


juhoinkinen:
In PR #796 Simplemma is updated to 1.1, so this could be merged after that.

There I already did some neccessary changes to the language detection code: https://github.com/NatLibFi/Annif/pull/796/commits/49d50520ee136dc159538f65bd3f810a0e4efae2

osma:
Rebased on current `main` and force-pushed.

juhoinkinen (approved):
LGMT

PR #725, opened 2023-08-10
Increase maximum startup time threshold (0.3 -> 0.4 s)

osma:
We have a CI test for startup time, implemented in PR #706, designed to prevent situations where basic Annif CLI commands suddenly take a long time due to e.g. slow imports.

The threshold value was set to 0.300 seconds. But nowadays this test quite often fails because the startup takes a bit longer, e.g. 0.35 seconds. I don't think anything has changed with the imports, it's just that the CI environment runs a bit slower than it used to.

This PR increases the threshold value to 0.400 seconds, which should prevent spurious test failures but still ensure that startup times have not increased dramatically.


osma:
Ironically, the CI startup time test failed because startup took .432 seconds.

Issue #532, opened 2021-10-07
Update to Optuna 3.x to get rid of deprecation message


osma:
In recent versions of Annif (which use Optuna 2.8.0), running the `annif hyperopt` command with a `--jobs` parameter gives the following warning:

    FutureWarning: `n_jobs` argument has been deprecated in v2.7.0. This feature will be removed in v4.0.0. See https://github.com/optuna/optuna/releases/tag/v2.7.0.

The Optuna [documentation for the 2.7.0 release](https://github.com/optuna/optuna/releases/tag/v2.7.0) explains that process-based parallelism should be used instead ([tutorial](https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/004_distributed.html#sphx-glr-tutorial-10-key-features-004-distributed-py)). But that would require a PostgreSQL or MySQL database (SQLite is possible but not recommended).

Another option is to keep using thread-level parallelism using ThreadPoolExecutor instead of the `n_jobs` parameter. That would be easier to implement at least.

osma:
It seems that Optuna has decided to continue supporting the `n_jobs` parameter after all: https://github.com/optuna/optuna/issues/2937
It is also possible that in the future it will become easier to run Optuna in parallel using real multiprocessing: https://github.com/optuna/optuna/issues/2950

Both of these are on the [Optuna version 3 roadmap](https://github.com/optuna/optuna/wiki/Optuna-V3-Roadmap).

So for now I think we should wait and see what happens on the Optuna side.

juhoinkinen:
> In recent versions of Annif (which use Optuna 2.8.0), running the `annif hyperopt` command with a `--jobs` parameter gives the following warning:
> 
> ```
> FutureWarning: `n_jobs` argument has been deprecated in v2.7.0. This feature will be removed in v4.0.0. See https://github.com/optuna/optuna/releases/tag/v2.7.0.
> ```

When updating now to latest Optuna release, i.e. v3.3.0, the above is not shown, but a different one is:
```
FutureWarning: suggest_uniform has been deprecated in v3.0.0. This feature will be removed in v6.0.0. See https://github.com/optuna/optuna/releases/tag/v3.0.0. Use suggest_float instead.
```

[...]

PR #726, opened 2023-08-14
Update dependencies v1.0

### LINKED ISSUE #697: Pin Flask and rdflib major/minor versions
Currently Flask and rdflib are loosely pinned in pyproject.toml:
```
flask = ">=1.0.4,<3"
...
rdflib = ">=4.2,<7.0"
```

I meant to pin all major/minor versions of all non-dev dependencies [as suggested](https://github.com/NatLibFi/Annif/pull/683#issuecomment-1477425251) in #683, but missed these two.

Lets do it for the next release, however wait for connexion 3 release, as the current connexion version 2.14 dependens on [not-latests Flask version](https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/setup.py#L36).

juhoinkinen:
Updates most of the outdated dependencies, and pins Flask and rdflib versions more tightly. 

Exceptions to updating to most-recent releases:
- Flask is pinned to 2.2.* instead 2.3.* due to [the requirement by Connexion v2.14.2](https://github.com/spec-first/connexion/blob/e8408406f6381657c7f0dc18b34bfd7efd980d91/setup.py#L36)
- rdflib is pinned 6.3.* instead of 7.* due to [the requirement by stwfsapy](https://github.com/zbw/stwfsapy/blob/master/pyproject.toml#L17)
- Updating to numpy 1.25.* would [require Python 3.9-3.11](https://github.com/numpy/numpy/releases/tag/v1.25.0)
- Updating to scipy 1.11.* would [require Python 3.9-3.11](https://github.com/scipy/scipy/releases/tag/v1.11.0)
- ~Updating to [Optuna 3.3.0](https://github.com/optuna/optuna/releases/tag/v3.3.0) (from 2.10.1) is left to wait for the possible implementation changes in Annif (see issue #532)~

Also closes #697 and #532.


juhoinkinen:
When updating to scikit-learn v1.3.0 old MLLM models cease to work:
```
...

  File "/home/local/jmminkin/git/Annif/annif/backend/backend.py", line 142, in suggest
    self.initialize()
  File "/home/local/jmminkin/git/Annif/annif/backend/mllm.py", line 119, in initialize
    self._model = self._load_model()
  File "/home/local/jmminkin/git/Annif/annif/backend/mllm.py", line 102, in _load_model
    return MLLMModel.load(path)
  File "/home/local/jmminkin/git/Annif/annif/lexical/mllm.py", line 366, in load
    return joblib.load(filename)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/joblib/numpy_pickle.py", line 658, in load
    obj = _unpickle(fobj, filename, mmap_mode)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/joblib/numpy_pickle.py", line 577, in _unpickle
    obj = unpickler.load()
  File "/usr/lib/python3.8/pickle.py", line 1212, in load
    dispatch[key[0]](self)
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/joblib/numpy_pickle.py", line 402, in load_build
    Unpickler.load_build(self)
  File "/usr/lib/python3.8/pickle.py", line 1705, in load_build
    setstate(state)
  File "sklearn/tree/_tree.pyx", line 714, in sklearn.tree._tree.Tree.__setstate__
  File "sklearn/tree/_tree.pyx", line 1418, in sklearn.tree._tree._check_node_ndarray
ValueError: node array from the pickle has an incompatible dtype:
- expected: {'names': ['left_child', 'right_child', 'feature', 'threshold', 'impurity', 'n_node_samples', 'weighted_n_node_samples', 'missing_go_to_left'], 'formats': ['<i8', '<i8', '<i8', '<f8', '<f8', '<i8', '<f8', 'u1'], 'offsets': [0, 8, 16, 24, 32, 40, 48, 56], 'itemsize': 64}
- got     : [('left_child', '<i8'), ('right_child', '<i8'), ('feature', '<i8'), ('threshold', '<f8'), ('impurity', '<f8'), ('n_node_samples', '<i8'), ('weighted_n_node_samples', '<f8')]
```

osma:
Good catch re: scikit-learn and MLLM. There's not much we can do about it, I'm afraid. I think it's better to update now, before 1.0, instead of postponing the inevitable. But it has to be mentioned in the release notes.

The scikit-learn documentation on Model Persistence has some [notes](https://scikit-learn.org/stable/model_persistence.html#interoperable-formats) about compatibility between different environments and versions. It mentions the PMML and ONNX formats that could possibly be more durable than serializing sklearn models directly via joblib or pickle, as we currently do in MLLM. But that would be a whole new investigation. I think we should just consider this an unfortunate situation that we couldn't avoid.

juhoinkinen:
> Good catch re: scikit-learn and MLLM. There's not much we can do about it, I'm afraid. I think it's better to update now, before 1.0, instead of postponing the inevitable. But it has to be mentioned in the release notes.
> 
> The scikit-learn documentation on Model Persistence has some [notes](https://scikit-learn.org/stable/model_persistence.html#interoperable-formats) about compatibility between different environments and versions. It mentions the PMML and ONNX formats that could possibly be more durable than serializing sklearn models directly via joblib or pickle, as we currently do in MLLM. But that would be a whole new investigation. I think we should just consider this an unfortunate situation that we couldn't avoid.

Also old stwfsa models won't work with updated skikit-learn, the error message is the same as for MLLM models.

osma:
I'm seeing this new TensorFlow/Keras warning for test_backend_nn_ensemble.py:

    UserWarning: You are saving your model as an HDF5 file via `model.save()`. This file format is considered legacy. We recommend using instead the native Keras format, e.g. `model.save('my_model.keras')

It seems there is a new Keras format (`.keras`) available, which is nowadays the [recommended](https://keras.io/guides/serialization_and_saving/) one for saving Keras models. We are using the legacy HDF5 (`.h5`) format: https://github.com/NatLibFi/Annif/blob/40cc2fd0325b1c648abf4d436f4a05ee83c9599b/annif/backend/nn_ensemble.py#L100

What should we do about this for Annif 1.0?

1. Nothing, just ignore the warning for now
2. Switch to the Keras format in a simplistic way (basically just switching the file extension to `.keras` on the above line)
3. Switch to `.keras` but additionally add fallback code that can load existing `.h5` models as well
4. Same as 3, but additionally add a deprecation warning that the fallback support will be removed in Annif 1.1.

I don't like 1, because it postpones an inevitable problem. Option 2 breaks compatibility with previously trained models, but provides a fresh start. Option 3 adds a few lines of extra code (and tests), while option 4 is more work but also "promises" to remove that extra code in the future.

juhoinkinen:
> What should we do about this for Annif 1.0?
> 
>     1. Nothing, just ignore the warning for now
> 
>     2. Switch to the Keras format in a simplistic way (basically just switching the file extension to `.keras` on the above line)
> 
>     3. Switch to `.keras` but additionally add fallback code that can load existing `.h5` models as well
> 
>     4. Same as 3, but additionally add a deprecation warning that the fallback support will be removed in Annif 1.1.
> 
> 
> I don't like 1, because it postpones an inevitable problem. Option 2 breaks compatibility with previously trained models, but provides a fresh start. Option 3 adds a few lines of extra code (and tests), while option 4 is more work but also "promises" to remove that extra code in the future.

Option 3 seems best to me. Better promise as little as possible.

osma:
> Option 3 seems best to me. Better promise as little as possible.

I initially implemented option 3 in PR #730, but the fallback code (especially the unit test) got messy and then I realized that it's not much use supporting old pre-1.0 NN ensemble models when MLLM and STWFSA models will break anyway due to changes in scikit-learn.

PR #727, opened 2023-08-15
Python 3.11 support

### LINKED ISSUE #703: Python 3.11 support
Python 3.11 was [released](https://www.python.org/downloads/release/python-3110/) in October 2022. We should aim at supporting it in Annif. Some concrete steps:

1. Enable running GitHub Actions CI tests under Python 3.11 (most likely in experimental mode i.e. allowed to fail)
2. Check all the dependencies and see if they work under 3.11
3. `tomllib` ([PEP 690](https://peps.python.org/pep-0680/)) is now part of the standard library so we could use that instead of `tomli`
4. Fix 3.11 compatibility issues in the Annif codebase, if any

It's quite likely that some libraries we depend on don't yet support 3.11, so there will be a transition period, perhaps with some optional dependencies not working. This happened with 3.10 as well.

Not sure whether this can be done before the 1.0 release, so I'm putting this into the "Long term" milestone for now.

juhoinkinen:
Allow installing Annif on Python 3.11, and run unit tests on it in the CI/CD pipeline.

Omikuji dependency does not work Python 3.11, see https://github.com/NatLibFi/Annif/issues/703#issuecomment-1614513903. Other dependencies seem to work (i.e. the unit tests pass).

`tomli` is now a (direct) dependency only on Python <3.11, on Python 3.11 `tomllib` of standard library is used instead ([PEP 690](https://peps.python.org/pep-0680/)). However, also on Python 3.11 `tomli` is still installed as it is a dependency of `black`, `coverage`, `pytest` and `schemathesis`.

Closes #703, includes merge of the branch of #726.

### DIFF
--- a/.github/actions/prepare/action.yml
+++ b/.github/actions/prepare/action.yml
@@ -22,11 +22,12 @@ runs:
         key: ignore-me
         restore-keys: |
           poetry-installation-and-cache-${{ inputs.python-version }}-${{ inputs.poetry-version }}-
-    - name: Install Poetry
-      shell: bash
-      run: |
-        pipx install poetry==${{ inputs.poetry-version }}
     - name: Set up Python ${{ inputs.python-version }}
       uses: actions/setup-python@5ccb29d8773c3f3f653e1705f474dfaa8a06a912  # v4.4.0
       with:
         python-version: ${{ inputs.python-version }}
+    - name: Install Poetry
+      shell: bash
+      run: |
+        pipx install poetry==${{ inputs.poetry-version }}
+        poetry env use ${{ inputs.python-version }}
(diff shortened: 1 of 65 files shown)


osma:
If the unit tests pass on Python 3.11 (except Omikuji, but that is not even installed for 3.11), then I don't see why we should mark the CI tests as experimental. It would be a different situation if some core tests were failing on 3.11 and we couldn't fix them right now.

osma:
The Python 3.11 parts of this PR look good to me, but since this includes all the commits from PR #726, we need to make sure those are good as well before merging this. I noticed a potential problem with TF/Keras, see https://github.com/NatLibFi/Annif/pull/726#issuecomment-1680047565

PR #730, opened 2023-08-16
Switch to Keras v3 save format for nn_ensemble

osma:
With the TensorFlow 2.13 update in PR #726, these warnings are shown:

    UserWarning: You are saving your model as an HDF5 file via `model.save()`. This file format is considered legacy. We recommend using instead the native Keras format, e.g. `model.save('my_model.keras')

This PR addresses the problem by switching to the new Keras v3 (.keras) format for saving the trained nn_ensemble models.

Keras imports are modernized so that they follow current conventions (no need to use `tensorflow.keras`, and switched to the new `keras.saving` submodule).

Note that this PR branch includes all the commits from PR #726.


juhoinkinen (approved):
LGTM

Issue #338, opened 2019-10-24
OpenRefine reconciliation API


wetneb:
Hi!

Would it make sense for an Annif instance to expose an [OpenRefine reconciliation API](https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation-Service-API) for each of the profiles it knows of? That would let users "reconcile" tabular data to topics.

We are working on [documenting the spec in a cleaner way](https://reconciliation-api.github.io/specs/) as part of the [W3C Entity Reconciliation Community Group](https://www.w3.org/community/reconciliation/).

This came up in a meeting with @acka47 and @literarymachine.

osma:
Hi @wetneb , thanks for the suggestion. It would make a lot of sense to implement this - I've actually had the same idea and thought about opening an issue, but you got there first :)

I'm aware of the W3C CG work, I've joined the group myself just to follow what's going on.

osma:
BTW, it would make the implementation easier if there was an OpenAPI (Swagger) spec for the API available. Annif uses Connexion, which is driven by an OpenAPI spec file.

osma:
This would be a lot nicer to implement if the reconciliation spec is modernized to better follow REST conventions, as suggested here: https://github.com/reconciliation-api/specs/issues/84

tombaker:
@webnet @osma @woody544 The USDA National Agricultural Library (NAL) uses Annif and Skosmos, and having Annif or Skosmos serve as a OpenRefine reconciliation API would be very helpful.

Our use case is reconciliation of local CSV files to [NALT](https://agclass.nal.usda.gov/).

Currently we are using a [relative old but workable JAR file](https://github.com/okfn/reconcile-csv) that creates a reconciliation interface for the local CSV. After some digging we have found no more up-to-date alternatives. There is a [Python module](https://github.com/gitonthescene/csv-reconcile) that seemingly does the same thing, but we could not get it to run properly.

tombaker:
See also https://github.com/NatLibFi/Skosmos/issues/23#issuecomment-1168414524

PR #734, opened 2023-09-08
Add reconciliation API endpoint to REST API

UnniKohonen:
This PR adds a [Reconciliation API](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#service-manifest) endpoint to REST API as mentioned in #338. It allows reconciling against a project using the `/v1/projects/{project_id}/reconcile` API method.

The endpoint accepts `GET` and `POST` requests:
- A `GET` request with no parameters returns a [service manifest](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#service-manifest).
- A`GET` request with a URL query parameter [`queries`](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#structure-of-a-reconciliation-query) serialized as JSON returns a [reconciliation result batch](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#reconciliation-query-responses).
- A `POST` request with an `application/x-www-form-urlencoded` parameter [`queries`](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#structure-of-a-reconciliation-query) serialized as JSON returns a [reconciliation result batch](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#reconciliation-query-responses).

Only the fields `query` and `limit` are supported in the `queries` objects. Reconciliation result objects contain the fields `id`, `name`, `score` and `match`. Reconciliation doesn't support types or properties for now.

The reconciliation API specification requires the usage of the [`identifierSpace`](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#dfn-identifier-space) field in the service manifest. Since Annif doesn't require entities to be in a specific URI space and doesn't have a way to extract this infomation from a project, the field is left empty for now.

### DIFF
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -184,26 +184,38 @@ paths:
     get:
       tags:
       - Reconciliation
-      summary: reconcile against a project
+      summary: get reconciliation service manifest or reconcile against a project
       operationId: annif.rest.reconcile_metadata
       parameters:
       - $ref: '#/components/parameters/project_id'
-      # queries parameter doesn't work for some reason
       - in: query
+        description: A call to the reconciliation service API
         name: queries
+        required: false
         schema:
-          type: object
+          type: string
           additionalProperties:
             type: object
             required:
-            - query
+              - query
             properties:
               query:
                 type: string
-                description: query string
-          example: 
-            {"q0": {"query": "query"}}
-        required: false
+                description: Query string to search for
+              limit:
+                type: integer
+                description: Maximum number of results to return
+        example:
+          '{
+            "q0": {
+              "query": "example query",
+              "limit": 10
+            },
+            "q1": {
+              "query": "another example",
+              "limit": 15
+            }
+          }'
       responses:
         "200":
           description: successful operation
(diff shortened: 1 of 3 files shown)


osma:
Excellent work!

> The reconciliation API specification requires the usage of the [identifierSpace](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#dfn-identifier-space) field in the service manifest. Since Annif doesn't require entities to be in a specific URI space and doesn't have a way to extract this infomation from a project, the field is left empty for now.

I opened an issue about this problem: https://github.com/reconciliation-api/specs/issues/139

I will soon provide more detailed comments about the code.

annif/rest.py
```diff
@@ -214,3 +215,72 @@ def learn(
         return server_error(err)
 
     return None, 204
+
+
+def _reconcile(project_id: str, query: dict[str, Any]) -> dict[str, Any]:
```
osma, line 220: The return type hint `dict` doesn't seem to match the actual returned type, which is a list of dicts.

(found by SonarCloud)

annif/rest.py
```diff
@@ -214,3 +215,72 @@ def learn(
...
+
+def reconcile(
+    project_id: str, body: dict[str, Any]
+) -> ConnexionResponse | dict[str, Any]:
+    """reconcile against a project and return a dict with results
+    formatted according to OpenAPI spec"""
+
+    queries = body["queries"]
+    results = {}
+    for key, query in queries.items():
```
osma, line 280: Using a for loop here means that the reconciliation isn't done in batches internally. It would be better (more efficient) to pass the whole batch to _suggest at once, similar to how the suggest_batch method works.

annif/openapi/annif.yaml
```diff
@@ -314,6 +460,105 @@ components:
                 type: string
                 example: Vulpes vulpes
       description: A document with attached, known good subjects
+    ReconcileMetadata:
```
osma, line 463: It could be better to call this type ReconciliationServiceManifest (although it's pretty long...) so that it matches the reconciliation API spec.

osma (changes requested):
See the detailed line-by-line comments.

In addition, tests/test_openapi.py is [failing](https://github.com/NatLibFi/Annif/actions/runs/6121900589/job/16616589806?pr=734) on GitHub Actions as well as for me locally. There seems to be some problem with the `queries` parameter. Possibly the OpenAPI spec needs to be adjusted somehow.

Issue #735, opened 2023-09-13
Dealing with overrepresented concepts / blacklisting


annakasprzik:
Several institutions have observed that some models / ensembles struggle with concepts that are overrepresented in the training data so that they are suggested way to often. One fix for that is to identify rules that limit the contexts in which those concepts can be suggested. Could we implement something in Annif that allows specifying those rules?

CC @schlawiner @lakshmi-bashyam

related: https://github.com/NatLibFi/Annif/issues/538 ; https://github.com/NatLibFi/Annif/issues/596

osma:
Thank you for the suggestion. Indeed this seems like a recurring problem, so a generic mechanism could be useful.

This was one of the ideas discussed in issue #538, especially in https://github.com/NatLibFi/Annif/issues/538#issuecomment-979109206 . But there were maybe too many ideas thrown around and so far nothing has been implemented. So let's keep this issue focused on only the problem of overrepresented concepts and the possible solution to make it possible to block problematic concepts, since it seems that both ZBW and ZPID have already decided to use such a mechanism implemented outside Annif.

I think this configuration example from https://github.com/NatLibFi/Annif/issues/538#issuecomment-976532224 is still valid:

```
[omikuji_stw_en]
vocab=stw_9_10
exclude_concepts=http://zbw.eu/stw/descriptor/19073-6,http://zbw.eu/stw/descriptor/17829-1
backend=omikuji
```

and the meaning of this would be that the two concepts (USA and Theory) listed in `exclude_concepts` are ignored both when reading/processing training data and when generating suggestions, but only for this particular project. There could still be other projects using the same vocabulary and the setting would of course not affect those. So in an ensemble, it would be possible to block specific concepts on the level of a particular backend project, if it has a tendency to suggest certain concepts too often without good reason.

As noted in #538, it would make sense to avoid the term "blacklisting" due to [connotations](https://www.splunk.com/en_us/blog/learn/blacklist-whitelist-inclusivity.html). I think "exclude", "block" or "deny" are all valid alternatives.

osma:
I've thought about the best way to implement something like this in Annif code.

I think this should be a general mechanism and ideally no changes to individual backend implementations should be necessary. This means that the setting should be handled on the level of AnnifProject. One possibility is that SubjectIndex would be made aware of the blocked/excluded concepts, similar to how it already handles deprecated concepts.

[...]

PR #736, opened 2023-09-22
optimization: load a vocabulary only once even if used in different languages

osma:
While looking at ways to implement #735, I discovered an opportunity for optimization in the registry code that handles loading of vocabularies. For some reason (probably my mistake) the registry loads vocabularies multiple times, once per language. This amounts to useless work and use of memory.

This PR adjusts the code slightly so that vocabularies are always loaded just once. This was always the intention since the introduction of multilingual vocabularies (#559, PR #600 etc.) and especially PR #610 which implemented vocabularies that are shared between projects.

I benchmarked this with an installation where I have three Finto AI MLLM projects (languages fi, sv, en) that all use the YSO vocabulary, but in different languages. I ran the command

    ANNIF_CONFIG=annif.default_config.ProductionConfig /usr/bin/time -v annif list-projects

The idea here is to use ProductionConfig which causes all projects to be loaded on startup, instead of on demand. This means that also the vocabulary is loaded.

# Before

(showing selected stats)

```
        User time (seconds): 13.04
	System time (seconds): 6.13
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:12.27
	Maximum resident set size (kbytes): 539600
```

# After

```
	User time (seconds): 12.82
	System time (seconds): 7.26
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:11.66
	Maximum resident set size (kbytes): 428940
```

So there's a slight speedup, and the memory usage drops by 110MB. Not bad for a patch that also reduces the amount of code by 3 lines.

### DIFF
--- a/annif/registry.py
+++ b/annif/registry.py
@@ -99,14 +99,11 @@ class AnnifRegistry:
         vocab_id = match.group(1)
         posargs, kwargs = parse_args(match.group(3))
         language = posargs[0] if posargs else default_language
-        vocab_key = (vocab_id, language)
 
         self._init_vars()
-        if vocab_key not in self._vocabs[self._rid]:
-            self._vocabs[self._rid][vocab_key] = AnnifVocabulary(
-                vocab_id, self._datadir
-            )
-        return self._vocabs[self._rid][vocab_key], language
+        if vocab_id not in self._vocabs[self._rid]:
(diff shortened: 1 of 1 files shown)


juhoinkinen (approved):
Good find

Issue #737, opened 2023-11-02
Error by Voikko analyzer: `ValueError: character U+fb50067 is not in range [U+0000; U+10ffff]`


juhoinkinen:
Word normalization by Voikko has failed at least on two occasions when using Annif v1.0 at Finto AI, which has resulted in 500 error response for the suggest request. On the occasions the error has been reproducible by sending suggest request for just the triggering word. Once noticed, the situation has been fixed by restarting Annif, after which the word has no longer produced the error.

The logs have show the following error (the character code has been different ):<details>
<summary>(Click to expand)</summary>

```
ValueError: character U+fb50067 is not in range [U+0000; U+10ffff]
    value = self.__lib.voikko_mor_analysis_value_ucs4(cAnalysis, key)
    File "/usr/local/lib/python3.10/site-packages/voikko/libvoikko.py", line 682, in analyze
      result = self.voikko.analyze(word)
    File "/Annif/annif/analyzer/voikko.py", line 30, in _normalize_word
      self._normalize_word(word)
    File "/Annif/annif/analyzer/analyzer.py", line 48, in <listcomp>
      return [
    File "/Annif/annif/analyzer/analyzer.py", line 47, in tokenize_words
      doc = tokenizer(doc)
    File "/usr/local/lib/python3.10/site-packages/sklearn/feature_extraction/text.py", line 112, in _analyze
      for feature in analyze(doc):
    File "/usr/local/lib/python3.10/site-packages/sklearn/feature_extraction/text.py", line 1276, in _count_vocab
      _, X = self._count_vocab(raw_documents, fixed_vocab=True)
    File "/usr/local/lib/python3.10/site-packages/sklearn/feature_extraction/text.py", line 1434, in transform
      sent_tokens = vectorizer.transform(sentences)
    File "/Annif/annif/lexical/mllm.py", line 87, in generate_candidates
      return generate_candidates(text, analyzer, self._vectorizer, self._index)
    File "/Annif/annif/lexical/mllm.py", line 156, in generate_candidates
      return self._model.generate_candidates(text, self.project.analyzer)
    File "/Annif/annif/backend/mllm.py", line 153, in _generate_candidates
      candidates = self._generate_candidates(text)
    File "/Annif/annif/backend/mllm.py", line 166, in _suggest
      [self._suggest(text, params) for text in texts],
    File "/Annif/annif/backend/backend.py", line 129, in <listcomp>
      [self._suggest(text, params) for text in texts],
[...]

PR #738, opened 2023-11-02
Debug ValueError by Voikko analyzer 

juhoinkinen:
This aims to help to debug why the ValueError by Voikko analyzer arises (#737).

When this is merged, the latest Docker image could be deployed to ai.dev.finto.fi, which could then be used to investigate the situation.

Is there anything else that could be logged, in addition to the triggering word, @osma?

### DIFF
--- a/annif/analyzer/analyzer.py
+++ b/annif/analyzer/analyzer.py
@@ -5,6 +5,8 @@ import abc
 import functools
 import unicodedata
 
+from annif.exception import OperationFailedException
+
 _KEY_TOKEN_MIN_LENGTH = "token_min_length"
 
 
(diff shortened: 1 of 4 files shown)


osma:
I think it would be helpful to be able to log more contextual information, but this might be a bit difficult since it would require accessing the call stack. At least in the traceback in #737 it seems that the error got triggered via the MLLM backend. So in that case it would be good to log the sentence that was being tokenized and maybe even the whole document text.

juhoinkinen:
Now the text being tokenized (full document) should be logged too.

juhoinkinen:
I build a Docker image from this branch, deployed it to ai.dev.finto.fi and started sending suggest requests continuously with Annif-client (with contents from Kirjastonhoitaja dataset documents). Hopefully the error occurs and we get some more clues about it.

I thought using some fuzzing tool, but this was an easy way.

Maybe this PR does not even need to be merged to main, if the root cause of the error and a fix to it can be found.

osma:
I think that text extracted from random PDFs, ideally using different text extraction tools, might be a better (more fuzzy) data source. But this is a good start, let's see if it will trigger the bug.

juhoinkinen:
The Voikko error has not occurred in a long time. Hopefully it's gone for good. I close this PR.

PR #745, opened 2023-11-20
Show Annif version in WebUI

juhoinkinen:
It would be nice to know which version of Annif is running for the Web UI.

This PR adds a version number string, e.g. `Annif v1.1.0` (or `Annif v1.1.0.dev0` for the development version) outside the text box in the lower right corner, see screenshot below.

An alternative position for the version number could be just below "Web UI" title in the header, but the position of this PR is the same as will be in Finto AI.

image

### DIFF
--- a/annif/static/css/style.css
+++ b/annif/static/css/style.css
@@ -31,6 +31,11 @@ h1 { font-size: 2rem; }
   border-radius: 0px;
   padding: 2px 7px;
 }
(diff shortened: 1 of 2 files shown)


osma:
Looks very good, but is it clear enough with only the version number? Would it be better to say "Annif v1.1.0" or something along those lines...

Issue #747, opened 2023-12-01
Tests error due to pinned schemathesis version 3.19.* / Docker rebuild fails


juhoinkinen:
Running unit tests fail with the following error:
```
...
  File "/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/schemathesis/lazy.py", line 9, in <module>
    from hypothesis.internal.escalation import format_exception, get_interesting_origin, get_trimmed_traceback
ImportError: cannot import name 'get_interesting_origin' from 'hypothesis.internal.escalation' (/home/local/jmminkin/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.8/lib/python3.8/site-packages/hypothesis/internal/escalation.py)
```

This originates from a change in Hypothesis, which makes it incompatible with Schemathesis: https://github.com/HypothesisWorks/hypothesis/issues/3804. On Schemathesis side this is already [fixed](https://github.com/schemathesis/schemathesis/pull/1894) and the [release v3.21.1](https://schemathesis.readthedocs.io/en/stable/changelog.html#v3-21-2) includes it, but because in Annif the Schemathesis version [is pinned to 3.19.*](https://github.com/NatLibFi/Annif/blob/a501e53b9103d339de30c173f43ccf0e60f7b963/pyproject.toml#L72), the fix does not propagate in use.

For development and future releases the Schemathesis version pinning can be just updated or removed, but this is not possible in rebuilds of the Annif 1.0 Docker image ([see failed, manually triggered rebuild attempt](https://github.com/NatLibFi/Annif/actions/runs/7047875817/job/19182742757)), because for that no code changes are made.

Depending on the need to keep the Annif 1.0 Docker image up-to-date and on the schedule of 1.1 release, we could consider releasing Annif 1.0.1, where Schemethesis version pinning is removed or updated. 

But for avoiding these kind of problems in the future the version pinning policy should be thought from this point of view: now the dev dependencies are fully unpinned except black and schemathesis.

juhoinkinen:
> Depending on the need to keep the Annif 1.0 Docker image up-to-date and on the schedule of 1.1 release, we could consider releasing Annif 1.0.1, where Schemethesis version pinning is removed or updated.
> 
> But for avoiding these kind of problems in the future the version pinning policy should be thought from this point of view: now the dev dependencies are fully unpinned except black and schemathesis.

[...]

PR #751, opened 2023-12-11
Create `SECURITY.md`

juhoinkinen:
It is a good practice to inform how to report vulnerabilities, which is what `SECURITY.md` file is for.

~I have worded this in `SECURITY.md` file:~
~> However, the [dependencies of a given Annif release](https://github.com/NatLibFi/Annif/blob/main/pyproject.toml) are pinned only on minor version level~

~But this is actually not true;~ [Edited this part to include a true statement.]
There are the following patch level version pinnings:
- connexion `2.14.2`
- fasttext `0.9.2`
- lmdb `1.4.1`
- yake `0.4.5`

Right now I don't remember why they are pinned to patch level. Need to check that. It would be good that all (security) patches of dependencies could be applied without the need to update Annif version. Does some of these dependencies have such a backward compatibility policy that conflicts with Annif's policy?

~Also reword the bottom list.~ [Done.]

Also at the moment Annif repo has [Security advisories](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability) enabled (I enabled it when looking at this security reporting a while ago), which allows anyone to give a report via navigating to the Security tab in this repo. Having two reporting ways (finto-posti email and this reporting) can be confusing, so maybe the Security advisories functionality should be disabled.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,14 +2,18 @@
 
 ## Supported Versions
 
-We aim to update all Annif dependencies to their latest versions
-on each Annif minor version release, but this can be restricted by the
-[backward compatibility policy](https://github.com/NatLibFi/Annif/wiki/Backward-compatibility-between-Annif-releases).
+The [most recent Annif (major/minor) release](https://github.com/NatLibFi/Annif/releases)
+is considered supported,
+in the sense that if a serious bug or vulnerability is encountered in it,
+a patch release is made to fix the issue.
 
-However, the [dependencies of a given Annif release](https://github.com/NatLibFi/Annif/blob/main/pyproject.toml)
+Generally, we aim to update all dependencies to their latest versions
+on each Annif major/minor release, but this can be restricted by the
+[backward compatibility policy](https://github.com/NatLibFi/Annif/wiki/Backward-compatibility-between-Annif-releases).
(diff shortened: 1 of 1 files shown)


juhoinkinen:
I don't see a reason to have patch-level pinning of dependencies. I propose to pin only the minor version levels:

- connexion `2.14.2` -> `2.14.*` 

- fasttext `0.9.2` -> `0.9.*` 

- lmdb `1.4.1` -> `1.4.*`

- yake `0.4.5` ->  `0.4.*`


This should be done in a separate PR/issue, possibly with the issue #747.

juhoinkinen:
Gunicorn is about to set up a security policy too, either via `SECURITY.md` or the Security advisories / Private Vulnerability Reporting: https://github.com/benoitc/gunicorn/issues/3106

We could wait and see what their solution will be.

juhoinkinen:
I asked for a review from ChatGPT and addressed the good comments.


> But this is actually not true; there are the following patch level version pinnings:
>
>   -  connexion 2.14.2
>    - fasttext 0.9.2
>    - lmdb 1.4.1
>    - yake 0.4.5

For this, I reworded  the relevant part: "...note that **most** of the dependencies of a given Annif release are pinned only on minor version level...".

Issue #760, opened 2024-01-30
Hugging Face integration


juhoinkinen:
The :hugs: Hugging Face Hub intends to facilitate the hosting and sharing of AI models and datasets (as well as demo applications), and now also NatLibFi has an [organization account in the Hugging Face Hub](https://huggingface.co/NatLibFi).

The data (models and datasets) in the HF Hub live in git repositories, and git can be used to handle the data (to commit, push, pull...) . However, also direct integration of applications with HF Hub is supported using the [`huggingface_hub` Python library](https://huggingface.co/docs/huggingface_hub), which is usable also as a [CLI tool](https://huggingface.co/docs/huggingface_hub/guides/cli).

Annif could have the functionality to push (and pull) projects or project sets to (and from) the HF Hub. It should to be able to operate on project sets because ensemble projects require the availability of also its base projects and also because of convenience.

There could be the following CLI command to push a set of projects to HF Hub:
```
annif upload-projects <glob-pattern> <username/reponame> [--options]
```
For example
```
annif upload-projects yso-*fi NatLibFi/FintoAI-data-YSO
```
would upload the specified projects to [NatLibFi/FintoAI-data-YSO](https://huggingface.co/NatLibFi/FintoAI-data-YSO) repository.

The files and dirs needed to be uploaded are  
- `data/projects/project-id` the project directories
- `data/vocabs/vocab-id` vocabularies of the projects
- `projects.{cfg,toml,d}` configurations of the projects

### 1. Single file
Bundle all files into one zip named: `yso-fi.zip` (possibly include only the configs of the selected projects). Upload to the root of the repo.

The filename could be derived by the glob pattern of the projects or it could be a required argument for the upload command (as 2nd argument, to be added to the above example).

This option would be easiest for downloads: just `wget` one file and unzip.

### 2. One file for projects and vocab, and one for projects configs
Bundle projects and vocabulary directories into one zip and leave projects config file uncompressed.

### 3. One file for projects, one for vocab, and one for projects configs
Bundle the selected projects into one zip (`yso-fi.zip`) and vocabularies into another (`yso.zip`) and leave projects config file uncompressed. Upload the projects zip to `data/projects` directory and the vocab zip to `data/vocabs`. 

### 4. Separate files for each project, vocab, and projects configs
Compress each project directory into its own zip (`<project-id>.zip`).

For this option for downloads one should use e.g. `wget --accept yso*-fi.zip` for the projects. 

---
Some details and ideas:
[...]

PR #762, opened 2024-02-01
Implement `annif upload` and `annif download` commands for Hugging Face Hub integration

juhoinkinen:
The organization of the uploaded files follows the 4. option of issue #760:
> 4. Separate files for each project, vocab, and projects configs
>
> Compress each project directory into its own zip file (in `projects/<project-id>.zip` in the repo). 

Each vocabularies zip is placed in `vocabs/<vocab-id>.zip`, and each project configuration is placed in `<project-id>.cfg` in the repo root.

This option is good for caching by preventing unnecessary uploads/downloads of the projects bundle when only one is changed, and for visibility of the repo contents.

The unzip after download places the data directories directly to `data/projects/` and `data/vocabs/`, and the project configurations as separate `<project-id>.cfg` files in `projects.d/` directory, so after download the projects are directly usable (if `projects.{cfg,toml}` does not exists or by using the `--projects` option).

# Upload
Push a set of selected projects and vocabularies to a Hugging Face Hub repository as zip files and the configurations of the projects as cfg/ini files. For example, the following command uploads all projects with ids matching `*-fi` to [juhoinkinen/Annif-models-upload-testing](https://huggingface.co/juhoinkinen/Annif-models-upload-testing/tree/main):
```
annif upload-projects "*-fi" juhoinkinen/Annif-models-upload-testing --commit-message "[OptMsg] Upload all Finnish projects"
```

# Download
Downloads the project and vocabulary archives and the configuration files of the projects that match the given
and unzip the archives to `data/` directory and places the configuration files to `projects.d/` directory:
```
annif download-projects "yake-fi" juhoinkinen/Annif-models-upload-testing
```
~Note that currently the download will **overwrite the project and vocab dirs** if they already exist.~ Edit: By default overwrite does not happen, it can be performed by adding the `--force` option to the command.

A git revision (commit hash, branch etc.) can be specified with `--revision` option.

The downloaded files remain in the huggingface_hub client cache dir.

---

For all uploads and downloads in the case of private repos the user needs to have logged in with `huggingface-cli login`, or the HF Hub token can be given also with the `--token` option of this Annif command.


annif/cli.py
```diff
@@ -583,6 +584,112 @@ def run_hyperopt(project_id, paths, docs_limit, trials, jobs, metric, results_fi
...
+    `project_ids_pattern` from the specified Hugging Face Hub repository and
+    unzips the archives to `data/` directory and places the configuration files
+    to `projects.d/` directory. An authentication token and revision can
+    be given with options.
+    """
+
+    project_ids = cli_util.get_selected_project_ids_from_hf_hub(
+        project_ids_pattern, repo_id, token, revision
+    )
+    click.echo(f"Downding project(s): {', '.join(project_ids)}")
```
juhoinkinen, line 670: TODO: Fix typo `Downding`

juhoinkinen:
TODO:

- [ ] Try to find a way to avoid the need to use quotes around projects-id pattern (without them shell expansion kicks in)
- [ ] Add `--force` option to allow overwrite of existing files when downloading; prevent overwrite otherwise
- [ ] Add tests: could use mocking of `huggingface_hub` methods to assert their calls, e.g. `@mock.patch("huggingface_hub.commands.upload.HfApi.upload_file")`
- [ ] Speed up CLI startup by moving `huggingface_hub` imports to functions

 @osma could you take a look at the working principles of the commands before I spend more time with this? 

The downloaded files now remain in cache for huggingface client, e.g. (`/home/local/jmminkin/.cache/huggingface/hub/models--juhoinkinen--Annif-models-upload-testing`), should the cache be cleared when successfully unzipped the datadirs...?

osma:
I took a look at this (sorry for the delay!) and I think this is a really good feature. Some thoughts:

1. Agree that these commands should be marked as experimental for now. We may want to change the command options, functionality etc. Maybe in the 1.2 release the experimental status could be dropped.
2. The commands are uploading/downloading not just the project(s), but also the vocabulary(/ies). Would it make sense to shorten the commands to just `upload` and `download`?
3. Regarding overwriting the vocabulary on download, I think this is a problematic feature and needs some thought. For example if I already have trained models with vocabularies in my data directory, and I download another model from HF Hub, it might come with another version of the same vocabulary and thus rewrite my existing vocabulary and thus break existing projects. So I agree that there should be some kind of check for this and perhaps an `--overwrite` or `--force` option before blindly overwriting existing vocabularies.

I tested this on my local install by running the command `annif download-projects "yake-fi" juhoinkinen/Annif-models-upload-testing`; I wanted to see what happens when I already have a local `yso` vocabulary. The overwrite failed with an exception:

```
$ annif download-projects "yake-fi" juhoinkinen/Annif-models-upload-testing
Downding project(s): yake-fi
yake-fi.zip: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2.63M/2.63M [00:00<00:00, 7.24MB/s]
yake-fi.cfg: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 119/119 [00:00<00:00, 470kB/s]
yso.zip: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 54.6M/54.6M [00:04<00:00, 11.0MB/s]
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/bin/annif", line 6, in <module>
    sys.exit(cli())
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/flask/cli.py", line 357, in decorator
    return __ctx.invoke(f, *args, **kwargs)
  File "/home/xxx/.cache/pypoetry/virtualenvs/annif-3KgT0m3n-py3.10/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/xxx/git/Annif/annif/cli.py", line 690, in run_download_projects
    cli_util.unzip(vocab_zip_local_cache_path)
  File "/home/xxx/git/Annif/annif/cli_util.py", line 316, in unzip
    zfile.extractall()  # TODO Disallow overwrite
  File "/usr/lib/python3.10/zipfile.py", line 1647, in extractall
    self._extract_member(zipinfo, path, pwd)
  File "/usr/lib/python3.10/zipfile.py", line 1701, in _extract_member
    open(targetpath, "wb") as target:
PermissionError: [Errno 13] Permission denied: '/home/xxx/git/Annif/data/vocabs/yso/subjects.dump.gz'
```

> The downloaded files now remain in cache for huggingface client, e.g. (/home/local/jmminkin/.cache/huggingface/hub/models--juhoinkinen--Annif-models-upload-testing), should the cache be cleared when successfully unzipped the datadirs and moved the configs?

Isn't the caching standard behaviour of HF Hub operations? I think just keeping the cache would be the least surprising thing to do here. Maybe there could be a `--no-cache` or `--no-keep` option to control caching? What do other tools with HF Hub integration (e.g. Ludwig) do?

osma:
About overwriting vocabularies - I think it would be great if Annif would notice, that the downloaded vocabulary differs from what exists locally. It should be no problem to download two different models that use the same vocabulary:

    annif download-projects "yake-fi" juhoinkinen/Annif-models-upload-testing
    annif download-projects "yso-fasttext-fi" juhoinkinen/Annif-models-upload-testing
    annif download-projects "yso-tfidf-fi" juhoinkinen/Annif-models-upload-testing

But when you have a different version of the vocabulary already present, Annif could show a message such as

    Your local version of "yso" vocabulary differs from the version you are downloading. If you are sure you want to overwrite it, run the command again with the `--overwrite` option. This may break projects that use your current vocabulary.

Even better would be to list the projects that may break.

But how to detect that the local vocabulary is different than the downloaded one? Calculate checksums on the files and compare them?

juhoinkinen:
> But how to detect that the local vocabulary is different than the downloaded one? Calculate checksums on the files and compare them?

Actually the ZipFile objects [allow getting ZipInfo](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.infolist) of each member of the archive, which in turn contain a [CRC-32 checksum](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.CRC) of the uncomprossed file. And [ZipInfo.from_file()](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.from_file) class method allows to get the CRC from the existing, uncompressed files, so the checksum comparisons should be quite easy to implement.

osma:
If it's the above bug zipfile, it's [claimed](https://bugs.python.org/issue26185#msg374697) to be

> Fixed in 3.9.0

Annif just dropped 3.8 support so maybe we're good? Just need to rebase this branch.

juhoinkinen:
The Python versions 3.9.18, 3.10.13 and 3.11.8 all behaved the same.

 This seems strange: the code
```python
zinfo = zipfile.ZipInfo.from_file("data/projects/yake-fi/yake-index")
print(dir(zinfo))
print(zinfo.file_size)
print(zinfo.CRC)
```
outputs
```
['CRC', 'FileHeader', '__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '_compresslevel', '_decodeExtra', '_encodeFilenameFlags', '_end_offset', '_raw_time', 'comment', 'compress_size', 'compress_type', 'create_system', 'create_version', 'date_time', 'external_attr', 'extra', 'extract_version', 'file_size', 'filename', 'flag_bits', 'from_file', 'header_offset', 'internal_attr', 'is_dir', 'orig_filename', 'reserved', 'volume']
2627360
  File "/home/local/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/bin/annif", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/flask/cli.py", line 357, in decorator
    return __ctx.invoke(f, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myusername/.cache/pypoetry/virtualenvs/annif-ul-EXdhi-py3.11/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/myusername/git/Annif/annif/cli.py", line 686, in run_download
    cli_util.unzip(project_zip_local_cache_path, force)
  File "/home/local/myusername/git/Annif/annif/cli_util.py", line 325, in unzip
    _is_existing(member)
  File "/home/local/myusername/git/Annif/annif/cli_util.py", line 351, in _is_existing
    print(zinfo.CRC)
          ^^^^^^^^^
AttributeError: 'ZipInfo' object has no attribute 'CRC'
```
The CRC attribute is reserved using [`ZipInfo.__slots__()`](https://github.com/python/cpython/blob/8813e5a2dea3fbea6be7cb4e3db8a64c8665bc90/Lib/zipfile.py#L366), but then it is not initialized when running `ZipInfo.from_file()`? 

Also trying to access `zinfo.FileHeader()` fails similarly, with Traceback ending
```
...
File "/usr/lib/python3.11/zipfile.py", line 453, in FileHeader
    CRC = self.CRC
          ^^^^^^^^
AttributeError: 'ZipInfo' object has no attribute 'CRC'
```

Anyway, the CRC-32 checksums for the existing, uncompressed files could be calculated by `[binascii.crc32()](https://docs.python.org/3.10/library/binascii.html#binascii.crc32)`.

juhoinkinen:
I added `--force` option to overwrite existing local files (project and vocabulary data files and configs), and modified the unzip/config-copy logic so that when the contents are identical (downloaded vs local), there the unzip/config-copy is skipped.

I did not see a reason to restrict the content comparison to only vocabularies(?) (and it was easier to implement both projects and vocabs :slightly_smiling_face:).

Demo with
```bash
#!/bin/bash

rm -r data/projects/{fasttext-fi,tfidf-fi,yake-fi}

echo "# Initial download"
annif download "tfidf-fi" juhoinkinen/Annif-models-upload-testing
echo "# Second download with identical content, no-op"
annif download "tfidf-fi" juhoinkinen/Annif-models-upload-testing

echo $RANDOM > data/projects/tfidf-fi/file.txt
echo $RANDOM >> projects.d/tfidf-fi.cfg
echo "# Third download over changed content, skips file overwrites with complains"
annif download "tfidf-fi" juhoinkinen/Annif-models-upload-testing

echo "# Forced third download over changed content"
annif download "tfidf-fi" juhoinkinen/Annif-models-upload-testing --force
tree data
```

The output:
```
# Initial download
Downloading project(s): tfidf-fi
# Second download with identical content, no-op
Downloading project(s): tfidf-fi
# Third download over changed content, skips file overwrites with complains
Downloading project(s): tfidf-fi
Not overwriting data/projects/tfidf-fi/file.txt (use --force to override)
Not overwriting projects.d/tfidf-fi.cfg (use --force to override)
# Forced third download over changed content
Downloading project(s): tfidf-fi
data
├── projects
│   └── tfidf-fi
│       └── file.txt
└── vocabs
    └── yso
        ├── subjects.csv
        ├── subjects.dump.gz
        └── subjects.ttl

4 directories, 4 files
```

juhoinkinen:
Force-pushed to squash multiple refactoring and renaming commits.

juhoinkinen:
I noticed that after 

    annif download yso-bonsai-en juhoinkinen/Annif-models-upload-testing

the modification time of omikuji projects as detected by Annif is wrong; it is shown to be the extraction time:
```
annif show-project yso-bonsai-en 
...
Modification time: 2024-03-13 11:36:22
```
This is because of the timetamp of the `data/projects/yso-bonsai-en/omikuji-model` which is set to the time when the files it contains get extracted.

This PR implements [restoration of the timestamps of the extracted files and directories](https://github.com/NatLibFi/Annif/blob/2fe5b739085e159c2fc3320db85f1ac6f9832605/annif/cli_util.py#L412), but this happens zipmember-by-zipmember, and typically the project subdirectories are extracted first (`yso-bonsai-en/omikuji-model/`). 

I think the [modification time detection](https://github.com/NatLibFi/Annif/blob/eba2b0c3f829e966e454299c800f0e49e6f71086/annif/backend/backend.py#L70-L78) should be changed to disregard subdirectories, and not to try handle this issue in this PR.

juhoinkinen:
I created a Wiki page about the HFH integration: https://github.com/NatLibFi/Annif/wiki/Hugging-Face-Hub-integration.

juhoinkinen:
I [commented](https://github.com/huggingface/huggingface_hub/issues/960#issuecomment-2013533386) on huggingface-hub repository about git tags, and they promptly created an [issue](https://github.com/huggingface/huggingface_hub/issues/2141) for creating, listing and deleting tags with the huggingface-cli (also wrote a [question](https://discuss.huggingface.co/t/tagging-with-git-tags-models-for-versioning/78489) on HF forum).

If/when it gets implemented, I think it would be most convenient for the case of Annif models if the tagging would be performed after `annif upload <projects-pattern> <repo-id>` using huggingface-cli, e.g. like `huggingface-cli tag <repo-id> <tag-name>`.

juhoinkinen:
> Actually multiple files could be included in one commit [using `CommitOperationAdd()` and `create_commit()`](https://huggingface.co/docs/huggingface_hub/guides/upload#createcommit), but it seems that for this all the included files should exist until running `create_commit()`, and the zips can be quite large.

The [`preupload_lfs_files()`](https://huggingface.co/docs/huggingface_hub/v0.22.0.rc0/en/guides/upload#preupload-lfs-files-before-commit) function helped to put all files in one commit but to upload them separately.

When testing with `yso-*fi` projects of FintoAI (~10 GBs uncompressed, ~4 GBs compressed) `/usr/bin/time -v` reported `Maximum resident set size (kbytes): 49616` (and I did not notice much increase in disk space usage); upload took about 20 min at the office.

juhoinkinen:
While uploading projects to [FintoAI-data-YSO](https://huggingface.co/NatLibFi/FintoAI-data-YSO/tree/main) with `annif upload "*" NatLibFi/FintoAI-data-YSO `, this error occurred on two tries:
```
...
  File "/home/local/jmminkin/git/Annif/annif/cli_util.py", line 259, in prepare_datadir_commit
    fobj = _archive_dir(data_dir)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/local/jmminkin/git/Annif/annif/cli_util.py", line 285, in _archive_dir
    with zipfile.ZipFile(fp, mode="w") as zfile:
  File "/usr/lib/python3.11/zipfile.py", line 1355, in __exit__
    self.close()
  File "/usr/lib/python3.11/zipfile.py", line 1911, in close
    self.fp.seek(self.start_dir)
OSError: [Errno 28] No space left on device
```
However I had about 60 GBs of free disk space. And [preuploading should free memory of the uploaded objects](https://huggingface.co/docs/huggingface_hub/v0.22.0.rc0/en/guides/upload#preupload-lfs-files-before-commit), so it should not be memory issue either... Anyway, this could be circumvented by uploading only projects of one language by one command.

osma (approved):
Looks very good.

The only thing that caught my eye is that this PR adds a lot of new code especially to cli_util.py, but also cli.py and test_cli.py. Would it make sense to put the HF-specific code in separate files instead, at least in the case of cli_util.py?

osma:
:tada:

PR #778, opened 2024-03-13
Fix project data files detection

juhoinkinen:
I noticed that after 

    annif download yso-bonsai-en juhoinkinen/Annif-models-upload-testing

the modification time of omikuji projects as detected by Annif is wrong; it is shown to be the extraction time:
```
annif show-project yso-bonsai-en 
...
Modification time: 2024-03-13 11:36:22
```
This is because of the timestamp of the `data/projects/yso-bonsai-en/omikuji-model/` _directory_, which is set to the time when the files it contains get extracted.

_Originally posted by @juhoinkinen in https://github.com/NatLibFi/Annif/issues/762#issuecomment-1994122090_

---

This PR fixes the above problem by making all subdirectories in project data directory to be disregarded as model data files (the directories themselves do not affect the model results).

Also make the files in the subdirectories be regarded as they should be.

### DIFF
--- a/annif/backend/backend.py
+++ b/annif/backend/backend.py
@@ -53,14 +53,15 @@ class AnnifBackend(metaclass=abc.ABCMeta):
 
     @property
     def _model_file_paths(self) -> list:
-        all_paths = glob(os.path.join(self.datadir, "*"))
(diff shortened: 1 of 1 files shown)


osma (approved):
LGTM. I could think of other ways of getting to the same result, but this is certainly good enough.

PR #780, opened 2024-03-21
Python 3.12 support

### LINKED ISSUE #779: Python 3.12 support
Python 3.12 was [released last October](https://www.python.org/downloads/release/python-3120/), so we should try to support it in Annif.

Currently support is not possible because of the following (to be continued...):
- ~gensim v4.3.2 does not provide a wheel for Python 3.12 in [PyPI](https://pypi.org/project/gensim/#files), and building the wheel locally fails:~

    ```
    ...
    gensim/models/word2vec_inner.c:25:10: fatal error: Python.h: No such file or directory
         25 | #include "Python.h"
            |          ^~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1
    ```
   - ~issues https://github.com/piskvorky/gensim/issues/3511, https://github.com/piskvorky/gensim/issues/3510~
- tensorflow-io-gcs-filesystem requires Python >=3.7, <3.12 
  - issues https://github.com/tensorflow/io/issues/1761, https://github.com/tensorflow/io/issues/1966)
- stwfsapy requires Python <3.12,>=3.9 
    - issue https://github.com/zbw/stwfsapy/issues/61

juhoinkinen:
Annif on Python 3.12 already works when the optional dependencies stwfsapy and tensorflow-cpu are not installed or even declared as dependecies.

~Closes #779.~ 
At the moment does not fully close #779 because stwfsapy and NN ensemble will not work on Python 3.12.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -61,7 +61,7 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     name: test on Python ${{ matrix.python-version }}
     steps:
     - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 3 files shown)


juhoinkinen:
While Poetry cannot install TensorFlow 2.16 because of its wrong package metadata, pip can, see branch [python3.12-support-tf-via-pip](https://github.com/NatLibFi/Annif/tree/python3.12-support-tf-via-pip). TF 2.16 also upgrades to [Keras 3](https://keras.io/guides/migrating_to_keras_3/), which required a few changes. 

So it could be an option to install TF with pip instead of Poetry; this would allow upgrading to TF 2.16  and thus use of NN ensemble on Python 3.12.

PR #771, opened 2024-02-23
Update dependencies for v1.1 release

juhoinkinen:
Updates the outdated dependencies.

Based on #770 as most upgraded dependencies do not work on Python 3.8.

Exceptions to upgrading to newest releases (for now, to be updated. _Edit: updated_):
- Flask and Connexion
  -  after #702 there will not be a direct Flask dependency
- TensorFlow: there is [TensorFlow 2.16.1](https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1), but upgrading to it fails when trying to install it:
    ```
    Because no versions of tensorflow-cpu match >=2.16.dev0,<2.16.0rc0 || >2.16.0rc0,<2.16.1 || >2.16.1,<2.17.dev0
    and tensorflow-cpu (2.16.0rc0) depends on tensorflow-cpu-aws (2.16.0-rc0), tensorflow-cpu (==2.16.0.* ||    >2.16.1,<2.17.dev0) requires tensorflow-cpu-aws (2.16.0-rc0).
   And because no versions of tensorflow-cpu-aws match 2.16.0-rc0
    ``` 
  - issue https://github.com/tensorflow/tensorflow/issues/63585
- ~Scikit-learn: there is [Scikit-learn 1.4.1](https://github.com/scikit-learn/scikit-learn/releases/tag/1.4.1), but stwfsapy requires scikit-learn <1.3~
  -  ~issue https://github.com/zbw/stwfsapy/issues/58~
- ~rdflib: also this awaits for stwfsapy to upgrade Scikit-learn~
- ~Black: upgrading to [24 series](https://black.readthedocs.io/en/stable/change_log.html#id5) leads to some minor new style changes in the code base (mostly "Enforce newline after module docstrings")~

PR #786, opened 2024-04-12
Update dependencies for v1.1 release - round 2

juhoinkinen:
Continuation to #771.

Also makes requests a regular dependency instead of a test dependency, because the [HTTP backend](https://github.com/NatLibFi/Annif/wiki/Backend%3A-HTTP) imports requests, so requests needs to be always installed.


juhoinkinen:
Also [SciPy v1.13.0 has been relesased](https://github.com/scipy/scipy/releases/tag/v1.13.0), and we could update to it from v1.12, but there is a breaking change

> - scipy.linalg.tri{,u,l} have been removed in favour of numpy.tri{,u,l}

which affects Gensim. Fortunately they have already [fixed this](https://github.com/piskvorky/gensim/pull/3524) in their development branch, and hopefully there will soon be a release to include the fix.

osma (approved):
lgtm

PR #789, opened 2024-05-26
adding stable gensim for supporting py3.12

tavallaie:
adding last stable commit of gensim for support python 3.12 solving: #779
I wrote similar problem here.
https://blog.techbend.io/overcoming-the-hazm-compatibility-challenge-with-python-312

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ click = "8.1.*"
 click-log = "0.4.*"
 joblib = "1.4.*"
 nltk = "3.8.*"
-gensim = "4.3.*"
+gensim={git = "https://github.com/piskvorky/gensim.git", rev ="da2f388"}
 scikit-learn = "1.4.*"
 scipy = "1.12.*"
 rdflib = "7.0.*"


osma:
Thank you for the PR @tavallaie !

Can you elaborate on what benefit this PR brings? In my understanding, Annif already mostly works in Python 3.12, though some dependencies will need to be compiled instead of installed from wheels. We use Python 3.12 in the new Annif-tutorial image and basically everything is working, even TensorFlow (though that needed some manual intervention).

According to the CI/CD pipeline, the newer Gensim cannot be installed on Python 3.9. So we would have to drop support for 3.9.

How is this "stable gensim"? It seems to be referencing a specific commit, not a proper release.

tavallaie:
The gensim version 4.3.2 uses the deprecated triu module from scipy, making it incompatible with Python 3.12. In the version I added, gensim uses triu from numpy, but this update has not yet been added to PyPI.

osma:
@tavallaie Python 3.12 with gensim 4.3.2 seems to work in practice at least for the purposes of Annif. See e.g. [this](https://github.com/NatLibFi/Annif/actions/runs/8833056428/job/24251558625) CI/CD run for the Annif 1.1.0 release, which succesfully installed dependencies (including scipy and gensim) and ran the test suite.

I'm sure you are correct in that there are compatibility issues between the software packages and versions that you mentioned, but I see little harm in practice from the perspective of Annif users. I think it would be best to wait for a formal gensim release on PyPI and then upgrade to it. We probably need to drop Python 3.9 support in the same go if the new gensim release doesn't support it anymore.

osma:
I'm curious if you had an actual problem running Annif on Python 3.12? What made you create this PR?

tavallaie:
I encountered an issue with Gensim in another project and found a temporary fix. Noticing similar issues in your project, I applied the same fix here. I do this to get involved in other projects and learn new things. Thanks!

PR #792, opened 2024-06-12
Fix limit parameter not passed to requests by HTTP backend

juhoinkinen:
I noticed that the limit parameter in the project configuration had no effect to the number of the suggestions in the case of the HTTP backend. This is fixed by adding that parameter to the POST request, if it is used.

Also adds a test to verify that the the POST request by the HTTP backend actually contains the intended arguments.


juhoinkinen:
@CodiumAI-Agent /review

osma (approved):
lgtm

PR #793, opened 2024-06-17
Automatically add metadata to Hugging Face Hub repos when uploading projects

### LINKED ISSUE #790: Automatically add metadata to Hugging Face Hub repos when uploading projects
The :hugs: Hugging Face Hub has good support for metadata of the model, dataset and spaces repositories with the yaml section of the [Model Cards](https://huggingface.co/docs/hub/model-cards) (i.e. the `README.md` files of the repos).

While updating [Annif-tutorial](https://github.com/NatLibFi/Annif-tutorial), specifically writing the [Model card section of the Hugging Face Hub exercise](https://github.com/NatLibFi/Annif-tutorial/blob/master/exercises/OPT_huggingfacehub.md#create-a-model-card-in-your-repository), I noted that it is possible to [create the Model card programmatically](https://huggingface.co/docs/huggingface_hub/guides/model-cards#create-model-cards) with the HFH client and also to directly [update the metadata of the Model card](https://huggingface.co/docs/huggingface_hub/guides/model-cards#update-metadata).

Adding the metadata to the Model card is easy via the HFH website UI, but I still assume most users won't remember or bother to do it, so it would be nice that the some metadata was added automatically when running `annif upload`. It could be the default behaviour, which could be opted-out with an option of `annif upload`.

For example, to set the model task to "text-classification" and a custom tag "annif" this code could be used:
```python
from huggingface_hub import metadata_update

metadata_update(
    "juhoinkinen/Annif-models-upload-testing",
    {
        "pipeline_tag": "text-classification",
        "tags": ["annif"]
    }
)
```
I think these are the most import metadata, and they could be hard coded, but also the language(s) could be considered, as they are available in the vocabulary. See e.g. this [NatLibFi/FintoAI-data-YSO](https://huggingface.co/NatLibFi/FintoAI-data-YSO) repo for also some other (manually added) metadata.

There is just one small issue with the `metadata_update()` use: if the `README.md` does not exist in a repo, it is created with with the [ModelCard template](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/templates/modelcard_template.md) contents, which I feel is not well applicable for Annif projects, because it expects very much information. So either an empty `README.md` should be uploaded before calling `metadata_update()`, or maybe that function the HFH client could be modified to accept a parameter to control whether the Model card template is used or not.

This would be continuation to PR #762.

juhoinkinen:
With this PR, when running `annif upload`:
-  if `README.md` (Model Card) does not exist in the destination repository, then `README.md` is created with default contents, project list (containing projects that have configuration (`*.cfg`) files in the repo) and some metadata of the uploaded projects,
- if `README.md` exists, its project list and metadata are updated as necessary. 

Closes #790.

The metadata includes these:
```
language:
- <language-code tags automatically obtained from the uploaded projects>
tags:
- annif   # custom tag
pipeline_tag: text-classification  # HFH tag
```

The text content is like this (from here https://huggingface.co/juhoinkinen/Annif-models-upload-testing):

> # Annif-models-upload-testing
> ## Usage
> 
> Use the annif download command to download selected projects with Annif; for example, to download all projects in this repository run
> 
>     annif download "*" juhoinkinen/Annif-models-upload-testing
> 
> ## Projects
> ```
> Project ID          Project Name            Vocabulary ID   Language   
> --------------------------------------------------------------------
> dummy-en            Dummy English           dummy           en         
> dummy-fi            Dummy Finnish           dummy           fi         
> dummy-sv            Dummy Swedish           dummy           sv
> ```
> This text should retain on updates of project list.


juhoinkinen:
About @osma's  suggestions in https://github.com/NatLibFi/Annif/issues/790#issuecomment-2137376118:
> For example it could include the Annif version used for training, the backend, vocabulary name and size, possibly some of the hyperparameters / configuration settings as well.


- Annif version:
  - The Annif version used for training is not stored anywhere at the moment; the version performing the upload is not necessarily the same. This kind of metadata should be first stored somewhere, for which there is the issue https://github.com/NatLibFi/Annif/issues/329
- Backend, vocabulary name and other project configuration:
  - These are available in the `<project-id>.cfg` files, accessible from the Files and versions tab, e.g. https://huggingface.co/NatLibFi/FintoAI-data-YSO/blob/main/yso-en.cfg, so I think they are not worth putting to the Model Card.

juhoinkinen:
@CodiumAI-Agent /review

tests/test_hfh_util.py
```diff
@@ -101,3 +101,67 @@ def test_copy_project_config_overwrite(copy, exists):
...
     )
+
+
+@mock.patch("annif.hfh_util._list_files_in_hf_hub", return_value=["README.md"])
+@mock.patch(
+    "huggingface_hub.ModelCard",
+)
+def test_upsert_modelcard_existing_card(ModelCard, _list_files_in_hf_hub, project):
+    repo_id = "annif-user/Annif-HFH-repo"
+    project.vocab_lang = "fi"
```
juhoinkinen, line 112: The [project fixture](https://github.com/NatLibFi/Annif/blob/3b5f7a1ded9e841bc38a5c5e1a2c55a75cd0521c/tests/conftest.py#L137) does not provide `vocab_lang` so for these tests it is just set here, not super clean maybe?

juhoinkinen:
> **Possible Bug:**
> Ensure that the upsert_modelcard function handles cases where project language data might be missing or malformed. > The current implementation assumes that proj.vocab_lang is always available and valid.

Good point by the AI, but I think the project language is always set if this point is reached...?

juhoinkinen:
I added an automatically updating Projects section to the modelcard, like this: https://huggingface.co/juhoinkinen/Annif-models-upload-testing#projects

annif/cli.py
```diff
@@ -616,8 +616,15 @@ def run_hyperopt(project_id, paths, docs_limit, trials, jobs, metric, results_fi
     "--commit-message",
     help="""The summary / title / first line of the generated commit.""",
 )
+@click.option(
+    "--modelcard/--no-modelcard",
+    default=True,
+    help="Update or create a ModelCard with upload.",
```
osma, line 622: I think this should be "Model Card" not "ModelCard"

annif/cli.py
```diff
@@ -690,7 +700,9 @@ def run_download(project_ids_pattern, repo_id, token, revision, force):
     `project_ids_pattern` from the specified Hugging Face Hub repository and
     unzips the archives to `data/` directory and places the configuration files
     to `projects.d/` directory. An authentication token and revision can
-    be given with options.
+    be given with options. If the README.md does not exist in the repository it is
+    created with default contents and metadata of the uploaded projects, if it exisits,
```
osma, line 704: typo: exists

osma (approved):
I can't think of anything to improve; I think we need to find out through actual use whether there is something that could be fixed or made better.

I did find two typos though :)

PR #800, opened 2024-09-17
Sort language detection results by descending score

osma:
This is an improvement to the language detection REST API method (#631) which improves the implementation that was made in #659.

Results are now sorted by descending score, so the most likely language is returned first.


juhoinkinen (approved):
:+1:

PR #796, opened 2024-07-17
Update dependencies for v1.2 release

juhoinkinen:
Updates the outdated dependencies.

Depedencies, which are not upgraded to the newest releases:
- numpy
  - Because tensorflow-cpu (2.17.0) depends on numpy (>=1.23.5,<2.0.0)
- scipy
  -  v1.14 requires Python >=3.10

Also resolves two DeprecationWarnings by Schemathesis.

PR #803, opened 2024-09-19
Automate NLTK datapackage `punkt_tab` download

juhoinkinen:
Based on the update-dependencies-v1.2 branch (PR #796) because it upgrades NLTK to a version that needs the `punkt_tab` package (not `punkt` anymore): https://github.com/nltk/nltk/pull/3283) 

The download is performed automatically when instantiating an Analyzer object, if the `punkt_tab` data is not found. The `README.md` has been adjusted for this (i.e. removed the instructions to run `python -m nltk.downloader punkt_tab`). NB: Adjust Wiki pages too on release.

In Dockerfile [this data download step](https://github.com/NatLibFi/Annif/blob/automate-nltkdata-download/Dockerfile#L31) is retained, so when using the Dockerimage everything is ready without download needs (otherwise the data would be re-downloaded every time Annif is used in a new container, because the NLTK data storage is not mounted to a volume).

Also in the CI/CD pipeline there is still [this explicit download step](https://github.com/NatLibFi/Annif/blob/45be777cfd12609150973b95d496cf232862526f/.github/workflows/cicd.yml#L104).

BTW, now when modifying this, it would be a good time to consider changing the location of the NLTK data storage, which is discussed [here](https://github.com/NatLibFi/Annif/pull/540#issuecomment-996714399). However, the current `~/nltk_data/` is not bad, I think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,6 @@ The recommended way is to install Annif from
     source annif-venv/bin/activate
     pip install annif
 
-You will also need NLTK data files:
-
-    python -m nltk.downloader punkt_tab
-
 Start up the application:
 
     annif
@@ -112,10 +108,6 @@ By default development dependencies are included. Use option `-E` to install dep
 Enter the virtual environment:
 
     poetry shell
-
-You will also need NLTK data files:
-
-    python -m nltk.downloader punkt_tab
 
 Start up the application:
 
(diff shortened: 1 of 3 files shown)


juhoinkinen:
@CodiumAI-Agent /review

juhoinkinen:
One line is missing test coverage, but that line "should never be run" and the involved logic is very simple.

osma (approved):
LGTM

PR #805, opened 2024-09-30
Upgrade Docker baseimage to Python 3.12

juhoinkinen:
Now when Python 3.12 will be fully supported (after https://github.com/NatLibFi/Annif/pull/796), lets upgrade to it in Docker image.

The upgrade should bring some performance improvements.


osma (approved):
LGTM. We need to test if old models can still be loaded after this.

juhoinkinen:
I checked the projects at ai.dev.finto.fi, and other projects than those using NN ensemble seem to work.

PR #807, opened 2024-09-30
Add `--trust-repo` option to `download` CLI command

juhoinkinen:
Downloading projects from Hugging Face Hub should be done from trusted sources only. 

After this PR, if the repository hasn’t been used for downloads previously (i.e., it doesn’t appear in the Hugging Face Hub cache on local system), the `--trust-repo` option needs to be used.


osma (approved):
lgtm

PR #818, opened 2024-11-12
EstNLTK analyzer

osma:
This PR adds a new analyzer to support lemmatization using [EstNLTK](https://github.com/estnltk/estnltk), a natural language analysis toolkit for the Estonian language.

Note that the indirect dependencies of EstNLTK are quite large, with around ~500MB of libraries.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -85,7 +85,7 @@ jobs:
         fi
         # For Python 3.10:
         if [[ ${{ matrix.python-version }} == '3.10' ]]; then
-          poetry install -E "fasttext spacy";
+          poetry install -E "fasttext spacy estnltk";
           # download the small English pretrained spaCy model needed by spacy analyzer
           poetry run python -m spacy download en_core_web_sm --upgrade-strategy only-if-needed
         fi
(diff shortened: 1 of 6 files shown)


osma:
The test coverage is not 100%. I think that's due to how optional backends are handled in the initialization code, that I copied from the spaCy analyzer. The same problem is already there for spaCy, so I think it would make sense to fix that first in main, then apply the same solution in this PR.

osma:
I added a brief [section](https://github.com/NatLibFi/Annif/wiki/Analyzers#estnltk-analyzer) about this analyzer on the wiki page for Analyzers.

I don't see any mention of specific analyzers in the Annif tutorial, so I don't think this needs to be mentioned there.

juhoinkinen (approved):
:+1:

osma:
I realized that EstNLTK is GPL licensed, so probably should be mentioned alongside YAKE when we talk about licensing. Need to fix that before merging this PR.

osma:
I [asked](https://github.com/estnltk/estnltk/issues/123) the EstNLTK developers about their thoughts on license compatibility. They were very positive about the Annif integration. Now that the licensing situation is at least somewhat clarified in the README, I don't think there are obstacles for merging this, so I will do it now.

PR #820, opened 2024-11-21
Smarter initialization of optional analyzers

osma:
This PR fixes a problem with the initialization of optional analyzers that I noticed while working on the EstNLTK analyzer in PR #818 . For the spaCy analyzer, the code in `analyzer/__init__.py` never noticed the case when spaCy is not installed, because the `import spacy` statement only happens inside a method (due to performance reasons).

This PR adds a new static method `is_available` to all Analyzer implementations and uses that to determine whether the analyzer is available (i.e. the optional dependency is installed) or not. For spaCy and Voikko analyzers, the method is implemented by using importlib.find_spec to see if the necessary modules are available or not. This should be faster than actually importing the module, which can be done at a later time within the methods if it's actually needed.

### DIFF
--- a/annif/analyzer/__init__.py
+++ b/annif/analyzer/__init__.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 import annif
 from annif.util import parse_args
 
-from . import simple, simplemma, snowball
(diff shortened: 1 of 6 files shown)


juhoinkinen (approved):
LGTM

PR #825, opened 2024-12-20
Optimize MLLM ambiguity calculation

### LINKED ISSUE #822: Slow calculation of ambiguity feature in MLLM
Dear Osma, dear annif-team, 

As discussed in the annif-users-group (https://groups.google.com/g/annif-users/c/8d3AL4LAzBQ), I have added the debugging lines and performed the suggest operation with an MLLM model trained with the full GND vocabulary set we use (1.4M subjects) on a document with a long processing time (305.72 sec.). Please find the ziped _tsets.jsonl_ file attached to this issue.

Best regards
Clemens

[tsets.zip](https://github.com/user-attachments/files/18147470/tsets.zip)

osma:
This PR is an attempt to optimize the calculation of the ambiguity feature in MLLM. In cases of extremely many matches, the ambiguity calculation can take a long time, as reported by @RietdorfC in #822.

This PR changes the calculation so that it will first group together TokenSets with the same tokens; for example, all concepts with the same (or very similar) label can be considered together instead of calculating ambiguity for each of them separately. Due to the quadratic nature of the ambiguity calculation (it's O(N^2) where N is the number of matches found in a piece of text), reducing N by grouping TokenSets may reduce the amount of necessary comparisons quite drastically.

I'm leaving this as a draft PR because this needs to be tested further. I'm not yet 100% sure that the calculation result matches the original.

Closes #822

### DIFF
--- a/annif/lexical/tokenset.py
+++ b/annif/lexical/tokenset.py
@@ -20,7 +20,7 @@ class TokenSet:
         subject_id: int | None = None,
         is_pref: bool = False,
     ) -> None:
-        self._tokens = set(tokens)
+        self._tokens = frozenset(tokens)
         self.key = tokens[0] if len(tokens) else None
         self.subject_id = subject_id
         self.is_pref = is_pref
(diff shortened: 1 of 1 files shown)


osma:
This was tested by @RietdorfC in https://github.com/NatLibFi/Annif/issues/822#issuecomment-2621974515 and the results were good. I've reviewed the code once more and I think it's doing what it should.

juhoinkinen (approved):
LGTM

PR #832, opened 2025-02-03
Make sure permissions of saved files adhere to umask setting

### LINKED ISSUE #829: Annif does not adhere umask setting for data files
Annif uses `tempfile.mkstemp()` for saving some vocabulary and project data files in the [`atomic_save()` method](https://github.com/NatLibFi/Annif/blob/736353f3b57f8647a8fa127c7a3b0cbba5ef0b31/annif/util.py#L40). The file permissions of the files created are always only read and write for the owner; the permissions cannot be set by umask ([Python docs](https://docs.python.org/3/library/tempfile.html#tempfile.mkstempare)).

For example:
```
umask 0007  # normally results to permission -rw-rw----

annif load-vocab arch tests/corpora/archaeology/yso-archaeology.ttl
ls -l data/vocabs/arch/
total 76
-rw------- 1 <myuser> <mygroup> 10597 Jan 27 13:10 subjects.csv
-rw------- 1 <myuser> <mygroup> 29177 Jan 27 13:10 subjects.dump.gz
-rw-rw---- 1 <myuser> <mygroup> 32252 Jan 27 13:10 subjects.ttl
```
As seen above, only `subjects.ttl` gets the desired permissions (because that file is [just copied](https://github.com/NatLibFi/Annif/blob/736353f3b57f8647a8fa127c7a3b0cbba5ef0b31/annif/corpus/skos.py#L148) to the vocabulary directory).

This is a problem for multiuser environments. The same problem might exist also in some of the libraries Annif uses and which save data files, but not in Omikuji. The backends use that create the `vectorizer` file are affected via Annif's `atomic_save()`.

At least these approaches have been used for resolving this issue:
- Get the umask value and then chmod the files with the value, like in [this SO answer](https://stackoverflow.com/a/44130605/19835364).
- Do not use tempfile, but create the data files directly in the destination directory, but with filenames with UUIDs, like in this PR.

osma:
Fixes #829 

Many types of files saved by Annif using the `atomic_save()` operation had tighter permissions than expected. This is because they were first saved as temporary files, then renamed.

This PR makes sure that the file permissions respect the umask setting. The unit test verifies that this works using different commonly used umask values.

### DIFF
--- a/annif/backend/nn_ensemble.py
+++ b/annif/backend/nn_ensemble.py
@@ -272,7 +272,7 @@ class NNEnsembleBackend(backend.AnnifLearningBackend, ensemble.BaseEnsembleBacke
         lmdb_path = os.path.join(self.datadir, self.LMDB_FILE)
         if not cached and os.path.exists(lmdb_path):
             shutil.rmtree(lmdb_path)
-        return lmdb.open(lmdb_path, map_size=lmdb_map_size, writemap=True)
+        return lmdb.open(lmdb_path, map_size=lmdb_map_size, writemap=True, mode=0o775)
 
     def _fit_model(
         self,
--- a/annif/util.py
+++ b/annif/util.py
@@ -48,7 +48,7 @@ def atomic_save(
         newname = fn.replace(tempfilename, os.path.join(dirname, filename))
         logger.debug("renaming temporary file %s to %s", fn, newname)
         os.rename(fn, newname)
-        umask = os.umask(0o666)
+        umask = os.umask(0o777)
         os.umask(umask)
         os.chmod(newname, 0o666 & ~umask)
 
--- a/annif/util.py
+++ b/annif/util.py
@@ -48,6 +48,9 @@ def atomic_save(
         newname = fn.replace(tempfilename, os.path.join(dirname, filename))
         logger.debug("renaming temporary file %s to %s", fn, newname)
         os.rename(fn, newname)
+        umask = os.umask(0o666)
+        os.umask(umask)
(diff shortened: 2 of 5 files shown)


juhoinkinen:
I quickly trained some projects with the code in this PR, and noted that while `atomic_save()` now behaves as expected permission-wise, the `nn-train.mdb/` directory and the files in it (created by NN ensemble backend) does not get the permissions following umask; e.g. with umask for read-write for group, 0007, the write bit is missing:
```
$ ls -ld data-tmp/./projects/nn-ensemble-fi/nn-train.mdb/
drwxr-x--- 2 <myuser> <mygroup>  4096 Feb  4 17:17 data-tmp/./projects/nn-ensemble-fi/nn-train.mdb/

$ ls -l data-tmp/./projects/nn-ensemble-fi/nn-train.mdb/
total 124
-rw-r----- 1 <myuser> <mygroup> 1073741824 Feb  4 17:17 data.mdb
-rw-r----- 1 <myuser> <mygroup>                8192 Feb  4 17:17 lock.mdb
```
I wonder if there is something that can be done for this too...?

Otherwise LGTM!

juhoinkinen (approved):

juhoinkinen:
It looks like [`lmdb.open()`](https://github.com/NatLibFi/Annif/blob/5e5f8832fc759b77e96b346ced096806c8cc9d00/annif/backend/nn_ensemble.py#L275) could uses a [mode argument](https://github.com/jnwatson/py-lmdb/blob/master/lmdb/cffi.py#L718), which by default is `O_0755`. Maybe this could be get from the umask.

PR #833, opened 2025-02-07
Update dependencies for v1.3 release

juhoinkinen:
Dependencies not upgraded to the newest version:
- NumPy to 2.2.2, because it requires Python >=3.10 and already [NumPy 2.0](https://github.com/numpy/numpy/releases/tag/v2.0.0) could lead to incompatibilities (to be confirmed)
- SciPy to [1.15.1](https://github.com/scipy/scipy/releases/tag/v1.15.0), because it requires Python >=3.10


pyproject.toml
```diff
@@ -67,7 +67,7 @@ pytest-watch = "*"
 pytest-flask = "*"
 flake8 = "*"
 bumpversion = "*"
-black = "24.*"
+black = "25.*"
```
osma, line 70: I'm a little surprised if this didn't change anything in our codebase?

osma (approved):
Looks reasonable to me. It would be good to check model compatibility so that any issues can be reported in the Release Notes.

pyproject.toml
juhoinkinen: But so it seems. Even when ran with `--preview` option, black does not detect anything to change (although in one of my helper scripts the `--preview` option does find some format change).

juhoinkinen:
> Looks reasonable to me. It would be good to check model compatibility so that any issues can be reported in the Release Notes.

The models at ai.dev.finto.fi remain working, when #834 is included.

PR #834, opened 2025-02-07
Ensure compatibility with MLLM models trained using Annif 1.2 or older

osma:
This adds a check to ensure that MLLM TokenSet objects use frozenset to store tokens. In models prior to Annif 1.3, the tokens were stored in a mutable set. The `__setstate__` method is called during unpickling and handles the conversion so that older models will still work.

We could drop this later on if compatibility with older models is no longer relevant, for example if there is another reason that they will stop working anyway.

- Related to #822 
- Post fix to #825

### DIFF
--- a/annif/lexical/tokenset.py
+++ b/annif/lexical/tokenset.py
@@ -41,6 +41,13 @@ class TokenSet:
 
         return other._tokens.issubset(self._tokens)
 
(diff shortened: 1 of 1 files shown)


juhoinkinen (approved):
I tested this quickly, works.

PR #835, opened 2025-02-11
Add CI/CD jobs for testing on Windows and MacOS

juhoinkinen:
WIP


juhoinkinen:
I got Poetry installation by pipx on MacOS runners finally work, but it does not use cache.

PR #838, opened 2025-03-13
Fix Swagger UI 500 error `Failed to load API definition.`

juhoinkinen:
There is a [bug in Connexion 3.2.0](https://github.com/spec-first/connexion/issues/2029) affecting Swagger UI: Just after Annif startup its Swagger UI (`/v1/ui`) works, but if any other(?) API request has been made, a 500 error is thrown when trying to access it.

A workaround for now is to downgrade to Connexion ~3.1.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = ">=3.9,<3.13"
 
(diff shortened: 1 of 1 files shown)


juhoinkinen:
This was the release 1.3.1: https://github.com/NatLibFi/Annif/releases/tag/v1.3.1

PR #843, opened 2025-04-01
Disable /learn API method by default

juhoinkinen:
The learn functionality is rarely used, so it is best to be disabled by default for API requests.

This pull request introduces a new feature that restricts learning operations on projects unless explicitly enabled with `allow_learn` project config parameter.

Adds a new `403` response in the OpenAPI specification for trying to access the `/v1/<project>/learn` when the project is not configured to allow this.

### DIFF
--- a/annif/project.py
+++ b/annif/project.py
@@ -19,6 +19,7 @@ from annif.exception import (
     ConfigurationException,
     NotInitializedException,
     NotSupportedException,
+    OperationFailedException,
 )
 from annif.util import parse_args
 from annif.vocab import SubjectIndexFilter
@@ -295,7 +296,12 @@ class AnnifProject(DatadirMixin):
         beparams = backend_params.get(self.backend.backend_id, {})
         corpus = self.transform.transform_corpus(corpus)
         if isinstance(self.backend, annif.backend.backend.AnnifLearningBackend):
-            self.backend.learn(corpus, beparams)
+            if annif.util.boolean(self.config.get("allow_learn", False)):
+                self.backend.learn(corpus, beparams)
+            else:
+                raise OperationFailedException(
+                    "Learning not enabled for project", project_id=self.project_id
+                )
         else:
             raise NotSupportedException(
                 "Learning not supported by backend", project_id=self.project_id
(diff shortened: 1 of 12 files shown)


osma:
The existing visibility settings only affect the REST API. I think this could also be scoped the same way; the CLI probably shouldn't care about this setting.

annif/rest.py
```diff
@@ -248,6 +258,11 @@ def learn(
     try:
         corpus = _documents_to_corpus(body, project.subjects)
         project.learn(corpus)
+    except OperationFailedException as err:
+        if str(err) == "Learning not enabled for project":
+            return learning_not_enabled_error(project_id)
+        else:
+            return server_error(err)
```
juhoinkinen, line 265: I [added a test](https://github.com/NatLibFi/Annif/pull/843/commits/0fe8f0b079ebad7db786c94ef16d59032df1539e) with mocking just to have test coverage also on this line (but in reality this line should not be reached, but this could change in the future?). Not very good coding?

annif/project.py
```diff
@@ -295,7 +296,12 @@ def learn(
         beparams = backend_params.get(self.backend.backend_id, {})
         corpus = self.transform.transform_corpus(corpus)
         if isinstance(self.backend, annif.backend.backend.AnnifLearningBackend):
-            self.backend.learn(corpus, beparams)
+            if annif.util.boolean(self.config.get("allow_learn", False)):
+                self.backend.learn(corpus, beparams)
+            else:
+                raise OperationFailedException(
```
osma, line 302: How about defining and raising here a new type of exception such as OperationNotEnabledException (or something along those lines)? This would indicate that the operation failed not because of a technical failure, but due to a configuration setting.

annif/rest.py
```diff
@@ -248,6 +258,11 @@ def learn(
     try:
         corpus = _documents_to_corpus(body, project.subjects)
         project.learn(corpus)
+    except OperationFailedException as err:
+        if str(err) == "Learning not enabled for project":
```
osma, line 262: If a custom exception was used to indicate disabled functionality, instead of OperationFailedException, there would be no need to perform the fragile string comparison here.

osma (changes requested):
I suggested defining and using a new exception type for this case. Otherwise LGTM and can be merged!

annif/rest.py
juhoinkinen: The test with mocking is no more needed when using the new NotEnabledException.

PR #849, opened 2025-04-07
Drop Python 3.9 support

juhoinkinen:
Drops Annif's support for Python 3.9, rotates Python versions in CI/CD pipeline and updates Python version for Scrutinizer to 3.12.

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -23,7 +23,7 @@ jobs:
     - name: "Prepare: restore caches, install Poetry, set up Python"
       uses: ./.github/actions/prepare
       with:
-        python-version: "3.11"
+        python-version: "3.12"
         poetry-version: ${{ env.POETRY_VERSION }}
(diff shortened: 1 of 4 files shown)


osma:
Codecov shows a number of [indirect changes](https://app.codecov.io/gh/NatLibFi/Annif/pull/849/indirect-changes) to coverage. Basically around 20 code lines related to backends that are not installed are not being covered by the tests. Why is that? Is it possible that all optional features get installed for all Python versions, or something along those lines?

juhoinkinen:
When versions were rotated and new dependencies got installed to a Python version specific job, the old dependencies remained in cache, so all features were available.

I cleared the caches. Now [indirect changes](https://app.codecov.io/gh/NatLibFi/Annif/pull/849/indirect-changes) actually increases with few lines (surprising, why?).

osma:
Indeed that's a bit mysterious. I can't think of a reason for the change.

osma (approved):
lgtm

PR #859, opened 2025-07-08
WIP LLM ranking/scoring backend

### LINKED ISSUE #856: LLM ranking/scoring backend
Following the success by [DNB-AI-Project](https://arxiv.org/html/2504.21589v1) (@mfakaehler &  Lisa Kluge) at LLMs4Subjects using LLM to rank candidate subjects, we have been experimenting with a similar approach incorporated as an Annif backend. ~The code currently is in the branch [experiment-llm-ensemble-backend](https://github.com/NatLibFi/Annif/blob/experiment-llm-ensemble-backend/annif/backend/llm_ensemble.py).~ Edit: See the PR #859. 

The name should maybe not be "LLM ensemble" as it is now, because the functionality does not necessitate multiple source projects: LLM could very well just score the subjects given by just one source project (which could be a simple ensemble). Possible name alternatives: "LLM ranker", "LLM scorer", "LLM grader".

The chosen name should allow the possible implementation of also other backends with similar ranking/scoring functionalities.

juhoinkinen:
Closes #856.

### DIFF
--- a/annif/backend/llm_ensemble.py
+++ b/annif/backend/llm_ensemble.py
@@ -154,6 +154,7 @@ class LLMEnsembleBackend(BaseLLMBackend, ensemble.EnsembleBackend):
         "llm_exponent": 1.0,
         "labels_language": "en",
         "sources_limit": 10,
+        "max_workers": 32,
     }
 
     def get_hp_optimizer(self, corpus: DocumentCorpus, metric: str) -> None:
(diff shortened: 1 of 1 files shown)


juhoinkinen:
Remainder for implementation: We could check if querying the LLM multiple times and averaging the results gives an improvement (this could work when using a non-zero temperature).

This approach is used in https://kata.rara.ee/ service.

juhoinkinen:
When some suitable [Jev](https://typesafe.ai/blog/introducing-system-one-models-and-jev)-like framework has matured enough, this backend is probably best implemented using it instead of an LLM.

PR #866, opened 2025-08-07
WIP: CLI suggest option for additional metadata

osma:
Part of #813

Issue #813, opened 2024-11-01
Flexible fusion backend


c-poley:
With Annif, it is possible to use several specialised models for prediction in an ensemble. However, all models in an Annif ensemble, can only be given one specific single kind of text for prediction, so it is not possible to pass on different kinds of text to each single model.
The only way to adapt the text for prediction currently is to use the transform parameter. We can make use of that parameter to read either a limited amount of characters from the beginning or read all of the text. A parameter that would enable us to set a specific range (from character x to character y) to be read from the text would give us additional way to specify/cut down the text for specific models.

Could the annif ensemble functionality be extended in such a way that the individual models of an ensemble could be given different kinds of text (expressions of a document) for processing?

Another flexibility for the ensemble functionality would be the use of subsets of vocabularies for individual models in ensembles as discussed in issue #596 .

Anyway, the interface for the predictions need enhancments. In the following we describe some ideas we already discussed a few weeks ago:

1. Allow to submit text as structured data:
Use json: 
`-d '%7B%22headline%22%3A%22Wonderful%22%2C%22fulltext%22%3A%22Oh%2C%20what%20a%20wonderful%20world%22%7D'`
or
Use xml tags:
`-d '%3Cheadline%3EWonderful%3C%2Fheadline%3E%3Cfulltext%3EOh%2C%20what%20a%20wonderful%20world%3C%2Ftext%3E'`
... with the possibility to define the tags at the right places in the projects.conf, like:
`submitted_text=headline,' ',fulltext`
The empty space between the "headline" and the "fulltext" defines the character(s), how to glue the parts of the text to submit.
`submitted_text=headline,'.',toc`
Here is a headline and a toc to submit. They are connected with a "."

2. An approach on the way to allow a fusion is an enhancement of the limit parameter. This will allow us to define the part of the submitted text. 
In the projects.conf we only need a little enhancement, that defines the starting point and the number of characters to proceed:
`transform=limit(500,2000)`

We (and we think the whole community) would really benefit from the implementation of a fusion with freely configurable structured data like in (1). We have to admit, that the usage of structured data would be the most favorite and clean implementation.

Best regards,
Christoph, Frank, Jan-Helge and Sandro from the German National Library

juhoinkinen:
A third possible approach to passing different variants/subsets of a full text to different backends, a kind of fusion of the approaches (1) and (2):

[...]

PR #867, opened 2025-08-07
Flexible fusion part 4: REST API document metadata support

osma:
4th part of #813, after #866.

This PR adds support for document metadata into the REST API methods `suggest`, `suggest_batch` and `learn`. It also adds metadata support to the `http` backend so that it will pass along any document metadata to the REST API call.

# `suggest` method

The `suggest` method accepts a request body of the type `application/x-www-form-urlencoded`. This is a bit old-fashioned and means that the basic data structure is a flat list of `field=value` pairs. Before this PR, there were only a few valid field names in use, including `text`, `limit` and `threshold`. I considered several alternatives on how to add support for metadata fields, which could have arbitrary names:

1. Use the metadata field directly as the field name, e.g. `title=As We May Think`
2. Use the metadata field with a prefix, e.g. `metadata_title=As We May Think`
3. Use an indexed field name that looks a bit like a dict lookup, e.g. `metadata[title]=As We May Think`
4. Use a structured JSON value that includes all metadata fields and their values, e.g. `metadata={"title": "As We May Think"}`

Option 1 has the problem that the metadata field names could clash with existing (or future) field names. Option 3 is a bit difficult to implement both in the client and in the server code, though could be done (this pattern is sometimes used in PHP web applications). Option 4 seemed a bit weird to me. So I went for Option 2, using the prefix `metadata_`.

## How to test

Sadly, it appears that the Swagger UI for testing the REST API does not understand `additionalProperties` in this context so it's not possible to include metadata properties in a test request. Here is a way of doing it using `curl` on the command line (note the last part with `metadata_title`):

```
curl  'http://127.0.0.1:5000/v1/projects/myproject/suggest' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'text=my_test&metadata_title=As We May Think'
```

# `suggest_batch` and `learn` methods

These methods already accept a structured JSON request body, so it was fairly simple to add a new, optional `metadata` field with an object as the value, storing metadata fields and their values.

## How to test

This can be tested by sending a request body like this to `suggest_batch`:

```json
{
  "documents": [
    {
      "text": "A quick brown fox jumped over the lazy dog.",
      "document_id": "doc-1234",
      "metadata": {
        "title": "As We May Think"
      }
    }
  ]
}
```

for example using curl:

```
curl -X 'POST' \
  'http://127.0.0.1:5000/v1/projects/myproject/suggest-batch?limit=10&threshold=0' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "documents": [
    {
      "text": "A quick brown fox jumped over the lazy dog.",
      "document_id": "doc-1234",
      "metadata": {
        "title": "As We May Think"
      }
    }
  ]
}'
```

The Swagger UI also supports this. You simply have to edit the JSON request body. It already contains placeholder keys and values for metadata.

# `http` backend

The `http` backend currently only supports the plain `suggest` method (one document at a time), not `suggest_batch`. I simply amended the code to pass along any metadata values using the `metadata_` prefix for field names.

### DIFF
--- a/annif/backend/http.py
+++ b/annif/backend/http.py
@@ -70,6 +70,8 @@ class HTTPBackend(backend.AnnifBackend):
         self, doc: Document, params: dict[str, Any]
     ) -> list[SubjectSuggestion]:
         data = {"text": doc.text}
+        for key, value in doc.metadata.items():
+            data[f"metadata_{key}"] = value
         if "project" in params:
             data["project"] = params["project"]
         if "limit" in params:
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -392,13 +392,21 @@ components:
         document_id:
           type: string
           example: doc-1234
-      description: A document with an optional identifier
+        metadata:
+          type: object
+          additionalProperties:
+              type: string
+      description: A document with an optional identifier and/or metadata
     IndexedDocument:
       type: object
       properties:
         text:
           type: string
           example: A quick brown fox jumped over the lazy dog.
+        metadata:
+          type: object
+          additionalProperties:
+              type: string
         subjects:
           type: array
           items:
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -66,6 +66,10 @@ paths:
       tags:
       - Automatic subject indexing
       summary: suggest subjects for a given text
+      description: >
+        This method suggests relevant subjects for a given input text (with
+        optional metadata). It supports optional parameters to control the
+        number of results, score threshold, and language of the subject labels.
       operationId: annif.rest.suggest
       parameters:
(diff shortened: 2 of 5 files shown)


osma:
Ping @c-poley @RietdorfC more things for you to test, maybe :)

juhoinkinen (approved):
LGTM

We could mention the possibility to send metadata in the /suggest method description (which we don't have at all currently), as suggested in the workaround of the [Swagger UI issue](https://github.com/swagger-api/swagger-ui/issues/8576), but I don't know if that is really needed.

osma:
@juhoinkinen I added descriptions for the suggest and suggest_batch methods, mentioning also metadata.

Merging this now.

PR #872, opened 2025-08-13
Add JSON fulltext corpus format

### LINKED ISSUE #868: Metadata support in fulltext corpus format
One of the pieces necessary for flexible fusion (#813) is support for additional metadata in a full text corpus format. Metadata support for the new CSV short text format was already added in #864.

The current full text corpus format is a directory of files. There is always a `.txt` file with raw text (for example converted from a PDF file, as in the Annif tutorial data sets) and optionally a `.tsv` file with the same basename, containing subject annotations.

If we want to extend this format so that documents may contain metadata (key=value pairs), there are a few different ways of implementing it:

1. Add a new (third) file with the same basename alongside the `.txt` and `.tsv` files, containing the metadata. The metadata file could be in a simple format such as YAML, JSON or CSV (with columns "field" and "value").
2. Make it possible to include metadata within the `.txt` file using [YAML front matter](https://jekyllrb.com/docs/front-matter/) which is often used with Markdown files, for example [on GitHub](https://docs.github.com/en/contributing/writing-for-github-docs/using-yaml-frontmatter).
3. Same as 2, but require a `.md` extension for corpus files with metadata; so they are now essentially Markdown files, although raw text is usually also valid Markdown.

# Option 1: extra file with metadata

Here we assume that the text is in the file `12345.txt`, with subjects in `12345.tsv`. Here are examples of different variants:

## Variant 1a. Metadata in a YAML file `12345.yml`

```yaml
title: As We May Think
author: Bush, Vannevar
```

## Variant 1b. Metadata in a JSON file `12345.json`

```json
{"title": "As We May Think", "author": "Bush, Vannevar"}
```

## Variant 1c. Metadata in a CSV file `12345.csv`

```csv
field,value
title,"As We May Think"
author,"Bush, Vannevar"
```

# Option 2: Metadata within TXT file using YAML Front Matter

The `12345.txt` file may contain a YAML metadata header section, separated by triple dashes above and below:

```
---
title: As We May Think
author: Bush, Vannevar
---
As We May Think

“Consider a future device ... in which an individual stores all his books, records, and communications, and which is mechanized so that it may be consulted with exceeding speed and flexibility. It is an enlarged intimate supplement to his memory.”
[...]
```

# Option 3: Markdown file with YAML front matter

This is the same as 2, but the file must be called `12345.md` if it contains YAML front matter. `12345.txt` works as before, but cannot contain metadata.

# Thoughts

Any of these could be made to work. If we go for YAML front matter (options 2 and 3), then another question that arises is whether the same syntax should be supported also in `annif suggest myproject <document.txt` and the REST API methods `suggest`, `suggest_batch` and `learn`, i.e., if all input texts are potential sources of metadata, or whether the parsing is only done in operations that expect a full text corpus (CLI commands `train`, `eval`, `index`, `learn`...).

Opinions?

osma:
This PR adds a new JSON-based fulltext corpus format, based on the discussion in issue #868 (ping @RietdorfC @c-poley).

It extends the existing support for document directories. Previously, the directory had to contain `.txt` files, with gold standard subjects stored in `.tsv` (or `.key`) files with the same basename. This PR adds another option: the directory may instead (or also) contain `.json` files with JSON data of the following form:

```json
{
  "text": "A quick brown fox jumped over the lazy dog.",
  "metadata": {
    "title": "As We May Think",
    "author": "Bush, Vannevar"
  },
  "subjects": [
    { "uri": "http://www.yso.fi/onto/yso/p817", "label": "future" },
    { "uri": "http://www.yso.fi/onto/yso/p3295", "label": "visions (prospects)" },
    { "uri": "http://www.yso.fi/onto/yso/p15527", "label": "science fiction" }
  ]
}
```

All top level fields (`text`, `metadata` and `subjects`) are optional. Subject labels are also optional and included only for illustration; they do not affect the parsing. The JSON format is the same as the one used by the REST API method `learn`, specified by the [IndexedDocument schema](https://github.com/NatLibFi/Annif/blob/140a1b63748639426f9e9d2d793116acf99bdfce/annif/openapi/annif.yaml#L408-L431) in the OpenAPI specification, except ~~there are no required fields~~ the set of required fields is a bit different (either `text` or `metadata` is required on the top level, and if subjects are included, they must specify either `uri` or `label`).

Closes #868.

### DIFF
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -332,7 +332,7 @@ def run_index(
     results = project.suggest_corpus(documents, backend_params).filter(limit, threshold)
 
     for docfilename, suggestions in zip(documents, results):
-        subjectfilename = re.sub(r"\.txt$", suffix, docfilename)
+        subjectfilename = re.sub(r"\.(txt|json)$", suffix, docfilename)
         if os.path.exists(subjectfilename) and not force:
             click.echo(
                 "Not overwriting {} (use --force to override)".format(subjectfilename)
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -328,11 +328,11 @@ def run_index(
         raise click.BadParameter(f'language "{lang}" not supported by vocabulary')
     backend_params = cli_util.parse_backend_params(backend_param, project)
 
-    documents = DocumentDirectory(directory, require_subjects=False)
-    results = project.suggest_corpus(documents, backend_params).filter(limit, threshold)
+    corpus = DocumentDirectory(directory, require_subjects=False)
+    results = project.suggest_corpus(corpus, backend_params).filter(limit, threshold)
 
-    for docfilename, suggestions in zip(documents, results):
-        subjectfilename = re.sub(r"\.(txt|json)$", suffix, docfilename)
+    for doc, suggestions in zip(corpus.documents, results):
+        subjectfilename = re.sub(r"\.(txt|json)$", suffix, doc.file_path)
         if os.path.exists(subjectfilename) and not force:
             click.echo(
                 "Not overwriting {} (use --force to override)".format(subjectfilename)
--- a/annif/cli.py
+++ b/annif/cli.py
@@ -331,7 +331,7 @@ def run_index(
     documents = DocumentDirectory(directory, require_subjects=False)
     results = project.suggest_corpus(documents, backend_params).filter(limit, threshold)
 
-    for (docfilename, _), suggestions in zip(documents, results):
+    for docfilename, suggestions in zip(documents, results):
         subjectfilename = re.sub(r"\.txt$", suffix, docfilename)
         if os.path.exists(subjectfilename) and not force:
             click.echo(
--- a/annif/corpus/document.py
+++ b/annif/corpus/document.py
@@ -79,6 +79,17 @@ class DocumentDirectory(DocumentCorpus):
             )
         return Document(text=text, subject_set=subjects)
 
+    def _subjects_to_subject_set(self, subjects):
+        subject_ids = []
+        for subj in subjects:
+            if "uri" in subj:
+                subject_ids.append(self.subject_index.by_uri(subj["uri"]))
+            else:
+                subject_ids.append(
+                    self.subject_index.by_label(subj["label"], self.language)
+                )
+        return SubjectSet(subject_ids)
+
     def _read_json_file(self, filename: str) -> Document | None:
         if os.path.getsize(filename) == 0:
             logger.warning(f"Skipping empty file {filename}")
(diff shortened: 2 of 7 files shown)


osma:
@juhoinkinen 

> [jsonschema](https://github.com/python-jsonschema/jsonschema) could be used for validating the JSON data and for error handling. However, for such a simple schema, it might be unnecessary.

Yes, that's a great idea! I think that the schema is just about complex enough to justify JSON Schema validation. I'll look into it. (I already added basic error handling for cases where the file is empty or has broken JSON syntax)

osma:
Note to self: there's now a subtle bug in `annif index`. The list of paths returned by DocumentDirectory may contain broken/empty files, because their content is not checked during initial iteration. That means that the list of file names may get out of sync with the list of suggestion results, causing results to be written into the wrong files.

osma:
@juhoinkinen I added JSON Schema validation, made it possible to specify subjects via labels instead of URIs (as in the .txt/.tsv/.key format), and fixed some issues with `annif index`. I can't think of anything else right now, so from my side, this is getting ready for merging. Do you want to take another look?

Also @c-poley @RietdorfC feel free to test and comment!

RietdorfC:
Hi @osma,
we started testing the JSON fulltext format but unfortunally encountered one error and one problem.
The error occured when we tried to run the `annif index` command. We prepared two files from the same document: a .txt file containing simply the full text of the document (you can find this file attached as 1000694364.txt) and one .json file with an additional metadata field for the title and its subjects (you can find this file attached as 1000694364.json). We placed both files in directories to use the `annif index` command. In our test, we used the same model to run the index command (an omikuji model trained on full texts). To test if both files could be processed, we ran the following commands:
```
annif index A-omikuji-FT tmp-txt/
annif index A-omikuji-FT tmp-json/
```
The first one completed without any problems and produced the expected output, but the second one, which refers to the JSON-file, generated an error massage (you can find the message in the file error.txt). We couldn’t trace the error back to our input, but maybe we overlooked something?

A question occured when we tried to run the `annif suggest` command on a JSON file. First, we tested the `annif suggest` command with an omikuji model without the `transform=select(title)` parameter. We tested the model on a file from the classic full text corpus format (same .txt file as above) and got a valid result. Then we tested the same model but on a JSON file (same as mentioned above) and also got valid results.
Next, we tested a omikuji model with the new parameter `transform=select(title)` set. First, we tested it on the .txt file and, as expected got no result, as there is no title for the model to process. Second, we tested the model on the JSON file but again got no result, even though there is a title given in the metadata section of the JSON file. Why didn't it work? You can find the complete input and output in the file suggest.txt.

Thanks a lot and best regards
Clemens

[1000694364.txt](https://github.com/user-attachments/files/21792864/1000694364.txt)
[1000694364.json](https://github.com/user-attachments/files/21792865/1000694364.json)
[error.txt](https://github.com/user-attachments/files/21792866/error.txt)
[suggest.txt](https://github.com/user-attachments/files/21792867/suggest.txt)

osma:
Thank you @RietdorfC for testing and for reporting the problems. 

For the first problem, it appears that `subject_index` is not properly set for some reason. For the second problem, I'm not sure what happened.

I will investigate and see if I can reproduce these problems.

RietdorfC:
Hi @osma, 
Yes, we used the newest version with the final pull request merged.

PR #878, opened 2025-08-19
Remove gensim dependency in tfidf backend

osma:
This PR reimplements the core `tfidf` backend functionality (tf-idf vector similarity calculations) using scikit-learn instead of gensim, and drops the gensim dependency. Gensim is a bit problematic because it doesn't currently work on Python 3.13 or with recent versions of SciPy, and we are only using a small part of its functionality.

TODO:

- [ ] benchmark performance
- [ ] benchmark quality


juhoinkinen:
I tested this PR using the [TFIDF train exercise](https://github.com/NatLibFi/Annif-tutorial/blob/main/exercises/02_tfidf_project.md)  and [evaluation exercise](https://github.com/NatLibFi/Annif-tutorial/blob/main/exercises/04_evaluate.md) of Annif-tutorial (YSO vocabulary, full training data):

|  | before (main) | after (PR) |
| ---------- | ---------- | ----------------- | 
| train user time (s) | 260.76           |  287.45        | 
| train wall time  | 4:31.23           |  4:56.32      |
| eval user time (s)  | 128.76           |  157.81       |
| eval wall time  | 3:00.78           | 4:09.40        |
| eval F1@5  | 0.1708           | 0.1708        |
| model disk size | 140M           | 76M        |

juhoinkinen:
Now the error message for trying to use TFIDF projects trained using Gensim fail with the message:

    Error: Couldn't initialize backend 'tfidf': tf-idf matrix data/projects/<poroject-id>/tfidf-matrix.npz not found

There could be a message like the [one for Omikuji](https://github.com/NatLibFi/Annif/blob/5c99f3c02765c3700e8ee79041a1751e29af5e61/annif/backend/omikuji.py#L55):

    "TFIDF models trained on Annif versions older than 1.4 cannot be loaded. Please retrain your project."

osma:
@juhoinkinen can you try `annif eval` one more time? I implemented _suggest_batch, which should be a bit faster, and there is less vector mangling now.

I also added the suggested error message when trying to load an older TFIDF model.

juhoinkinen:
I run eval twice with the latest changes and updated the table above: again not much can be said, apart from that performance-wise this seems alright.

PR #879, opened 2025-08-19
Upgrade to Schemathesis 4 and Pytest 9

juhoinkinen:
Updates two development dependencies:
- Schemathesis to version 4
- Pytest to version 9

Also, the updated Schemathesis noticed an issue with the OpenAPI spec, which is corrected: The `/learn` method documentation did not include the 400 error response, which Annif returns when using an empty subject label, e.g.: `[{"text": "A quick brown fox jumped over the lazy dog.", "metadata": {}, "subjects": [{"uri": "http://example.org/subject1", "label": {}}]}]`

However, upgrading Schemathesis required disabling 
- positive data acceptance testing and
- setting extra parameters to queries.

Only this way I managed to make the Schemathesis tests pass. I think it should be possible to enable these again, but for now I'll just open an issue about this.

The Schemathesis configuration is also slightly simplified by removing the suppression of a warning (about filtering-too-much, which was not needed after 32a234d8a466ed55971a3b51af05bcb6ff998691).


juhoinkinen:
In addition to upgrading to Schemathesis 4, this PR currently enables [`strict_validation` in Connexion](https://connexion.readthedocs.io/en/stable/validation.html#parameter-validation) "to disallow any extra parameters that are not defined in your specification". I think this would be a good feature, because then typos in API requests parameters are noticed at once, not after debugging (with strict validation e.g. using `theshold` instead of `threshold` gives 400 Bad request error, without it the extra, nonfunctional parameter is just silently ignored).

However, Annif 1.4 introduced `metadata_*`  parameters to suggest/, suggest_batch/ and learn/ API methods (#867), and to return a 400 error for requests with parameters _not_ starting with `metadata_` I made this: [Add Custom request body validator](https://github.com/NatLibFi/Annif/pull/879/commits/c2aa7fa652debb4a9986f37beb45e45201367cad).

But because of this, Schemathesis 4.7 (correctly, because the API schema does not define the `metadata_` prefix requirement) complains:
```
  | - API rejected schema-compliant request
  | 
  |     Valid data should have been accepted
  |     Expected: 2xx, 401, 403, 404, 409, 5xx
  | 
  | [400] Bad Request:
  | 
  |     `{"type": "about:blank", "title": "Bad Request", "detail": "Extra formData parameter(s) x-schemathesis-additional not in spec", "status": 400}`
  | 
  | Reproduce with: 
  | 
  |     curl -X POST -H 'Content-Type: application/x-www-form-urlencoded' -d 'language=&limit=10&text=&threshold=0&x-schemathesis-additional=' http://localhost/v1/projects/dummy-fi/suggest
```
Schemathesis should be configurable to ignore this, but I did not manage in that for now.

@osma what do you think about enabling the strict validation and the custom validator? Those could and should be done in another PR than this, if we want start using those.

juhoinkinen:
Schemathesis detected and failed unit test due to [this Connexion bug](https://github.com/spec-first/connexion/issues/1860) which makes Connexion 3 crash, essentially with this message:
```
  | schemathesis.core.failures.FailureGroup: Schemathesis found 2 distinct failures
  | 
  | - Undocumented HTTP status code
  | 
  |     Received: 500
  |     Documented: 200, 400
  | 
  | - Server error
  | 
  | [500] Internal Server Error:
  | 
  |     `{"type": "about:blank", "title": "Internal Server Error", "detail": "The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.", "status": 500}`
  | 
  | Reproduce with:
  | 
  |     curl -X POST -H 'Content-Type: application/json' -d $'\x12�\x16��' http://localhost/my-method
  | 
```
I asked how to skig generating non-UTF8 content in [Schemathesis discussions](https://github.com/schemathesis/schemathesis/discussions/3531), because the current way of just ignoring all server errors and wrong status codes is not good.

juhoinkinen:
Also, I removed setting the [strict_validation in Connexion](https://connexion.readthedocs.io/en/stable/validation.html#parameter-validation). That should be another PR if it is done.

(NB: [Schemathesis configuration](https://schemathesis.readthedocs.io/en/stable/reference/configuration/) includes a setting `schema.config.generation.allow_extra_parameters`:
>Controls whether Schemathesis produces unexpected query, header, or cookie parameters. Leave it enabled (default) to exercise additionalProperties: false; set it to false to skip generating those extras entirely.

The setting is now set to False, but if strict validation is enabled in Connexion, `schema.config.generation.allow_extra_parameters` should be set to `True`.)

juhoinkinen:
Schemathesis 4.11 [fixed an issue](https://github.com/schemathesis/schemathesis/pull/3566) regarding filtering of request body content. However, using Schemathesis 4.11 requires pytest 9, so let's postpone this PR until upgrading all dependencies.

Also, the git history had again got messy, which is why I made a force push with after clean up.

osma (approved):
I think this looks sensible and we need to move forward with this anyway.

PR #884, opened 2025-08-27
Upgrade to Poetry 2.0 / Resolve Poetry deprecation warnings

juhoinkinen:
When using [Poetry 2.0](https://python-poetry.org/blog/announcing-poetry-2.0.0), running any Poetry command has given a deprecation warning:

    The "poetry.dev-dependencies" section is deprecated and will be removed in a future version. Use "poetry.group.dev.dependencies" instead.

and `poetry check` gives even more deprecation warnings, so I updated the `pyproject.toml` to resolve these.

I checked with  `poetry check` and [validate-pyproject tool](https://validate-pyproject.readthedocs.io/en/latest/index.html) that `pyproject.toml` should work™ with these modifcations.

I also added classifier "Development Status :: 5 - Production/Stable" and keyword "multilabel-classification" to be shown on the [Annif's PyPI project page](https://pypi.org/project/annif/).

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -10,7 +10,7 @@ on:
 env:
   PIPX_HOME: "/home/runner/.cache/pipx"
   PIPX_BIN_DIR: "/home/runner/.local/bin"
-  POETRY_VERSION: "1.8.2"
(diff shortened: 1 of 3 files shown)


osma (approved):
LGTM

PR #871, opened 2025-08-12
Update dependencies for 1.4 release

juhoinkinen:
Some dependencies cannot be upgraded to latest versions at this time:
- connexion
  - Pin to ~3.1.0 is kept due to a bug: https://github.com/spec-first/connexion/issues/2029
- scipy
  - ~Pin to ~1.13.1 is kept because gensim 4.3.2 does not work with scipy 1.14 (gensim crashes) and gensim 4.3.3 depends on scipy (>=1.7.0,<1.14.0) (thus avoiding the crash)~
     - ~See https://github.com/piskvorky/gensim/issues/3623~
  - Pin scipy to ~1.15.3 because the latest scipy (1.16.1) requires Python >=3.11
- ~stwfsapy~
  - ~Pin to ~0.5.0 is kept because stwfsapy (0.6.0) depends on scipy (>=1.14.1,<2.0.0)~
- numpy
  - ~Pin to 1.26.* is kept:~ Edit: upgrade to numpy 2.2 became possible after [removing gensim dependency](https://github.com/NatLibFi/Annif/pull/878) and upgrading to TF 2.2
    - Upgrade to numpy 2.3 is blocked because of its Python requirement >=3.11
    - ~Upgrade to numpy 2.1 is blocked because tensorflow-cpu (2.18.0) depends on numpy (>=1.26.0,<2.1.0)~
    - ~Upgrade to numpy 2.0 is blocked because gensim (4.3.3) depends on numpy (>=1.18.5,<2.0)~
      - ~See https://github.com/piskvorky/gensim/pull/3615~
- ~tensorflow-cpu~ Edit: upgrade to tensorflow 2.20 became possible after upgrading to numpy 2.0
  - ~Pin to ~2.18.0 is kept:~
    - ~Upgrade to tensorflow-cpu 2.19 is blocked because `no versions of tensorflow-cpu match >2.19.0,<2.20.0 and tensorflow-cpu (2.19.0) depends on tensorflow-intel (2.19.0), tensorflow-cpu (>=2.19.0,<2.20.0) requires tensorflow-intel (2.19.0).` (says Poetry, and the latest [tensorflow-intel in PyPI](https://pypi.org/project/tensorflow-intel/) is v2.18.0)~
        - ~This seems to be a related issue: https://github.com/tensorflow/tensorflow/issues/75415~

Issue #795, opened 2024-07-05
Need for maintained fastText project


juhoinkinen:
I noticed the official [fastText repository](https://github.com/facebookresearch/fastText/forks?include=active&page=1&period=2y&sort_by=stargazer_counts) was archived Mar 19, 2024 and is thus no more maintained at all.

Annif already installs fastText from the [fasttext-wheel PyPI project](https://pypi.org/project/fasttext-wheel/), which builds code from https://github.com/messense/fasttext-wheel (Annif initially switched to that because of [this installation issue](https://github.com/python-poetry/poetry/issues/6113)). 

At some point installation from the fasttext-wheel project probably won't work. In the fastText mailing list someone already [reported](https://groups.google.com/g/fasttext-library/c/4EOM0-S6xHU) a compatibility issue with NumPy 2.0. Actually when I now try to install the fasttext backend dependencies of Annif on Python 3.12 on my laptop, Poetry complains `Unable to find installation candidates for fasttext-wheel (0.9.2)` (two weeks ago, but _after_ NumPy 2.0 release, CICD [successfully installed fastText](https://github.com/NatLibFi/Annif/actions/runs/9562429822/job/26358788439)).

If we want retain the fasttext backend in Annif for a long term, some alternative provider for fastText should be found. There are a number of more or less active [fastText forks](https://github.com/facebookresearch/fastText/forks?include=active&page=1&period=1y&sort_by=last_updated).

juhoinkinen:
An interesting candidate is [Floret](https://github.com/explosion/floret), available in [PyPI](https://pypi.org/project/floret/). 

It could be more memory-efficient than fastText:

> In order to store word and subword vectors in a more compact format, we turn to an algorithm that's been used by [spaCy](https://spacy.io/) all along: Bloom embeddings. Bloom embeddings (also called the "hashing trick", or known as [HashEmbed](https://thinc.ai/docs/api-layers#hashembed) within spaCy's ML library [thinc](https://thinc.ai/)) can be used to store distinct representations in a compact table by hashing each entry into multiple rows in the table. By representing each entry as the sum of multiple rows, where it's unlikely that two entries will collide on multiple hashes, most entries will end up with a distinct representation.

It should provide the same functionality as fastText:
> floret supports all existing fasttext commands and does not modify any fasttext defaults.

juhoinkinen:
[...]

PR #890, opened 2025-09-01
Use NumPy 2 compatible fastText fork

juhoinkinen:
Upgrading to NumPy 2 in #871 resulted in a bug with fasttext backend: it gives incorrect suggestions, both for previously trained and retrained projects. The scores are the same for every suggestion. fastText doesn't specify a required NumPy version, so it installs without any warnings, but it's not compatible with NumPy 2. 

This is resolved by switching to use the fork of fastText at https://github.com/simon-ging/fasttext-numpy2 / https://pypi.org/project/fasttext-numpy2/, where the compatibility issue has been resolved.

For long term, we should probably find a maintained project offering the fastText algorithm, see tracking issue #795.


osma:
There are compatibility problems between spaCy and fastText. If both the spaCy and fasttext unit tests are run in the same session (`test_analyzer_spacy.py` followed by `test_backend_fasttext.py` and/or `test_project.py`), the fasttext train jobs tend to crash with an "Encountered NaN" error. Perhaps spaCy somehow changes the process state in a way that interfers with fasttext training..? It's really hard to debug this NaN issue. It is intermittent and hard to reproduce locally - some of the crashes happen only under GitHub Actions CI but I can't reproduce them locally.

Some of this (all of the failing tests in `test_backend_fasttext.py`) was mitigated by raising the fasttext model dimensions from 100 to 200. But this didn't help with the tests in `test_project.py`. Maybe it is very sensitive to the training data as well..? In the end I removed the fasttext-related tests entirely from `test_project.py`, because I just could not make them work. I also removed the fasttext project configurations from `tests/projects.cfg`, because they are no longer needed, and adjusted tests accordingly.

We need to get rid of the fasttext dependency soon, because it's obviously suffering from serious bitrot. But for now this PR is a stopgap measure to keep things mostly working for the Annif 1.4 release cycle.

Issue #902, opened 2025-10-10
Field size limitation in CSV corpus format


RietdorfC:
Dear Annif-Team, 

While working with the new CSV corpus format and longer texts, we encountered an error. It seems like the csv library used in Annif has an inherent field size limit. Could this be changed, or should we use a different corpus format if we want to work with longer texts? I've attached the error output below. 

Best regards 
Clemens

```
$ annif train mllm-fulltext test.csv
Traceback (most recent call last):
  File "/home/clemens/test_venvs/annif/bin/annif", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clemens/test_venvs/annif/lib/python3.12/site-packages/flask/cli.py", line 400, in decorator
    return ctx.invoke(f, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #903, opened 2025-10-10
Fix field size limitation in csv corpus format

RietdorfC:
I created this pull request as discussed [here](https://github.com/NatLibFi/Annif/issues/902).


juhoinkinen (approved):
FYI @RietdorfC We are going to make Annif 1.4.1 release soon with this fix, meaning this week.

PR #915, opened 2025-11-11
Fix member paths in zip archive in Hugging Face uploads when using custom data directory

juhoinkinen:
The fix in #908 was only partial: the member paths inside the uploaded zip archives still included the full datadir part, when `ANNIF_DATADIR` environment variable was used:

```
unzip -l ~/Downloads/tfidf-fi.zip 
Archive:  /home/jmminkin/Downloads/tfidf-fi.zip
Archived by Annif 1.5.0.dev0
  Length      Date    Time    Name
---------  ---------- -----   ----
   406760  2025-11-11 13:45   tmp/data/projects/tfidf-fi/vectorizer
   159508  2025-11-11 13:45   tmp/data/projects/tfidf-fi/tfidf-matrix.npz
```

After this everything should be ok.

### DIFF
--- a/annif/hfh_util.py
+++ b/annif/hfh_util.py
@@ -127,8 +127,9 @@ def _archive_dir(data_dir: str) -> io.BufferedRandom:
             encoding="utf-8",
         )
         for fpath in fpaths:
-            logger.debug(f"Adding {fpath}")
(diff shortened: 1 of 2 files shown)


osma (approved):
I trust that it's good now... :crossed_fingers:

PR #923, opened 2025-12-16
Use uv for managing dependencies

### LINKED ISSUE #919: Consider switching to uv for dependency management (esp. PyTorch)
There are currently several development efforts ongoing that would need PyTorch as a dependency:
* #798
* #914
* #895 

PyTorch is a bit special in that it provides different library variants for different kinds of hardware, especially GPUs. `uv` has special support to manage that, while Poetry doesn't.

Some discussion from https://github.com/NatLibFi/Annif/pull/914#issuecomment-3546400885 

> Here are some things that I think would be desirable:
>
>1. It should be possible to use Annif (with EBM) without installing GPU dependencies (even if it's slow). This would be especially desirable in the case of GitHub Actions CI because CI jobs run very often so they should be as lightweight as possible and complete as fast as possible.
>2. It should be possible to use different brands of GPUs, not just limited to NVIDIA/CUDA but also AMD/ROCm and possibly others (e.g. Vulkan) if PyTorch has support.
>
>I realize that these may be difficult to achieve in our current way of managing dependencies. In my understanding, Poetry only has limited support for PyTorch variants making it difficult to implement flexible choices. I think `uv` has better support - see e.g. [here](https://docs.astral.sh/uv/guides/integration/pytorch/). So we could consider switching to `uv` if it helps in this area.

and https://github.com/NatLibFi/Annif/pull/914#issuecomment-3558849902 :
> I must say I'm tempted by the special support for PyTorch that [uv provides](https://docs.astral.sh/uv/guides/integration/pytorch/). It would e.g. allow specifying the PyTorch variant at install time, something like this:
>
>    uv pip install annif[ebm] --torch-backend=cu126
>
>or
>
>    uv pip install annif[ebm] --torch-backend=cpu
>
>If we want to make use of that for Annif itself, we would have to switch from Poetry to uv, which is probably not trivial but should be doable. We have already switched dependency management systems several times in Annif history: I think we started with pipenv, then switch to plain pip+venv, and more recently have been using Poetry.

Switching to uv would affect at least these parts of Annif:

- pyproject.toml: drop Poetry specific configuration, possible add some uv-specific config
- managing development installs (documented in README.md)
- GitHub Actions workflows
- wiki documentation (e.g. how to install optional dependencies)

osma:
This PR switches to uv for dependency management instead of Poetry.

Included changes:
- migrate `pyproject.toml` to work with `uv`; for example, declare the dev dependencies using standard `[dependency-groups]` section
- update README.md to explain how to use `uv` commands for development installs
- migrate GitHub Actions to use `uv` instead of Poetry
- migrate Dockerfile to use `uv` instead of Poetry
- configure `flake8` to skip e.g. `.venv` directory (`uv` will create the virtual environment there by default)

Some notes:

1. `uv` doesn't have a command for permanently activating the venv like the old `poetry shell` or the newer equivalent `eval $(poetry env activate)`. It is recommended to use `uv run` every time, but I think that's cumbersome. However, the venv can simply be activated using `source .venv/bin/activate` so I recommended that in the README. (There is an [open issue](https://github.com/astral-sh/uv/issues/1910) about this, but it seems that implementing this properly to handle all edge cases is a bit difficult, so the developers hesitate to do it.)
2. I changed the Docker base image to one provided by Astral / uv: `ghcr.io/astral-sh/uv:python3.12-bookworm-slim`. An alternative would have been to keep the vanilla base image `python:3.12-slim-bookworm` and just install `uv` within that.
3. I _think_ that the GitHub Actions action `astral-sh/setup-uv` is handling all caching of dependencies, but I haven't verified that it works in a sane way...

Overall, `uv` seems a bit faster than Poetry. For example, the GitHub Actions run completes in less than 3 minutes, when it used to take a bit more than 3 minutes (though with a lot of variation).

In the future, using `uv` should enable better management of PyTorch variants when we add it as a dependency (for example for the EBM backend - see #914 ).

Closes #919

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ __pycache__
 *.pyc
 data
 docs/_build/
-poetry.lock
+uv.lock
 projects.cfg
 venv/
 .hypothesis/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,47 +1,53 @@
-FROM python:3.12-slim-bookworm
+# Use a Python 3.12 + uv image (Debian bookworm-slim)
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
 LABEL org.opencontainers.image.authors="grp-natlibfi-annif@helsinki.fi"
 SHELL ["/bin/bash", "-c"]
 
 ARG optional_dependencies="voikko fasttext nn omikuji yake spacy stwfsa"
-ARG POETRY_VIRTUALENVS_CREATE=false
 
 # Install system dependencies needed at runtime:
 RUN apt-get update && apt-get upgrade -y && \
-	if [[ $optional_dependencies =~ "voikko" ]]; then \
-		apt-get install -y --no-install-recommends \
-			libvoikko1 \
-			voikko-fi; \
-	fi && \
-	# Install rsync for model transfers:
-	apt-get install -y --no-install-recommends rsync && \
-	rm -rf /var/lib/apt/lists/* /usr/include/*
+    if [[ $optional_dependencies =~ "voikko" ]]; then \
+        apt-get install -y --no-install-recommends \
+            libvoikko1 \
+            voikko-fi; \
+    fi && \
+    # Install rsync for model transfers:
+    apt-get install -y --no-install-recommends rsync && \
+    rm -rf /var/lib/apt/lists/* /usr/include/*
 
 WORKDIR /Annif
-RUN pip install --upgrade pip "poetry~=2.0" --no-cache-dir
 
+# Copy only project metadata first to maximize Docker layer caching
 COPY pyproject.toml setup.cfg README.md LICENSE.txt CITATION.cff projects.cfg.dist /Annif/
 
-# First round of installation for Docker layer caching:
-RUN echo "Installing dependencies for optional features: $optional_dependencies" \
-	&& poetry install -E "$optional_dependencies" --no-root \
-	&& rm -rf /root/.cache/pypoetry  # No need for cache because of poetry.lock
+# First round: install dependencies only (no project), with selected extras.
+RUN extras=(); \
+    for e in ${optional_dependencies}; do extras+=(--extra "$e"); done; \
+    uv sync --no-install-project --no-dev "${extras[@]}"
 
 # Download nltk data
-RUN python -m nltk.downloader punkt_tab -d /usr/share/nltk_data
+RUN uv run --no-sync python -m nltk.downloader punkt_tab -d /usr/share/nltk_data
 
-# Download spaCy models, if the optional feature was selected
+# Download spaCy models only if 'spacy' extra is selected
 ARG spacy_models=en_core_web_sm
 RUN if [[ $optional_dependencies =~ "spacy" ]]; then \
-		for model in $(echo $spacy_models | tr "," "\n"); do \
-			python -m spacy download $model; \
-		done; \
-	fi
+        for model in $(echo "$spacy_models" | tr "," "\n"); do \
+            uv run --no-sync python -m spacy download "$model"; \
+        done; \
+    fi
 
-# Second round of installation with the actual code:
+# Second round: add source and install the actual project (editable by default)
 COPY annif /Annif/annif
 COPY tests /Annif/tests
-RUN poetry install -E "$optional_dependencies"
+RUN extras=(); \
+    for e in ${optional_dependencies}; do extras+=(--extra "$e"); done; \
+    uv sync --no-dev "${extras[@]}"
 
+# Make virtualenv executables available to shell and entrypoint
+ENV PATH="/Annif/.venv/bin:${PATH}"
+
+# Enable Annif bash completion (now available on PATH)
 WORKDIR /annif-projects
 RUN annif completion --bash >> /etc/bash.bashrc  # Enable tab completion
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,13 @@ RUN apt-get update && apt-get upgrade -y && \
 
 WORKDIR /Annif
 
+RUN groupadd -g 998 annif_user && \
+    useradd -m -u 998 -g annif_user annif_user && \
+    chown -R annif_user:annif_user /Annif
+USER annif_user
+
 # Copy only project metadata first to maximize Docker layer caching
-COPY pyproject.toml setup.cfg README.md LICENSE.txt CITATION.cff projects.cfg.dist /Annif/
+COPY --chown=annif_user:annif_user pyproject.toml setup.cfg README.md LICENSE.txt CITATION.cff projects.cfg.dist /Annif/
 
 # First round: install dependencies only (no project), with selected extras.
 RUN extras=(); \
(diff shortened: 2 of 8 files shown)


osma:
There's a potential problem when using `uv` together with custom spaCy models. `python -m spacy download` downloads a model and installs it as a Python package into the current venv using `pip` (that's why I added `pip` as a dependency in this PR; see also https://github.com/explosion/spaCy/issues/13846). But the new model package is not registered in `pyproject.toml`, so `uv` doesn't know about it. Next time you run `uv sync` for any reason, it will delete this "extra" package, unless you give it the `--inexact` flag every time (there is no permanent setting for this).

This has been reported as https://github.com/astral-sh/uv/issues/12481 but it was closed by `uv` developers as they are not planning to support this kind of use.

This has also been discussed in https://github.com/explosion/spaCy/issues/13747 where it was suggested to call e.g. `uv run spacy info en_core_web_sm --url` to get the package URL and then passing it to `uv add` to add a real dependency tracked by `uv`. But we want to be flexible and allow users to install any spaCy models they need, so having to modify `pyproject.toml` (via `uv add`) seems a bit wrong.

OTOH, this only affects development installs (and currently Docker images that also use `uv`). A regular user installing Annif via `pip install annif[spacy]` would not be affected by this unless they specifically use `uv pip`.

osma:
Note to self: The `pip` dependency could be moved to the `spacy` extra instead of having it as a top level dependency. Nothing but spaCy depends on it.

EDIT: Done in 65abb149

.gitignore
juhoinkinen, line 1: I recalled that in addition to the `.gitignore` file, there is is also the [`.dockerignore`](https://github.com/NatLibFi/Annif/blob/issue919-uv-for-dependencies/.dockerignore).

Maybe the `.venv` could be added to it, to avoid adding that directory to Docker build context? With this PR,  the venv with all dependencies it is 2.2 GBs, not so much, but unnecessary for Docker builds anyway.

juhoinkinen: A sidenote: the `.venv/` is ignored by git, because there is a .gitignore file in `.venv/.gitignore` (as shown by `git check-ignore .venv/ -v`).

Dockerfile
```diff
@@ -1,47 +1,53 @@
...
+# Second round: add source and install the actual project (editable by default)
+COPY annif /Annif/annif
+COPY tests /Annif/tests
+RUN extras=(); \
+    for e in ${optional_dependencies}; do extras+=(--extra "$e"); done; \
+    uv sync "${extras[@]}"
+
+# Download spaCy models only if 'spacy' extra is selected
 ARG spacy_models=en_core_web_sm
 RUN if [[ $optional_dependencies =~ "spacy" ]]; then \
```
juhoinkinen, line 41: The layer by this line has a size of 2.02 GBs in Docker image.

Dockerfile
juhoinkinen, line 1: There is a possibilty to [enable bytecode compilation](https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode) for uv installs (apparently uv does not perform it by default?); we could add `UV_COMPILE_BYTECODE=1` to Dockerfile (although when I checked its effect with plain `annif` command, I cannot see a difference in startup time).

juhoinkinen (commented):
I noticed that the Dockerimage size doubles, and AI tools recommended the changes as in the attached file.

juhoinkinen:
The mentioned Dockerfile here, with .txt extension to allow sharing in GitHub.

[Dockerfile.txt](https://github.com/user-attachments/files/24472165/Dockerfile.txt)

.gitignore
osma: Added .venv to .dockerignore

Dockerfile
osma: Fixed using suggested Dockerfile, thanks!

osma:
@juhoinkinen Thanks for the review! I made the suggested changes. The Docker image size is now 2.36 GB, which is close to the current `main` image size 2.3GB.

But now SonarCloud complains about ownership of copied files. See [full explanation](https://sonarcloud.io/organizations/natlibfi/rules?open=docker%3AS6504&rule_key=docker%3AS6504).

The problem (if it is a problem indeed) is not new but SonarCloud only noticed it now that the COPY commands were adjusted instead of doing separate chown operations.

I wonder if file ownership for the annif user is really necessary within the image?

juhoinkinen:
I tried using `COPY --chmod=644` instead of `COPY --chown=annif_user:annif_user `, but it does not give access permissions to subdirectories; this is what resulted when building an image:
```
ls -l /Annif/annif/
total 192
-rw-r--r-- 1 root root  3542 Dec 18 11:46 __init__.py
drw-r--r-- 2 root root  4096 Dec 18 11:55 analyzer
drw-r--r-- 2 root root  4096 Dec 18 11:55 backend
-rw-r--r-- 1 root root 30330 Dec 18 11:46 cli.py
-rw-r--r-- 1 root root 10176 Dec 18 11:46 cli_util.py
...
```
The subdirectories cannot be accessed (as annif_user) and Annif cannot start:
```
ls -l /Annif/annif/analyzer/
ls: cannot access '/Annif/annif/analyzer/spacy.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/estnltk.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/simplemma.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/__init__.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/snowball.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/analyzer.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/voikko.py': Permission denied
ls: cannot access '/Annif/annif/analyzer/simple.py': Permission denied
total 0
-????????? ? ? ? ?            ? __init__.py
-????????? ? ? ? ?            ? analyzer.py
-????????? ? ? ? ?            ? estnltk.py
-????????? ? ? ? ?            ? simple.py
-????????? ? ? ? ?            ? simplemma.py
-????????? ? ? ? ?            ? snowball.py
-????????? ? ? ? ?            ? spacy.py
-????????? ? ? ? ?            ? voikko.py
```

Previously the chmod command was `chmod -R a+rX /Annif`, the capital X giving the directory access.

I think that was good: everyone could read the files, but only root could write. To achieve that with uv without doubling the image size, maybe a multistage build should be utilized. :thinking:

osma:
@juhoinkinen It turned out that the chown/chmod options were mostly useless. Apparently most of these were introduced by the AI tools used to suggest changes to the Dockerfile. I adjusted the Dockerfile so it's now much closer to the original (before this PR). Image size is still ~2.37GB.

There are still a couple of open questions:
* I dropped the `chmod -R a+rX /Annif` command from Dockerfile because it didn't seem to do anything useful (the file permissions look OK to me even without it), but I may have missed something
* I didn't try the bytecode compilation that you [suggested](https://github.com/NatLibFi/Annif/pull/923#discussion_r2668369787)

juhoinkinen (approved):

osma:
I added the chmod command back because it is useful after all: if a developer has strict file permissions (non world readable/traversable), then they will be copied to the container with those permissions and cannot be accessed by annif_user.

I had to adjust the command a little to avoid matching `.venv`, which would blow up the size of the container.

osma:
Posted a heads up about the change on annif-users: https://groups.google.com/g/annif-users/c/J1tikQAQQAE

osma:
Based on what I learned here in this PR, I [suggested](https://github.com/astral-sh/uv/issues/17511) a feature for uv to support a shorter syntax for uv sync extras: `uv sync -E foo,bar`

Issue #934, opened 2026-02-17
Documentation Issue in Annif README with uv sync


mfakaehler:
When installing a dev-Version of Annif with uv and multiple backends, the README.md says to use the syntax:

```
uv sync --extra "extra1 extra2"
```
This did not work for me. However, what worked is 
```
uv sync --extra extra1 --extra extra2
```

In the [uv-docs](https://docs.astral.sh/uv/reference/cli/#uv-sync--extra) it says "May be provided more then once". So I think this is the intended syntax. 

I will open a PR for the Annif-README to suggest a fix.

PR #935, opened 2026-02-17
Fix documentation for usage of uv sync

mfakaehler:
This fixes issue #934, explaining how to install multiple extras with uv sync.


juhoinkinen (approved):
Thanks!

PR #941, opened 2026-06-02
Fix REST API 500 error and schemathesis flaky test

osma:
While working on #926 I had problems with a failing schemathesis test under GitHub Actions CI. Investigating it revealed two problems:

1. The REST API can return 500 errors for untrained projects, even though the response should be 503 with a proper error message.
2. The schemathesis test fails when the API returns a 503 response, even though it is correct and documented in the schema.

This PR fixes the two errors above.


osma (commented):
LGTM (can't approve this myself)

PR #943, opened 2026-06-15
Raise a readable error when a vocabulary file fails to parse

### LINKED ISSUE #861: Readable error message for failed vocabulary file loading
If an error occurs on loading subjects from a vocabulary file a full traceback is shown (as in #860, reported by @s0rin). The exception could be catched and only the relevant part showed.

Catching could be just around RDF graph parsing in here 
https://github.com/NatLibFi/Annif/blob/5109f0ea9adfc2805e8cd478a25b909fca800ce1/annif/corpus/skos.py#L70

or around that and `joblib.load(path)` a few lines before, or even in [`run_load_vocab()`](https://github.com/NatLibFi/Annif/blob/5109f0ea9adfc2805e8cd478a25b909fca800ce1/annif/cli.py#L155) in `annif/cli.py` file to handle also errors on reading CSV/TSV files.

avalyset:
Fixes #861.

When `load-vocab` is given a corrupt or malformed SKOS/RDF file, the raw rdflib traceback surfaces instead of a readable message. This wraps the `graph.parse()` call in `VocabFileSKOS.__init__` and raises `OperationFailedException` with the file path and the underlying error.

I wrapped only the `graph.parse()` call per your steer and left `joblib.load()` untouched, since that path loads Annif's own stored dump, not user-supplied input. I used a broad `except Exception` because rdflib's parse surface (ParserError, BadSyntax, SAXParseException, plus format guessing) is large and unpredictable for arbitrary user files; happy to narrow it to an explicit exception tuple if you'd prefer.

I added a `test_load_vocab_corrupt` test in `tests/test_vocab_skos.py` covering an invalid Turtle file.

### DIFF
--- a/annif/vocab/skos.py
+++ b/annif/vocab/skos.py
@@ -12,6 +12,7 @@ import rdflib.util
 from rdflib.namespace import OWL, RDF, RDFS, SKOS
 
 import annif.util
+from annif.exception import OperationFailedException
 
 from .types import Subject, VocabSource
 
@@ -67,7 +68,12 @@ class VocabFileSKOS(VocabSource):
             self.graph = joblib.load(path)
         else:
             self.graph = rdflib.Graph()
-            self.graph.parse(self.path, format=rdflib.util.guess_format(self.path))
+            try:
+                self.graph.parse(self.path, format=rdflib.util.guess_format(self.path))
+            except Exception as err:
+                raise OperationFailedException(
+                    f"Cannot parse vocabulary file {self.path}: {err}"
+                ) from err
 
     @property
     def languages(self) -> set[str]:
--- a/tests/test_vocab_skos.py
+++ b/tests/test_vocab_skos.py
@@ -141,3 +141,4 @@ yso:p8993
     with pytest.raises(OperationFailedException) as excinfo:
         VocabFileSKOS(tmpfile_path)
     assert "Cannot parse vocabulary file" in str(excinfo.value)
+    assert tmpfile_path in str(excinfo.value)
--- a/tests/test_vocab_skos.py
+++ b/tests/test_vocab_skos.py
@@ -137,6 +137,7 @@ yso:p8993
     """  # missing closing period -> invalid Turtle
     )
 
+    tmpfile_path = str(tmpfile)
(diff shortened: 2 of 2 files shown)


avalyset:
The latest commit is empty — it only retriggers CI, no code change.

The earlier red was unrelated to this diff: `test on Python 3.13` failed with `RuntimeError: Encountered NaN` during fastText model training, and the 3.10/3.11/3.12 jobs were cancelled by fail-fast rather than failing on their own. The test that exercises this change, `test_exception.py`, passed. This looks like fastText training flakiness on 3.13 rather than anything in the vocabulary-parse error handling.

The re-run needs maintainer approval to start (fork PR). Happy to rebase if you'd prefer — just let me know.

osma:
Thanks for this PR @avalyset ! Can you please address this SonarCloud complaint shown in the [changes view](https://github.com/NatLibFi/Annif/pull/943/changes):

> Refactor this exception test to have only one invocation possibly throwing an exception.
> Check warning: Refactor this exception test to have only one invocation possibly throwing an exception.
>
> See more on https://sonarcloud.io/project/issues?id=NatLibFi_Annif&issues=AZ9M_RS4pfBkRQAQEhJt&open=AZ9M_RS4pfBkRQAQEhJt&pullRequest=943

Basically you need to move the `str(tmpfile)` operation outside the `with` block so that the block only contains one function/method call, not two.

avalyset:
Done in f6619f6 — `str(tmpfile)` now happens before the `with` block, so only the VocabFileSKOS call is inside it.

avalyset:
The SonarCloud complaint is addressed in `f6619f6` — `str(tmpfile)` now happens before the `with` block. Both SonarCloud checks pass and the PR shows as clean. Let me know if anything else is needed.

avalyset:
Added the path assertion from the second review comment — `tmpfile_path` was already in scope, so the test now checks that the path reaches the message and not just the prefix.

On the first comment: Annif leaves file paths unquoted in its error messages, `corpus/document.py:158` has the same shape, and no message anywhere in `annif/` quotes a path. I kept the wording consistent with that rather than switching to repr.

PR #944, opened 2026-06-16
Configure cool-down period for uv-handled dependencies

juhoinkinen:
uv can be configured to not update any dependency whose artifact has been uploaded in a given period of time: https://docs.astral.sh/uv/reference/settings/#exclude-newer

This PR adds config for this "cool-down" period of one week, which can help in preveventing supply-chain attacks: https://pydevtools.com/handbook/how-to/how-to-protect-against-python-supply-chain-attacks-with-uv/


osma (approved):
lgtm

PR #946, opened 2026-06-30
Expose project metadata fields via REST API

### LINKED ISSUE #893: Add information on metadata fields of projects to REST API
The `suggest`, `suggest_batch` and `learn` methods of REST API now accept metadata of a document in addition to text (implemented in #867 ), which a backend can then use if a project configuration sets a select transform (implemented in #864). From the transform configuration, e.g. `select(title,description,text)` it is clear which metadata fields the project in question uses, but this information is not available for REST API users.

The project information accessible via `/v1/projects/<projectid>` could show also the metadata fields that the project uses when it performs the suggest operation.

For example, the response could include a field named `metadata`  containing the names of the metadata fields:
```
{
  "backend": {
    "backend_id": "dummy"
  },
  "is_trained": true,
  "language": "fi",
  "modification_time": null,
  "name": "Dummy Finnish",
  "project_id": "dummy-fi",
  "vocab": {
    "languages": [
      "en",
      "fi"
    ],
    "vocab_id": "dummy"
  },
  "vocab_language": "fi",
  "metadata": [  // Added list
    "title",
    "description",
    "text"
   ]
}
```

For ensemble projects, the implementation should allow gathering the metadata fields that the source projects use, in addition to the fields the ensemble project itself uses.

avalyset:
Closes #893.

Exposes the document metadata fields a project uses during `suggest` via the
REST API, as requested in the issue. `GET /v1/projects` and
`GET /v1/projects/<project_id>` now return a `metadata` field — a flat list of
the field names a project consumes through `select(...)` transforms:

    "metadata": ["title", "description", "text"]

Following the issue's spec for ensemble projects ("gather the metadata fields
that the source projects use, in addition to the fields the ensemble project
itself uses"), an ensemble's `metadata` is the order-preserving, de-duplicated
union of its own select fields and those of its source projects.

One implementation note: `metadata_fields()` skips a source project that can't
be resolved (`get_project` raising `ValueError`) rather than propagating, so it
adds no new failure mode to `dump()`. This is narrower than the `vocab`
`try/except` above it — `dump()` already evaluates `is_trained` (which resolves
ensemble sources) before this point, so a genuinely missing source still
surfaces there exactly as today; the skip only keeps `metadata_fields()` itself
from being a second raise point. Happy to make it strict (raise) if you'd prefer.

## Changes
- `annif/project.py`: `metadata_fields()` + `_select_fields()` helpers; `metadata` added to `dump()`.
- `annif/openapi/annif.yaml`: `metadata` added to the `Project` schema (keeps schemathesis response validation green).
- Tests: a `select(...)` project and a metadata-gathering ensemble fixture; unit + REST coverage for own fields, ensemble union/dedup, and unloadable-source skip.

## Validation
- `pytest tests/test_project.py tests/test_rest.py tests/test_openapi.py` — green (pre-existing tfidf-training failures unrelated to this change).
- `flake8` / `black --check` / `isort --check` — clean.

### DIFF
--- a/annif/openapi/annif.yaml
+++ b/annif/openapi/annif.yaml
@@ -353,6 +353,14 @@ components:
           format: date-time
           example: 2020-05-19T16:42:42Z
           nullable: true
+        metadata:
+          type: array
+          items:
+            type: string
+          example:
+          - title
+          - description
+          - text
       description: A project definition
     ProjectList:
       type: object
--- a/annif/project.py
+++ b/annif/project.py
@@ -340,11 +340,15 @@ class AnnifProject(DatadirMixin):
                 fields.extend(trans.fields)
         return fields
 
-    def metadata_fields(self) -> list[str]:
-        """return the metadata fields this project uses during suggest. For
-        ensemble projects, the fields of the source projects are gathered in
-        addition to the project's own fields. The result is a flat, order-
-        preserving, de-duplicated list."""
+    def _gather_metadata_fields(self, seen: set[str]) -> list[str]:
+        """collect the select(...) fields of this project and, recursively, of
+        its source projects. The recursion mirrors suggest(), which delegates
+        to each source project's own suggest(). `seen` guards against cyclic
+        sources configurations, which are not rejected elsewhere."""
+
+        if self.project_id in seen:
+            return []
+        seen.add(self.project_id)
 
         fields = list(self._select_fields())
 
@@ -359,10 +363,18 @@ class AnnifProject(DatadirMixin):
                     source = self.registry.get_project(source_id)
                 except ValueError:
                     continue  # source project not loadable - skip best-effort
-                fields.extend(source._select_fields())
+                fields.extend(source._gather_metadata_fields(seen))
+
+        return fields
+
+    def metadata_fields(self) -> list[str]:
+        """return the metadata fields this project uses during suggest. For
+        ensemble projects, the fields of the source projects are gathered
+        recursively, in addition to the project's own fields. The result is a
+        flat, order-preserving, de-duplicated list."""
 
         # de-duplicate while preserving first-seen order
-        return list(dict.fromkeys(fields))
+        return list(dict.fromkeys(self._gather_metadata_fields(set())))
 
     def dump(self) -> dict[str, str | list | dict | bool | datetime | None]:
         """return this project as a dict"""
--- a/annif/project.py
+++ b/annif/project.py
@@ -22,7 +22,7 @@ from annif.exception import (
     NotInitializedException,
     NotSupportedException,
 )
-from annif.util import parse_args
+from annif.util import parse_args, parse_sources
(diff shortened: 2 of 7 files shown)


avalyset:
Thanks — the three inline comments from 27 July, taken in turn. One is a real
bug and is now fixed; on the other two I'd push back.

**`annif/project.py:363` — source fields not gathered recursively. Fixed.**

Correct, and it reproduces. I added `ensemble-meta-nested`, an ensemble whose
only source is `ensemble-meta`, and on the old code it returned
`["subtitle", "description"]` — dropping `title` and `text`, which
`select-meta` contributes one level further down. Since the ensemble backend
calls each source project's full `suggest()`, those fields really are used at
suggest time, so `metadata` was under-reporting.

Gathering is now recursive, with one deviation from the suggestion: instead of
calling `source.metadata_fields()`, the recursion runs through a private
`_gather_metadata_fields(seen)`, and `metadata_fields()` stays a thin wrapper
over it. A cyclic `sources` configuration doesn't appear to be rejected
anywhere in Annif, and plain recursion would turn `GET /v1/projects` into a
`RecursionError`; threading the `seen` set needs a parameter that doesn't
belong on the public method. It also keeps de-duplication as a single pass at
the top rather than re-running it at every level, which preserves the existing
ordering. Added tests for both the nested case and the cycle guard, and bumped
the fixture project counts 19 → 20.

**`annif/openapi/annif.yaml:349` — invalid YAML. Not reproducible; no change.**

The example parses correctly as written:

```python
>>> yaml.safe_load(open("annif/openapi/annif.yaml"))["components"]["schemas"] \
...     ["Project"]["properties"]["metadata"]["example"]
['title', 'description', 'text']
```

A block sequence at the same indentation as its parent key is valid YAML —
indenting it further is optional, not required. `tests/test_openapi.py` (22
tests, which load the spec through connexion) passes unchanged. I'm happy to
indent it anyway if that's the house style you'd prefer, but it isn't a
correctness problem.

**`annif/project.py:341` — pre-check on `transform_spec`. No change.**

`self.transform` is memoized, and building a transform chain is cheap: I
measured ~2 µs each for `pass`, `select(...)`, `limit(...)` and `filter_lang`
(the `LanguageDetector` that `filter_lang` constructs doesn't load
dictionaries until it's used). The suggested pre-check would be a substring
test against `transform_spec`, which duplicates what `parse_specs()` already
does and would misfire on any future transform whose name or arguments happen
to contain "select". That seems like a poor trade for microseconds on a
property `suggest()` builds anyway.

---

Full suite after the change: **462 passed, 13 skipped**. Running only a subset
of the test files shows a handful of failures (the tfidf training tests and
three `EOFError`s in the initialization tests), but those are ordering
artefacts — I diffed the failure set against a clean checkout of the branch
point and it is identical, so there is no regression from this change.
`black`, `isort` and `flake8` are clean on the touched files.

PR #955, opened 2026-09-21
Update dependencies for version 1.5

juhoinkinen:
WIP

I deliberately did not update to black 26, because using the black 26 format would touch 14 files (all unit tests) and at this point of the current year the black 27 is already not too far in the future, and we could update directly from black 25 to 27.

TODO:
- [ ] Upgrade to omikuji 0.5.2 once the cooldown period (5 days) allow this


juhoinkinen:
NumPy, SciPy etc. cannot be upgraded to latest versions when having support for Python 3.10:
```
Because the requested Python version (>=3.10, <3.14) does not satisfy
      Python>=3.11 and numpy>=2.3.3,<=2.3.5 depends on Python>=3.11, we can
      conclude that numpy>=2.3.3,<=2.3.5 cannot be used.
```

